PR #15, opened 2017-01-09
Buildsamplecode

yungez:
this is to add build scripts for raspberry pi and Edison c sample codes.
@tameraw  @aovechki


tameraw:
@yungez - Please resubmit against develop (code involved).

PR #17, opened 2017-01-10
Buildsamplecode

yungez:
this is to add build scripts for raspberry pi and Edison c sample codes. @tameraw @aovechki


anporumb:
Hello @yungez 

This is what (in broad strokes) the scripts to be merged are doing:
1) clone Azure-Samples
2) clone azure-iot-sdk-c
3) samples from Azure-Samples are build using azure-iot-sdk-c

Should be accept these scripts, we would end up with the following situation: when samples are changed, the scripts might fail (samples might not compile) and we would then fail our gated builds. Notice how we do not own the samples repo. Effectively this means we would gate our repo on the work that exists in another repo! I am very sorry to say, but as it is, we cannot accept this pull request for the reasons above.

At this moment I would like to propose the following two alternatives:
A) you submodule our repo in Azure-Samples - that gives you peace of mind because you are working with a fixed SHA and only advance the SHA when *you* chose.
B) you can consider moving your samples inside of our repo - this would give us peace of mind that no uncontrolled (gated) changes appear in our repo.

Which of the above alternatives would you chose?

Best Regards,
Andrei Porumb

PR #57, opened 2017-02-04
Fix build errors iothub_client, serializer module and iothub_client_sample_mqtt_dm (FW update) build

abhatikar:
Fix build errors in iothub_client & serializer module.			
Fix iothub_client_sample_mqtt_dm (FW update) build


anporumb:
Hello @abhatikar ,

We suspect you are using an old branch ("develop"). Would you please redo your pull request against master branch?

Why is the agenttypes change needed?

Best Regards,
Andrei Porumb

dcristoloveanu:
The reason this was closed is because the develop branch was removed.
We are switching to only using the master branch as we did not have significant advantages from using master and develop.

Thanks!
/Dan

PR #60, opened 2017-02-07
Fix iothub_client_sample_mqtt_dm (FW update) build

abhatikar:
Fix iothub_client_sample_mqtt_dm (FW update) build


anporumb:
The content of this PR has been checked in. 

I will close it once it is released.

Best Regards,
Andrei Porumb

PR #77, opened 2017-02-24
Handle MQTT_CLIENT_COMMUNICATION_ERROR

### LINKED ISSUE #76: When sending fails, MQTT client goes to the unknown error code path
If sending fails, for example, in a scenario like https://github.com/Azure/azure-iot-sdk-c/issues/73 the MQTT client ends up logging `INTERNAL ERROR: unexpected error value received`.

This is because in `mqtt_client.c`, the error is returned like this:

```
        else if (send_result == IO_SEND_ERROR)
        {
            LOG(AZ_LOG_ERROR, LOG_LINE, "MQTT Send Complete Failure send_result: %d", (int)send_result);
            set_error_callback(mqtt_client, MQTT_CLIENT_COMMUNICATION_ERROR);
        }
```

And in `iothubtransport_mqtt_common.c` that error is considered unknown:

```
            case MQTT_CLIENT_COMMUNICATION_ERROR:
            case MQTT_CLIENT_UNKNOWN_ERROR: 
            {
                LogError("INTERNAL ERROR: unexpected error value received %s", ENUM_TO_STRING(MQTT_CLIENT_EVENT_ERROR, error));
            }
```

Sending data through a socket may fail for various reasons and one option for better handling of this scenario might be to try to reconnect.

vjrantal:
Fixes #76.

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -1436,6 +1436,8 @@ static void mqtt_error_callback(MQTT_CLIENT_HANDLE handle, MQTT_CLIENT_EVENT_ERR
             case MQTT_CLIENT_PARSE_ERROR:
             case MQTT_CLIENT_MEMORY_ERROR:
             case MQTT_CLIENT_COMMUNICATION_ERROR:
+                LogError("Mqtt Communication Error. Reconnecting device...");
+                break;
             case MQTT_CLIENT_UNKNOWN_ERROR: 
             {
                 LogError("INTERNAL ERROR: unexpected error value received %s", ENUM_TO_STRING(MQTT_CLIENT_EVENT_ERROR, error));


anporumb:
Hello @vjrantal ,

At this moment we believe your code would not be a good fix. Essentially it is a "same as before with a different message on the screen". Maybe not all the files are attached?

Best Regards,
Andrei Porumb

vjrantal:
@anporumb All files that I intended to have in the PR were there.

Here is the reason I proposed this change.

I was looking at my logs when connection errors occur and I saw `INTERNAL ERROR: unexpected error value received` getting printed. Without knowing the internals of the code (which is the case for many users of this SDK), it looked to me as something was going wrong. That is why I proposed (with this PR) that the error message printed would instead be `Mqtt Communication Error. Reconnecting device...` which indicates better what is going on.

Even though this is a non-functional change, it is helpful for developers if the logging from the SDK is as accurate as possible.

ritazh:
I agree with @vjrantal. Adding this logging helps with developer and debugging experience. Why is this closed?

olivierbloch:
I don't think we disagree on the fact that we need better error messages and we will definitively add some more usable error message in the context of the device connection status.
That said, the actual pull request doesn't really solve much and is actually incorrect: this code will log a "communication error" even for parse and memory errors which is actually even more misleading than logging "unknown error" for all errors.

```c
             case MQTT_CLIENT_PARSE_ERROR: 
             case MQTT_CLIENT_MEMORY_ERROR: 
             case MQTT_CLIENT_COMMUNICATION_ERROR: 
+                LogError("Mqtt Communication Error. Reconnecting device..."); 
+                break; 
             case MQTT_CLIENT_UNKNOWN_ERROR:  
             { 
                 LogError("INTERNAL ERROR: unexpected error value received %s", ENUM_TO_STRING (MQTT_CLIENT_EVENT_ERROR, error)); 
```

vjrantal:
@olivierbloch Good point about the fact that the commit would change also logging for parse and memory errors (that wasn't the intent). For that reason, I agree with the resolution.

PR #93, opened 2017-03-22
Added Mac support

sHorst:
# Checklist

# Description of the problem
you cannot build the python module on a Mac.

# Description of the solution
I copied the linux version and made the changes to compile on an mac.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,11 +32,18 @@ IF(WIN32)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
     # Make warning as error
     add_definitions(/WX)
-ELSE()
+ELSEIF(APPLE)
     # Make warning as error
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+ELSE()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
-ENDIF(WIN32)
+ENDIF()
+
+IF(APPLE)
+    SET(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
+ENDIF(APPLE)
 
 #the following variables are project-wide and can be used with cmake-gui
 option(use_amqp "set use_amqp to ON if amqp is to be used, set to OFF to not use amqp" ON)
--- a/build_all/mac/build.sh
+++ b/build_all/mac/build.sh
@@ -1,0 +1,161 @@
+#!/bin/bash
+#set -o pipefail
+#
+
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+build_root=$(cd "${script_dir}/../.." && pwd)
+log_dir=$build_root
+run_e2e_tests=OFF
+run_longhaul_tests=OFF
+build_amqp=ON
+build_http=ON
+build_mqtt=ON
+no_blob=OFF
+use_wsio=OFF
+run_unittests=OFF
+build_python=OFF
+build_javawrapper=OFF
+run_valgrind=0
+build_folder=$build_root"/cmake/iotsdk_linux"
+make=true
+toolchainfile=" "
+cmake_install_prefix=" "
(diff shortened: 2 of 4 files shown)


damonbarry:
I don't know about python, but this repo already builds for macOS. Every change gets built on a Mac before it is merged into master ([here](https://github.com/Azure/azure-iot-sdk-c/blob/master/jenkins/osx_c.sh) is the Jenkins script). So this change is just about adding a build script for macOS.

CMakeLists.txt
```diff
@@ -32,11 +32,18 @@ IF(WIN32)
...
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+ELSE()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
-ENDIF(WIN32)
+ENDIF()
+
+IF(APPLE)
+    SET(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
```
damonbarry:
```
SET(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
```
What's your reasoning for making this change?

sHorst: on mac python does not load a .dylib but a .so so it would not load otherwise.

sHorst:
to the mac building. yes, if you run the linux/build.sh it compiles. but with the --build-python it does not build. and this patch will make it work even with this option. We need this for the python sdk to compile

CMakeLists.txt
damonbarry: OK, so the Python tools have a requirement that the thing they load must have an .SO suffix, whether it's a Mach-O shared library or a Mach-O module. I assume Python will dynamically load it with dlopen? In that case it really could be a shared library or a module. But other projects (e.g. the [gateway SDK](https://github.com/Azure/azure-iot-gateway-sdk)) need to be able to **link against** the IoT Hub device client library, so it can't be a module; it _has_ to be a shared library.

By convention, shared libraries have a DYLIB extension, and I'd rather not rename all the libraries in this repo to .SO just to make the Python tools happy.

Over in the Python repo (I see that you have an open pull request over there), why not set `CMAKE_SHARED_LIBRARY_SUFFIX` in the root CMakeLists.txt file, right before the call to  `add_subdirectory(c)`? Then all shared libraries in this repo will be renamed to .SO, but _only_ in the context of the Python SDK.

damonbarry:
@sHorst I raised the point about the build script because the C-language repos for the Azure IoT SDKs are generally moving away from maintaining build scripts in favor of documenting the CMake commands you'd use to do it yourself (e.g. see the [setup guide](https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/devbox_setup.md#macos) in this repo). In some of the smaller repos we removed the build.cmd/.sh files altogether. In nearly every repo, we removed all references to build.cmd/.sh from the README files.

Adding another (mostly duplicated) build script for Mac isn't the end of the world, but it seems like it would be easier to just add the two or three needed commands directly into the build script in the Python repo (basically just `cmake [options] ..` and `cmake --build .`.

BTW, build scripts probably still make sense in the Python repo--language interop always makes the build more complicated.

tameraw:
@sHorst - We're moving away from build.sh scripts and relying on Cmake. We will retire the current ones as soon as we can, so we do not want to add more scripts to the build. Instructions for building on the mac are available here:
https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/devbox_setup.md

sHorst:
so you do not use this pull request, and so the python sdk will not build on mac. This patch is needed for this pr: https://github.com/Azure/azure-iot-sdk-python/pull/20

damonbarry:
@sHorst You could move two changes from this PR into the Python SDK's repo, and then this PR would no longer be required:
1. Add `SET(CMAKE_SHARED_LIBRARY_SUFFIX ".so")` to the root CMakeLists.txt file, right before the call to `add_subdirectory(c)`
2. Instead of calling `./c/build_all/Mac/build.sh ...` from build_all/Mac/build.sh, invoke the three or four cmake commands directly.

sHorst:
ok .. i will look into this.. thx

PR #134, opened 2017-05-11
Update the format of the release tags

ollyw:
The documentation has the incorrect format of the release tags for the git repo (no `release_` prefix, and `-` instead of `_`. This lead to failure of the git clone command.

### DIFF
--- a/doc/devbox_setup.md
+++ b/doc/devbox_setup.md
@@ -19,11 +19,11 @@ This document describes how to prepare your development environment to use the *
 - Install [CMake]. Make sure it is in your PATH by typing `cmake -version` from a command prompt. CMake will be used to create Visual Studio projects to build libraries and samples.
 
 - Locate the tag name for the [latest release][latest-release] of the SDK.
-  > Our release tag names are date values in `yyyy-mm-dd` format.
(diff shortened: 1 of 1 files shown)


ewertons:
Hi @ollyw ,

we see your point, indeed the "release" tags are in the format release_YYYY_MM_DD.
The document refers to commit tags, in the format of YYYY-MM-DD, which we use more widespread.

You can use the format bellow, what would work:
```
git clone -b 2017-05-05 --recursive https://github.com/Azure/azure-iot-sdk-c.git
```

Let us discuss how to better improve the documentation, and we will reply back briefly.

ewertons:
Now giving you more details...

if you see any branches named release_YYYY_MM_DD in the repo, they are temporary and should not be used to clone the repo.

For example, if you try running the command below, you will get a failure:


```
ewertons@ewertons-ubuntu02:~/test$ git clone -b release_2017_04_21 https://github.com/Azure/azure-iot-sdk-c.git
Cloning into 'azure-iot-sdk-c'...
fatal: Remote branch release_2017_04_21 not found in upstream origin
```

So the change you are proposing would not produce the correct behavior.

ewertons:
The PR will be closed, but please let us know if you have any further comments or questions.

PR #145, opened 2017-05-26
Fix for incorrect generation of USERNAME string when trying to reconnect in MQTT protocol.

pgonzalezJean:
When the MQTT protocol tries to reconnect, the 'DeviceClientType' property is appended to the username string on each connection retry. So, that property is repeated the number of times a reconnection has happened. These screenshots are self-explanatory.

First connection:
image

After several connection retries:
image

The solution is to create a temporary string to append the 'DeviceClientType' and use it as the username for the MQTT client options.

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -1721,6 +1721,8 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
     {
         void* product_info;
         STRING_HANDLE clone;
+		// Create a clone of 'transport_data->configPassedThroughUsername' to keep the original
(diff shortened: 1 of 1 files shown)


tameraw:
@pgonzalezJean - Thanks for the Pull Request.
We'll be reviewing and providing you feedback. If we have no comments, we'll be merging it directly.

ewertons:
Hi @pgonzalezJean , thanks for the PR.
By the time I got it assigned to me, I had done a fix for it already (commit 72d70b812f4ddf50eabce0238201bebe929790e3).
Thanks for reporting the issue.
If you don't mind, if you run into this scenario again could you confirm the fix works for you?
Thanks for using the Azure IoT SDKs.

pgonzalezJean:
Hi @ewertons,

As soon as we update the SDK on our devices, we will give you feedback about the issue.
Thanks!

PR #164, opened 2017-06-21
there's a dll where an exe should be...

darobs:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/tests/iothubclient_mqtt_ws_e2e/CMakeLists.txt
+++ b/iothub_client/tests/iothubclient_mqtt_ws_e2e/CMakeLists.txt
@@ -60,7 +60,7 @@ if(WIN32)
             aziotsharedutil
             rpcrt4
         )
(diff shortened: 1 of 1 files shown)


dcristoloveanu:
>         linkMqttLibrary(${theseTestsName}_dll)

I have to wonder why these are not passed with the ADDITIONAL_LIBS option of build_c_test_artifacts.
I guess this is so old that it was before the option was there...
Or maybe the option did not work?

If it would work this duplication would not exist and life would be rosier, wouldn't it :-)?

---
Refers to: iothub_client/tests/iothubclient_mqtt_ws_e2e/CMakeLists.txt:51 in 7f18d63. [](commit_id = 7f18d63e151b925e24598b33667509d15e94c8bc, deletion_comment = False)

dcristoloveanu:
:shipit:

Issue #79, opened 2017-02-27
Proxy option not set for file upload


AlexandreMarechalSE:
Hello,

On Linux, file upload does not use proxy set by IoTHubClient_SetOption(), but uses the system proxy which is set. Is there a way the file upload uses the proxy set for the client connection ?

Thanks.

Alexandre

tameraw:
@AlexandreMarechalSE - Apologies for the delay in response. This currently in our backlog; however, the support for this has not bubbled up at the top of the list in terms of priority. If interested in contributing, we'd happily review your Pull Request. Thanks.

kiranpradeep:
This has been unattended for 3 months. I don't see this as an enhancement. This is a blocker bug or crucial feature missing.

Consider a real world scenario of data acquisition from control systems in plant. In real world, plant engineers, won't allow public internet access in same plant network and so will need to use gateway with dual NICs. One NIC for plant network and another for public internet access. Setting proxy in environment at system level messes if we use HTTP on both NICs. Microsoft Azure proudly announces pre-configured solutions as "Connected Factory" for plants - But none of that will work in real world, unless this issue( not an enhancement ) is fixed.

tameraw:
@kiranpradeep - Understand your scenario. Very valid. We'll revise our priority for enabling this and get back to you with an ETA.

tameraw:
FYI - We'll pull this in and try to unblock you ASAP.

jebrando:
@kiranpradeep - Try setting the environment variable “http_proxy” in the format
"[protocol://][user:password@]machine[:port]" (e.g. https://user:password123@myproxy.domain.com:8080).  If you set this within a terminal it will only be set for executables started in that terminal.  We hope this can unblock you till we have a solution within the SDK.

kiranpradeep:
@jebrando We follow Azure/iot-edge architecture where in which each of component are written as in process modules([1](https://github.com/Azure/iot-edge/blob/master/core/devdoc/module.md)) and so don't have multiple executables. It is just one process with multiple modules as dlls.

bakrish:
@kiranpradeep Perhaps you can consider setenv(3) and unsetenv(3) to manage proxy environment variables within that specific module ? The proxy address could be kept in a configuration file. I guess this is not ideal, but could still help unblock this scenario for now. Bala

kiranpradeep:
[...]

PR #183, opened 2017-07-26
Allow file upload to use a web proxy

markrad:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/79

# Description of the problem
File upload only supports providing the web proxy information on Linux via the standard curl environment variables as in _export http_proxy=http://username:password@proxyname:port_. In the instance of a multi-homed device (as in one with multiple network cards) this does not provide any mechanism to specify the proxy for each network card.

# Description of the solution
Extended the existing option OPTION_HTTP_PROXY to pass this to the file upload logic too. This will cause the existing code to call the curl option that will set the web proxy for the file upload. _Note: Only HTTP proxies are currently supported and this only addresses the issue on Linux platforms._


iothub_client/src/iothub_client_ll_uploadtoblob.c
```diff
@@ -66,6 +67,10 @@ typedef struct IOTHUB_CLIENT_LL_UPLOADTOBLOB_HANDLE_DATA_TAG
         UPLOADTOBLOB_X509_CREDENTIALS x509credentials; /*assumed to be used when both deviceKey and deviceSasToken are NULL*/
     } credentials;                              /*needed for file upload*/
     char* certificates; /*if there are any certificates used*/
+	char* http_proxy_hostname;
+	int http_proxy_port;
+	char* http_proxy_username;
+	char* http_proxy_password;
```
jebrando: Would you want to add an HTTP_PROXY_OPTION object here instead of defining each variable?  You would have to do some casting from const char* to char*, but it may make it easier.

iothub_client/src/iothub_client_ll_uploadtoblob.c
```diff
@@ -1064,7 +1105,50 @@ IOTHUB_CLIENT_RESULT IoTHubClient_LL_UploadToBlob_SetOption(IOTHUB_CLIENT_LL_UPL
...
+				(proxy_options->username != proxy_options->password))
+			{
+				LogError("Only one of username and password for proxy settings was NULL");
+				result = IOTHUB_CLIENT_INVALID_ARG;
+			}
+			else
+			{
+				handleData->http_proxy_port = proxy_options->port;
+
+				if (mallocAndStrcpy_s(&handleData->http_proxy_hostname, proxy_options->host_address) != 0)
```
jebrando: >handleData->http_proxy_hostname [](start = 27, length = 31)

You need to free these first if they are already initialize from a previous call here.

iothub_client/src/iothub_client_ll_uploadtoblob.c
```diff
@@ -1064,7 +1105,50 @@ IOTHUB_CLIENT_RESULT IoTHubClient_LL_UploadToBlob_SetOption(IOTHUB_CLIENT_LL_UPL
...
+				{
+					LogError("failure in mallocAndStrcpy_s");
```
jebrando: >") [](start = 43, length = 2)

Can you add from http_proxy_hostname in this log.  If this fails it looks incomplete if you don't put the variable name that you're trying to mallocAndStrcpy

iothub_client/src/iothub_client_ll_uploadtoblob.c
jebrando: Here and all the char*.

---
In reply to: [130979180](https://github.com/Azure/azure-iot-sdk-c/pull/183#discussion_r130979180) [](ancestors = 130979180)

jebrando:
:shipit:

iothub_client/samples/iothub_client_sample_upload_to_blob/iothub_client_sample_upload_to_blob.c
```diff
@@ -57,18 +62,29 @@ void iothub_client_sample_upload_to_blob_run(void)
...
-            IoTHubClient_LL_Destroy(iotHubClientHandle);
+			else
+			{
+				HTTP_PROXY_OPTIONS http_proxy_options = {0};
+				http_proxy_options.host_address = proxyHost;
+				http_proxy_options.port = proxyPort;
+
+				if (proxyHost != NULL && IoTHubClient_LL_SetOption(iotHubClientHandle, OPTION_HTTP_PROXY, &http_proxy_options) != IOTHUB_CLIENT_OK)
+				{
+					(void)printf("failure to set proxy");
```
jebrando: \n

markrad:
Merged from another pull

PR #192, opened 2017-08-01
Fix Linux install for iothub_service_client

darobs:
Force a default for `CMAKE_INSTALL_INCLUDEDIR`, include GNUInstallDirs for service client install.

 

# Checklist
- [X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
- [X] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

# Description of the problem
Iot-Edge was not building - our build installs the C SDK libraries into an "install-deps" folder, but the service client was not installing there.  In Linux, it was attempting to install into `/azureiot` - `CMAKE_INSTALL_INCLUDEDIR` was not defined.

# Description of the solution

I included the GNUInstallDirs  into the service client CMakefile to guarantee the behavior the Edge wants, but I also set a default like we set a default for `CMAKE_INSTALL_LIBDIR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,6 +201,8 @@ else()
     set(THREAD_C_FILE ${SHARED_UTIL_ADAPTER_FOLDER}/threadapi_pthreads.c)
 endif()
 
+#Set CMAKE_INSTALL_LIBDIR if not defined
+include(GNUInstallDirs)
 
 if(NOT ${use_amqp} OR NOT ${use_http})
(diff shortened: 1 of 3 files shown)


darobs:
Confirms this is working for azure-iot-edge builds (installs the C-SDK binaries in the correct location).

zolvarga:
:shipit:

markrad:
:shipit:

Issue #196, opened 2017-08-05
Documentation about return type of C function of a method is wrong


maxpeng:
[serializer_apis.md](https://github.com/Azure/azure-iot-sdk-c/blob/master/serializer/devdoc/serializer_apis.md#with_methodmethodname-arg1type-arg1-arg2type-arg2-) describes the return type of a method is ```METHOD_RETURN_HANDLE```, but [methodreturn.h](https://github.com/Azure/azure-iot-sdk-c/blob/master/serializer/inc/methodreturn.h) defines the return type as ```METHODRETURN_HANDLE``` instead of ~~```METHOD_RETURN_HANDLE```~~.

Please correct the [serializer_apis.md](https://github.com/Azure/azure-iot-sdk-c/blob/master/serializer/devdoc/serializer_apis.md#with_methodmethodname-arg1type-arg1-arg2type-arg2-) to match the code. However ```METHOD_RETURN_HANDLE``` is more readable, so it could be better to update the code instead of documentation.

ewertons:
Hi @maxpeng,
thanks for pointing that out.
This item will not be addressed right away due to the priority (documentation), but you can expedite the fix if you would like by sending us a pull request, which we will take in right away.

Thanks,
Azure IoT Team

jspaith:
@kkmsft it appears that your pull requests has gone through successfully.  I'm closing this out, please re-open if I'm misreading this.  

Thanks for finding and fixing this issue for us!

PR #216, opened 2017-08-22
Fix the serializer api doc - #196

kkmsft:


kkmsft:
@ewertons: Was wondering any comments on this doc change ?

PR #232, opened 2017-09-15
Cross compile linux armhf

JuergenKosel:
# Checklist
- [ X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
  - [ X] This pull-request is submitted against the `master` branch. 

# Description of the problem

# Description of the solution
 
This pull request adds sample scripts to cross build the azure-iot-sdk-c for Linux-armhf. I have tested it successfully with our embedded Debian 8 armhf device.
This pull request also adds install rules for some of the existing sample applications.
The pull request needs an update of the c-utility sub-module (see https://github.com/Azure/azure-c-shared-utility/pull/96 ).


iothub_client/samples/iothub_client_sample_amqp/iothub_client_sample_amqp.c
```diff
@@ -228,8 +228,9 @@ void iothub_client_sample_amqp_run(void)
     }
 }
 
-int main(void)
+int main(int argc, char* const argv[])
 {
+    if (argc > 1) connectionString = argv[1];
```
ewertons: Hi @JuergenKosel,

for the samples we do not expose getting the arguments from command line.
Let's keep the standard used across the other samples and read it explicitly from the static const char* connectionString variable.

iothub_client/samples/iothub_client_sample_http/iothub_client_sample_http.c
```diff
@@ -239,8 +239,9 @@ void iothub_client_sample_http_run(void)
     }
 }
 
-int main(void)
+int main(int argc, char* const argv[])
 {
+   if (argc > 1) connectionString = argv[1];
```
ewertons: for the samples we do not expose getting the arguments from command line.
Let's keep the standard used across the other samples and read it explicitly from the static const char* connectionString variable.

iothub_client/samples/iothub_client_sample_mqtt/iothub_client_sample_mqtt.c
```diff
@@ -213,8 +213,9 @@ void iothub_client_sample_mqtt_run(void)
     }
 }
 
-int main(void)
+int main(int argc, char* const argv[])
```
ewertons: for the samples we do not expose getting the arguments from command line.
Let's keep the standard used across the other samples and read it explicitly from the static const char* connectionString variable.

ewertons:
I added @JetstreamRoySprowl and @mamokarz to give more insights about the cross compilation part.

Thanks for the PR!

iothub_client/samples/iothub_client_sample_mqtt/iothub_client_sample_mqtt.c
JuergenKosel: It has some disadvantages, if the connection string must be hard coded in the sources.
Especially, if the host which runs the binaries is not identical with the build machine which could run the cross-compiler.
Therefore I have added this changes.

ewertons:
Hello @JuergenKosel, we reviewed the code and although it is a really good clean change, it is actually not aligning with our next strategy for supporting new devices and platforms. The new strategy will be rolled out soon, so we would need to change/remove your code again.
But there is an alternative for the platform you are covering, we have a solution called Device Catalog, where devices/platforms can be registered with details of how to port the Azure IoT SDK to them. We believe that would be the best option for the moment to capture the good work you have done for armhf.

I will post more details about how to use Device Catalog briefly.

ewertons:
Hi @JuergenKosel ,

You can check the device catalog page to have an idea of how it works:
https://catalog.azureiotsuite.com/

It takes a few steps to get the instructions saved as a supported platform, but it is really worth since your contribution would be made available to all users. 
You will need a Microsoft account (e.g., @outlook.com) to create a profile, and then you can follow the instructions here to add the platform as a supported entity: 
https://catalogstorageprod.blob.core.windows.net/devicecatalogimageserver/Device_catalog_onboarding_instructions.pdf

We will gladly help you through the steps if you have any questions.

PR #233, opened 2017-09-20
Add version extractor function to build and call from root

markrad:
# Checklist

CSS customer issue.

# Description of the problem
Currently the version number of the SDK is hard coded in the CMake root file.

# Description of the solution
Added a function to parse the version number from the appropriate header file and called it from the root CMake file. Removes the possibility of the versions inadvertently coming out of sync. This is a will be more important when I check in the fix to add the appropriate version numbers to the .so files. Without the version numbers it causes an issue on Yocto system builds.


CMakeLists.txt
```diff
@@ -4,7 +4,10 @@
 cmake_minimum_required(VERSION 2.8.11)
 project(azure_iot_sdks)
 
-set(IOT_SDK_VERSION 1.1.23)
```
jebrando: This may require a release script update.  Please follow up on this so it doesn't get found at the last moment.

configs/azure_iot_sdksFunctions.cmake
```diff
@@ -1,6 +1,30 @@
...
 #Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+function(getIoTSDKVersion)
+	# First find the applicable line in the file
+	file (STRINGS "${CMAKE_SOURCE_DIR}/iothub_client/inc/iothub_client_version.h" iotsdkverstr
+		REGEX "^[\t ]*#[\t ]*define[\t ]*IOTHUB_SDK_VERSION[\t ]*\"([0-9]+)[.]([0-9]+)[.]([0-9]+)\"")
+		
+	if (!MATCHES)
+		message(FATAL_ERROR "Unable to find version in ${CMAKE_SOURCE_DIR}/iothub_client/inc/iothub_client_version.h")
+	else(!MATCHES)
```
jebrando: This seem peculiar to me the if and the else both have !MATCHES.  I think you want the else just to haven and empty () after it.

CMakeLists.txt
```diff
@@ -4,7 +4,10 @@
...
+include("configs/azure_iot_sdksFunctions.cmake")
+
+getIoTSDKVersion()
+message("IoT Client SDK Version = ${IOT_SDK_VERSION}")
```
ewertons: Please use message(STATUS ... to match the other statements.

configs/azure_iot_sdksFunctions.cmake
markrad: CMake allows one to put the condition statement on the else, and endif. In my opinion this is useful because it will match these up and if one, for example, misses an endif in a later code modification it will flag the mismatch of the conditional block's start and end making the omission easier to find. However, if you really hate it, I'll take it out. Your call.

---
In reply to: [140001036](https://github.com/Azure/azure-iot-sdk-c/pull/233#discussion_r140001036) [](ancestors = 140001036)

PR #237, opened 2017-09-21
AMQP transport to process C2D messages with UUID/ULONG in message-id …

ewertons:
…and correlation-id

 

# Checklist
- [x ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/src/uamqp_messaging.c
+++ b/iothub_client/src/uamqp_messaging.c
@@ -2,14 +2,18 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #include <stdlib.h>
 #include "uamqp_messaging.h"
+#include "azure_c_shared_utility/gballoc.h"
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/xlogging.h"
+#include "azure_c_shared_utility/uuid.h"
 #include "azure_uamqp_c/message.h"
 #include "azure_uamqp_c/amqpvalue.h"
 #include "iothub_message.h"
 #ifndef RESULT_OK
 #define RESULT_OK 0
 #endif
+
+#define UUID_N_OF_OCTECTS 16
 
 static int encode_callback(void* context, const unsigned char* bytes, size_t length)
 {
@@ -430,10 +434,9 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
 
 static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_handle, MESSAGE_HANDLE uamqp_message)
 {
-    int return_value;
+    int result;
     PROPERTIES_HANDLE uamqp_message_properties;
     AMQP_VALUE uamqp_message_property;
-    const char* uamqp_message_property_value;
     int api_call_result;
 
     // Codes_SRS_UAMQP_MESSAGING_09_008: [The uAMQP message properties shall be retrieved using message_get_properties().]
@@ -441,11 +444,11 @@ static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_h
     {
         // Codes_SRS_UAMQP_MESSAGING_09_009: [If message_get_properties() fails, message_create_IoTHubMessage_from_uamqp_message shall fail and return immediately.]
         LogError("Failed to get property properties map from uAMQP message (error code %d).", api_call_result);
-        return_value = __FAILURE__;
+        result = __FAILURE__;
     }
     else
     {
-        return_value = 0; // Properties 'message-id' and 'correlation-id' are optional according to the AMQP 1.0 spec.
+        result = 0; // Properties 'message-id' and 'correlation-id' are optional according to the AMQP 1.0 spec.
 
         // Codes_SRS_UAMQP_MESSAGING_09_010: [The message-id property shall be read from the uAMQP message by calling properties_get_message_id.]
         if ((api_call_result = properties_get_message_id(uamqp_message_properties, &uamqp_message_property)) != 0)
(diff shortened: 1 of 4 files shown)


iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
     else
     {
-        return_value = 0; // Properties 'message-id' and 'correlation-id' are optional according to the AMQP 1.0 spec.
+        result = 0; // Properties 'message-id' and 'correlation-id' are optional according to the AMQP 1.0 spec.
 
         // Codes_SRS_UAMQP_MESSAGING_09_010: [The message-id property shall be read from the uAMQP message by calling properties_get_message_id.]
         if ((api_call_result = properties_get_message_id(uamqp_message_properties, &uamqp_message_property)) != 0)
         {
             // Codes_SRS_UAMQP_MESSAGING_09_011: [If properties_get_message_id fails, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.]
             LogInfo("Failed to get value of uAMQP message 'message-id' property (%d). No failure, since it is optional.", api_call_result);
```
jebrando: >LogInfo("Failed to get value of uAMQP message 'message-id' property (%d). No failure, since it is optional.", api_call_result); [](start = 12, length = 127)

If this is actually optional do we need to logInfo?

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
+                    // Codes_SRS_UAMQP_MESSAGING_09_014: [The message-id value shall be retrieved from the AMQP_VALUE as char sequence] 
+                    if (amqpvalue_get_ulong(uamqp_message_property, &ulong_value) != 0)
+                    {
+                        // Codes_SRS_UAMQP_MESSAGING_09_015: [If message-id fails to be obtained, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.] 
+                        LogError("Failed to get value of uAMQP message 'message-id' property (ulong)");
+                        result = __FAILURE__;
+                    }
+                    else
+                    {
+                        (void)sprintf(string_buffer, "%llu", ulong_value);
```
jebrando: >(void)sprintf(string_buffer, "%llu", ulong_value); [](start = 24, length = 50)

This can fail.

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
-                LogError("Failed to get value of uAMQP message 'message-id' property (%d).", api_call_result);
-                return_value = __FAILURE__;
-            }
-            // Codes_SRS_UAMQP_MESSAGING_09_016: [The message-id property shall be set on the IOTHUB_MESSAGE_HANDLE instance by calling IoTHubMessage_SetMessageId(), passing the value read from the uAMQP message.]
-            else if (IoTHubMessage_SetMessageId(iothub_message_handle, uamqp_message_property_value) != IOTHUB_MESSAGE_OK)
+            // Codes_SRS_UAMQP_MESSAGING_09_012: [The type of the message-id property value shall be obtained using amqpvalue_get_type().]
+            AMQP_TYPE value_type = amqpvalue_get_type(uamqp_message_property);
+
+            // Codes_SRS_UAMQP_MESSAGING_09_013: [If the type of the message-id property value is AMQP_TYPE_NULL, message_create_IoTHubMessage_from_uamqp_message() shall skip processing the message-id (as it is optional) and continue normally.]
+            if (value_type != AMQP_TYPE_NULL)
```
zolvarga: > if (value_type != AMQP_TYPE_NULL) [](start = 11, length = 34)

There is no handling for the "else". Probably we need to (at least)

iothub_client/src/uamqp_messaging.c
```diff
@@ -479,30 +539,87 @@ static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_h
...
-                LogError("Failed to get value of uAMQP message 'correlation-id' property (%d).", api_call_result);
-                return_value = __FAILURE__;
-            }
-            // Codes_SRS_UAMQP_MESSAGING_09_024: [The correlation-id property shall be set on the IOTHUB_MESSAGE_HANDLE by calling IoTHubMessage_SetCorrelationId, passing the value read from the uAMQP message.]
-            else if (IoTHubMessage_SetCorrelationId(iothub_message_handle, uamqp_message_property_value) != IOTHUB_MESSAGE_OK)
+            // Codes_SRS_UAMQP_MESSAGING_09_020: [The type of the correlation-id property value shall be obtained using amqpvalue_get_type().]
+            AMQP_TYPE value_type = amqpvalue_get_type(uamqp_message_property);
+
+            // Codes_SRS_UAMQP_MESSAGING_09_021: [If the type of the correlation-id property value is AMQP_TYPE_NULL, IoTHubMessage_CreateFromuAMQPMessage() shall skip processing the correlation-id (as it is optional) and continue normally.] 
+            if (value_type != AMQP_TYPE_NULL)
```
zolvarga: >if (value_type != AMQP_TYPE_NULL) [](start = 12, length = 33)

Same... "else" is missing....

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
+            AMQP_TYPE value_type = amqpvalue_get_type(uamqp_message_property);
+
+            // Codes_SRS_UAMQP_MESSAGING_09_013: [If the type of the message-id property value is AMQP_TYPE_NULL, message_create_IoTHubMessage_from_uamqp_message() shall skip processing the message-id (as it is optional) and continue normally.]
+            if (value_type != AMQP_TYPE_NULL)
             {
-                // Codes_SRS_UAMQP_MESSAGING_09_017: [If IoTHubMessage_SetMessageId fails, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.]
-                LogError("Failed to set IOTHUB_MESSAGE_HANDLE 'message-id' property.");
-                return_value = __FAILURE__;
+
+                char* string_value = NULL;
```
zolvarga: >char* string_value = NULL; [](start = 16, length = 26)

Do not initialize it here to have clear flow how the value gets initialized.

iothub_client/src/uamqp_messaging.c
```diff
@@ -479,30 +539,87 @@ static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_h
...
+            // Codes_SRS_UAMQP_MESSAGING_09_020: [The type of the correlation-id property value shall be obtained using amqpvalue_get_type().]
+            AMQP_TYPE value_type = amqpvalue_get_type(uamqp_message_property);
+
+            // Codes_SRS_UAMQP_MESSAGING_09_021: [If the type of the correlation-id property value is AMQP_TYPE_NULL, IoTHubMessage_CreateFromuAMQPMessage() shall skip processing the correlation-id (as it is optional) and continue normally.] 
+            if (value_type != AMQP_TYPE_NULL)
             {
-                // Codes_SRS_UAMQP_MESSAGING_09_025: [If IoTHubMessage_SetCorrelationId fails, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.]
-                LogError("Failed to set IOTHUB_MESSAGE_HANDLE 'correlation-id' property.");
-                return_value = __FAILURE__;
+                char* string_value = NULL;
```
zolvarga: >char* string_value = NULL [](start = 16, length = 25)

Same initialization comment as before

iothub_client/src/uamqp_messaging.c
ewertons: Yes, we do.
When we try to get the value, the function can succeed and return NULL (because it's optional) and then in that case we are ok, there is no error to log.
But the function can also fail (return non-zero), and that case is unexpected, so we need to treat as a failure.

iothub_client/src/uamqp_messaging.c
ewertons: Amen! Shall be fixed!

iothub_client/src/uamqp_messaging.c
ewertons: I'll add the elses.

iothub_client/src/uamqp_messaging.c
ewertons: I'll add the elses.

iothub_client/src/uamqp_messaging.c
ewertons: I'll change that.

iothub_client/src/uamqp_messaging.c
ewertons: Shall be fixed

iothub_client/src/uamqp_messaging.c
ewertons: I take that back. Sorry for the misunderstanding. Fixed now.

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
+
+            // Codes_SRS_UAMQP_MESSAGING_09_013: [If the type of the message-id property value is AMQP_TYPE_NULL, message_create_IoTHubMessage_from_uamqp_message() shall skip processing the message-id (as it is optional) and continue normally.]
+            if (value_type != AMQP_TYPE_NULL)
             {
-                // Codes_SRS_UAMQP_MESSAGING_09_017: [If IoTHubMessage_SetMessageId fails, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.]
-                LogError("Failed to set IOTHUB_MESSAGE_HANDLE 'message-id' property.");
-                return_value = __FAILURE__;
+
+                char* string_value = NULL;
+                char string_buffer[40];
```
jspaith: >40 [](start = 35, length = 2)

Magic # in code.  Ideally this and the UUID_N_OCTET stuff would be in c-utils but I get the submodule dance needed here.

ewertons: I'll add a define. This can either be a number, a string or an UUID.
So I'm arbitrarily changing the cap to 128 with:
#define MESSAGE_ID_MAX_SIZE 128

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
+                    }
+                    else if ((string_value = UUID_to_string((UUID*)uuid_value)) == NULL)
+                    {
+                        // Codes_SRS_UAMQP_MESSAGING_09_015: [If message-id fails to be obtained, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.] 
+                        LogError("Failed to get the string representation of 'message-id' UUID");
+                        result = __FAILURE__;
+                    }
+                    else
+                    {
+                        (void)memcpy(string_buffer, string_value, sizeof(string_value));
```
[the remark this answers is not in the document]
jspaith: Note though my comment RE breaking this up into sub functions, this may not be as relevant(?) since I wrote before the feedback at head of this fcn.

---
In reply to: [140559919](https://github.com/Azure/azure-iot-sdk-c/pull/237#discussion_r140559919) [](ancestors = 140559919)

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,46 +434,102 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
 
 static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_handle, MESSAGE_HANDLE uamqp_message)
```
jspaith: >readPropertiesFromuAMQPMessage [](start = 11, length = 30)

This fcn is +250 lines which is getting unweildy.  

Could we please have a helper function that would do something like a GetStringVariant (please don't call it a Variant though :)) where you'd do llike GetStringVarianr(property_name, string_buffer, string_buffer_length) and then it'd do the str/long/uuid determination.

iothub_client/src/uamqp_messaging.c
```diff
@@ -2,15 +2,19 @@
...
 #include "azure_c_shared_utility/xlogging.h"
+#include "azure_c_shared_utility/uuid.h"
 #include "azure_uamqp_c/message.h"
 #include "azure_uamqp_c/amqpvalue.h"
 #include "iothub_message.h"
 #ifndef RESULT_OK
 #define RESULT_OK 0
 #endif
 
+#define UUID_N_OF_OCTECTS 16
```
dcristoloveanu: >OCTECT [](start = 18, length = 6)

Octects!

dcristoloveanu: I think I saw the same typo earlier in another review. :-)

---
In reply to: [140561235](https://github.com/Azure/azure-iot-sdk-c/pull/237#discussion_r140561235) [](ancestors = 140561235)

iothub_client/src/uamqp_messaging.c
dcristoloveanu: >string_value [](start = 22, length = 12)

const

iothub_client/src/uamqp_messaging.c
dcristoloveanu: >ulong_value [](start = 61, length = 11)

This is not an unsigned long long, it is an uint64_t. PRIu64 shall be used.

iothub_client/src/uamqp_messaging.c
```diff
@@ -2,15 +2,19 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #include <stdlib.h>
```
dcristoloveanu: Missing stdint.h for fixed size types
And probably inttypes.h since I can see you want to print fixed size types for errors.

iothub_client/src/uamqp_messaging.c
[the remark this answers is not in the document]
ewertons: Makes sense. I'll take option B

iothub_client/src/uamqp_messaging.c
```diff
@@ -479,30 +539,87 @@ static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_h
...
+            AMQP_TYPE value_type = amqpvalue_get_type(uamqp_message_property);
+
+            // Codes_SRS_UAMQP_MESSAGING_09_021: [If the type of the correlation-id property value is AMQP_TYPE_NULL, IoTHubMessage_CreateFromuAMQPMessage() shall skip processing the correlation-id (as it is optional) and continue normally.] 
+            if (value_type != AMQP_TYPE_NULL)
             {
-                // Codes_SRS_UAMQP_MESSAGING_09_025: [If IoTHubMessage_SetCorrelationId fails, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.]
-                LogError("Failed to set IOTHUB_MESSAGE_HANDLE 'correlation-id' property.");
-                return_value = __FAILURE__;
+                char* string_value = NULL;
+                char string_buffer[40];
```
dcristoloveanu: >40 [](start = 35, length = 2)

I would add a comment here with the computation of why exactly 40. SO that future monkeys can know why not 39.

iothub_client/src/uamqp_messaging.c
```diff
@@ -479,30 +539,87 @@ static int readPropertiesFromuAMQPMessage(IOTHUB_MESSAGE_HANDLE iothub_message_h
...
+                }
+                else if (value_type == AMQP_TYPE_ULONG)
+                {
+                    uint64_t ulong_value;
+
+                    // Codes_SRS_UAMQP_MESSAGING_09_022: [The correlation-id value shall be retrieved from the AMQP_VALUE as char sequence] 
+                    if (amqpvalue_get_ulong(uamqp_message_property, &ulong_value) != 0)
+                    {
+                        // Codes_SRS_UAMQP_MESSAGING_09_023: [If correlation-id fails to be obtained, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.] 
+                        LogError("Failed to get value of uAMQP message 'correlation-id' property (ulong)");
```
dcristoloveanu: >ulong [](start = 98, length = 5)

uint64_t

iothub_client/src/uamqp_messaging.c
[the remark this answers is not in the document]
ewertons: I'll take option A.

This is the logic:
When the message-id is:
STRING => string_value is a const pointer to a char sequence held by uamqp
ULONG => string_value is a pointer to string_buffer
UUID => string_value is a new string that we need to free after using.

So the if check you are proposing is not sufficient to define if we should free or not.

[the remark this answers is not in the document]
ewertons: Fixed to:
                        (void)memcpy(string_buffer, string_value, sizeof(char) * strlen(string_value));

iothub_client/src/uamqp_messaging.c
ewertons: You are correct, but this is an improvement. It will be done in another time.
I'll go ahead with the check in, we need to kick the release today.

iothub_client/src/uamqp_messaging.c
ewertons: This define is gone. Not used.

iothub_client/src/uamqp_messaging.c
ewertons: Done

iothub_client/src/uamqp_messaging.c
ewertons: Done

iothub_client/src/uamqp_messaging.c
ewertons: Done

iothub_client/src/uamqp_messaging.c
ewertons: Added a define on top

iothub_client/src/uamqp_messaging.c
ewertons: I'll keep as is since it is mentioned as amqpvalue_get_ulong

iothub_client/src/uamqp_messaging.c
```diff
@@ -430,79 +443,219 @@ int message_create_uamqp_encoding_from_iothub_message(IOTHUB_MESSAGE_HANDLE mess
...
+                    else if ((string_value = UUID_to_string((UUID*)uuid_value)) == NULL)
+                    {
+                        // Codes_SRS_UAMQP_MESSAGING_09_015: [If message-id fails to be obtained, message_create_IoTHubMessage_from_uamqp_message() shall fail and return immediately.] 
+                        LogError("Failed to get the string representation of 'message-id' UUID");
+                        string_value = NULL;
+                        result = __FAILURE__;
+                    }
+                    else
+                    {
+                        (void)memcpy(string_buffer, string_value, sizeof(char) * strlen(string_value));
```
jspaith: >memcpy [](start = 30, length = 6)

Also - why is this a memcpy?  Shouldn't it just be a straight strcpy at this point?

ewertons: Strcpy goes pretty well when I don’t want to deal or don’t know the size of the string, and I’m counting on the '\0' as terminator.

Since here I know the size of the string, memcpy is preferable as behind the scenes the function doesn’t need to be searching for the ‘\0’ throughout the char sequence.

Plus, there are lots of talks about how strcpy can be “evil”. So I advocate to keep memcpy.

ewertons:
Thanks everyone

Issue #254, opened 2017-10-11
blob upload fails over consistently http proxy on Linux/macOS


kiranpradeep:
- **OS and version used:** macOS 10.12.3, Ubuntu 14.04 LTS, Ubuntu 16.04 LTS ( i believe everywhere )
- **SDK version used:**  1.1.24

# Description of the issue:
This is a continuation from #79. I tried setting proxy host and port on macOS and Ubuntu. Blob upload fails. Console log below. Blob upload succeeds when we try without proxy( and so not a https certificate issue ).

The sample still continues to work, if we set HTTP proxy in the environmental variable as before.

# Code sample exhibiting the issue:
Blob upload sample is used as is([link](https://github.com/Azure/azure-iot-sdk-c/blob/2017-09-25/iothub_client/samples/iothub_client_sample_upload_to_blob/iothub_client_sample_upload_to_blob.c)). Only edited the _connectionString_, _proxyHost_, and _proxyPort_ variables.

# Console log of the issue:
```
Info: IoT Hub SDK for C, version 1.1.24
Error: Time:Wed Oct 11 16:28:05 2017 File:/Users/779591ardp/Downloads/azure-iot-sdk-c/c-utility/adapters/httpapi_curl.c Func:HTTPAPI_SetOption Line:794 failure in curl_easy_setopt - CURLOPT_SSL_CTX_FUNCTION
Error: Time:Wed Oct 11 16:28:05 2017 File:/Users/779591ardp/Downloads/azure-iot-sdk-c/c-utility/src/httpapiex.c Func:HTTPAPIEX_ExecuteRequest Line:400 HTTPAPI_SetOption failed when called for option TrustedCerts
Error: Time:Wed Oct 11 16:28:05 2017 File:/Users/779591ardp/Downloads/azure-iot-sdk-c/c-utility/adapters/httpapi_curl.c Func:HTTPAPI_SetOption Line:794 failure in curl_easy_setopt - CURLOPT_SSL_CTX_FUNCTION
Error: Time:Wed Oct 11 16:28:05 2017 File:/Users/779591ardp/Downloads/azure-iot-sdk-c/c-utility/src/httpapiex.c Func:HTTPAPIEX_ExecuteRequest Line:400 HTTPAPI_SetOption failed when called for option TrustedCerts
[...]

PR #267, opened 2017-10-19
Fix bug in file upload over proxy

markrad:
# Checklist

Fixes https://github.com/Azure/azure-iot-sdk-c/issues/254

# Description of the problem
The file upload consists of two round trips, one to each of two web servers. The first round trip was correctly using the web proxy whereas the second round trip did not. This was due to the fact that Blob_UploadFromSasUri was not passed the proxy information so consequently it did not set the appropriate option for the lower level.

# Description of the solution
Added an additional parameter to Blob_UploadFromSasUri that contains the proxy information. If there is no proxy information then this can be either passed as NULL or the host address in the proxy structure can be set to NULL. In either case the code will bypass using a web proxy for the upload.

### DIFF
--- a/iothub_client/inc/blob.h
+++ b/iothub_client/inc/blob.h
@@ -15,6 +15,7 @@
 
 #include "azure_c_shared_utility/macro_utils.h"
 #include "azure_c_shared_utility/buffer_.h"
+#include "azure_c_shared_utility/shared_util_options.h"
 
 #ifdef __cplusplus
 #include <cstddef>
@@ -44,10 +45,11 @@ DEFINE_ENUM(BLOB_RESULT, BLOB_RESULT_VALUES)
 * @param    httpStatus      A pointer to an out argument receiving the HTTP status (available only when the return value is BLOB_OK)
 * @param    httpResponse    A BUFFER_HANDLE that receives the HTTP response from the server (available only when the return value is BLOB_OK)
 * @param    certificates    A null terminated string containing CA certificates to be used
+* @param    proxyOptions    A structure that contains optional web proxy information
 *
 * @return	A @c BLOB_RESULT. BLOB_OK means the blob has been uploaded successfully. Any other value indicates an error
 */
-MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadFromSasUri,const char*, SASURI, const unsigned char*, source, size_t, size, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates)
+MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadFromSasUri,const char*, SASURI, const unsigned char*, source, size_t, size, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates, HTTP_PROXY_OPTIONS, *proxyOptions)
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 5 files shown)


tameraw:
@jspaith is added to the review. #Closed

iothub_client/inc/blob.h
```diff
@@ -44,10 +45,11 @@ DEFINE_ENUM(BLOB_RESULT, BLOB_RESULT_VALUES)
 * @param    httpStatus      A pointer to an out argument receiving the HTTP status (available only when the return value is BLOB_OK)
 * @param    httpResponse    A BUFFER_HANDLE that receives the HTTP response from the server (available only when the return value is BLOB_OK)
 * @param    certificates    A null terminated string containing CA certificates to be used
+* @param    proxyOptions    A structure that contains optional web proxy information
 *
 * @return	A @c BLOB_RESULT. BLOB_OK means the blob has been uploaded successfully. Any other value indicates an error
 */
-MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadFromSasUri,const char*, SASURI, const unsigned char*, source, size_t, size, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates)
+MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadFromSasUri,const char*, SASURI, const unsigned char*, source, size_t, size, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates, HTTP_PROXY_OPTIONS*, proxyOptions)
```
tameraw: >HTTP_PROXY_OPTIONS*, proxyOptions [](start = 203, length = 33)

Can we look at passing this thru SetOption()?

iothub_client/tests/blob_ut/blob_ut.c
```diff
@@ -334,7 +335,66 @@ TEST_FUNCTION(Blob_UploadFromSasUri_with_certificates_happy_path)
...
+{
+    ///arrange
+    unsigned char c = '3';
+	HTTP_PROXY_OPTIONS proxyOptions = { "a", 8888, NULL, NULL };
+	
+    STRICT_EXPECTED_CALL(gballoc_malloc(strlen(TEST_HOSTNAME_1) + 1));
+    {
+        STRICT_EXPECTED_CALL(HTTPAPIEX_Create(TEST_HOSTNAME_1));
+        {
+            STRICT_EXPECTED_CALL(HTTPAPIEX_SetOption(IGNORED_PTR_ARG, OPTION_HTTP_PROXY, IGNORED_PTR_ARG));
```
tameraw: >IGNORED_PTR_ARG [](start = 89, length = 15)

Why not check against the proxyOptions?

markrad: Typo. That's what it should do.

---
In reply to: [150113572](https://github.com/Azure/azure-iot-sdk-c/pull/267#discussion_r150113572) [](ancestors = 150113572)

iothub_client/inc/blob.h
markrad: You do realize that this value comes from a SetOption call at the layer above? Unfortunately the options are not neatly packaged in a manner that would make it easy to pass them to this code. It would create a dependency on the layout of what is effectively an opaque structure.

---
In reply to: [149761106](https://github.com/Azure/azure-iot-sdk-c/pull/267#discussion_r149761106) [](ancestors = 149761106)

PR #300, opened 2017-11-16
Add docker support for module sender sample

avranju:
Add Docker support for the edge module sender sample so that we can build an image for this sample and test with Azure IoT Edge.

### DIFF
--- a/iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
+++ b/iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
@@ -23,13 +23,13 @@
 
 /*String containing Hostname, Device Id & Device Key, ModuleID, and GatewayHostName in the format:                          */
 /*  "HostName=<host_name>;DeviceId=<device_id>;SharedAccessKey=<device_key>;ModuleId=<Module_Id>;GatewayHostName=127.0.0.1" */
-static const char* connectionString = "[device connection string]";
+static char* connectionString = "[device connection string]";
 
 static int callbackCounter;
 static char msgText[1024];
 static char propText[1024];
 static bool g_continueRunning;
-#define MESSAGE_COUNT 5
+#define MESSAGE_COUNT 500
 #define DOWORK_LOOP_NUM     3
 
 
@@ -127,8 +127,13 @@ static void SendConfirmationCallback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, v
 void iothub_client_sample_module_sender(void)
 {
     IOTHUB_CLIENT_LL_HANDLE iotHubClientHandle;
-
     EVENT_INSTANCE messages[MESSAGE_COUNT];
+
+    char *envCS = getenv("EdgeHubConnectionString");
+    if (envCS != NULL)
+    {
+        connectionString = envCS;
+    }
 
     g_continueRunning = true;
     srand((unsigned int)time(NULL));
@@ -212,7 +217,7 @@ void iothub_client_sample_module_sender(void)
 
                     }
                     IoTHubClient_LL_DoWork(iotHubClientHandle);
-                    ThreadAPI_Sleep(1);
+                    ThreadAPI_Sleep(1000);
 
                     iterator++;
                 } while (g_continueRunning);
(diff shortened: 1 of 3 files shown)


kiranpradeep:
Sorry for jumping the gun. But the corresponding C# filter module([1](https://docs.microsoft.com/en-us/azure/iot-edge/tutorial-csharp-module)) takes threshold from module twin. Looking for a C sample which could read from module twin.

jspaith:
@kiranpradeep Please open a github issue to track the request. We'll close the pull request once it gets merged up and I don't want to lose your feedback here. Thanks

jspaith:
Oh yes @kiranpradeep, just to be more clear Raj's scope here is limited to "docker-ification" and limited changes to play nice in Edge which is why i asked for different item tracking your ask.  JS

avranju:
@jspaith For some reason I don't see your review comments on the PR. But since we discussed offline I submitted the changes you requested last night.

kiranpradeep:
@jspaith Thanks. Would like your guidance on which GitHub repo to file these requests under( these are not real issues but more user requests ). Is it iot-edge([1](https://github.com/Azure/iot-edge)), azure-iot-sdk-c([2](https://github.com/Azure/azure-iot-sdk-c)) or iot-edge suggestions page([3](https://feedback.azure.com/forums/907045-azure-iot-edge))? In fact, I was suprised to see docker files going into azure-iot-sdk-c rather than iot-edge.

iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
```diff
@@ -23,13 +23,13 @@
 
 /*String containing Hostname, Device Id & Device Key, ModuleID, and GatewayHostName in the format:                          */
 /*  "HostName=<host_name>;DeviceId=<device_id>;SharedAccessKey=<device_key>;ModuleId=<Module_Id>;GatewayHostName=127.0.0.1" */
-static const char* connectionString = "[device connection string]";
+static char* connectionString = "[device connection string]";
```
jspaith: Please leave this as a 'const char*'.  This means the value cannot be changed, but the pointer can, which is what we want.

iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
```diff
@@ -127,9 +127,14 @@ static void SendConfirmationCallback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, v
 void iothub_client_sample_module_sender(void)
 {
     IOTHUB_CLIENT_LL_HANDLE iotHubClientHandle;
-
     EVENT_INSTANCE messages[MESSAGE_COUNT];
 
+    char *envCS = getenv("EdgeHubConnectionString");
```
jspaith: We tend to not use abbreviations for this, so connectionStringFromEnvironment would be closer.

iothub_client/samples/iothub_client_sample_module_sender/start.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jspaith: This should probably also be under the "./linux" directory.  Unless we are tying into the "bash on Windows" subsystem stuff for Win10 or assuming a Cygwin or other type Unix'y environment on Win, this looks pretty Linux.

jspaith:
@avranju - I have a few more tweaks but in general looks good, once you address I can get checkin process going.

jspaith:
@kiranpradeep - for Edge wide suggestions / feedback - either for Edge Hub / Agent etc. or general programming model questions across all SDK's - please continue using one of the existing Edge channels.

If it's a request where the changes are going into the C SDK - e.g. the sample JSON you've made - it's better to go here.  I don't check the Edge so someone has to forward it to me, but I'll see C requests here and be able to react faster.

RE where to checkin the Docker files, historically when I've seen code separated from manifest then they tend to get out of alignment as someone updates one but not the other.  So by keeping Docker setup close to the actual sample code, less likely.  But for this case, if there's something different I'm not seeing I'm open to feedback.

iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
avranju: Sure.

iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
avranju: :) I like small variable names but that's me. Changed it to `connectionStringFromEnvironment`.

iothub_client/samples/iothub_client_sample_module_sender/start.sh
avranju: Its already in a folder called `docker\linux\amd64`. We should probably make a `Dockerfile` for ARM too. I didn't have the time to setup and test on a Pi.

PR #324, opened 2017-12-11
Add MQTT WS Transport header files to install-target

pboettch:
# Description of the problem

After running `make install` the file `iothubtransportmqtt_websockets.h` is not installed. However, it is needed to write a MQTT-over-WebSockets-client.

# Description of the solution

Add the file to the list of files to be installed (as it is already done with `iothubtransportmqtt.h`).

### DIFF
--- a/iothub_client/CMakeLists.txt
+++ b/iothub_client/CMakeLists.txt
@@ -207,6 +207,7 @@ if(${use_mqtt})
     set(iothub_client_h_install_files
         ${iothub_client_h_install_files}
         ${iothub_client_mqtt_transport_h_files}
+        ${iothub_client_mqtt_ws_transport_h_files}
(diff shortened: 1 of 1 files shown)


ewertons:
Thank you very much for this contribution! 
We will get it merged right away.

ewertons:
And it's done. 
Please verify on the latest code from master, and let us know if we can help in any way.
Thanks,
Azure IoT Team

pboettch:
Thanks. I'm wondering why the CLA-checker is happy, I never did anything - to my knowledge.

ewertons:
That is a good question. I am not sure why.
I have seen it pre-approving PRs before with changes that are not directly in source code files (like scripts, build config files, etc).

PR #347, opened 2018-01-08
Use nproc from coreutils to determine no. of CPUs

ftheile:
It's a neat tool, but even GregKH didn't know it (see https://plus.google.com/u/0/+gregkroahhartman/posts/SLHEeDZFeRm) ;-)


jspaith:
Thanks for this request, @ftheile.  It turns out not all distros have nproc which is why we have this as a lowest common denominator.

Was there a reason (beyond obviously being cleaner) you'd wanted this change?  If you're hitting problems let us know and we'll see if there's some other way of approaching.

Thanks again

ftheile:
Hi @jspaith, OK, I understand. Feel free to ignore my PR.

PR #365, opened 2018-01-27
Fixing Subject Alternative Name in generated certificates.

### LINKED ISSUE #362: Certificate generator generates invalid Subject Alternative Name client certificates
In https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/ca-certs.ps1#L57 `$subjectName` should not contain "CN=". This produces incorrect `Subject Alternative Name` values: 

`DNS Name=CN=myhost`.

Should produce:
`DNS Name=myhost`

/cc @jspaith

CIPop:
@JimacoMS2 @alextolp FYI

Fixes #362

### DIFF
--- a/tools/CACertificates/CACertificateOverview.md
+++ b/tools/CACertificates/CACertificateOverview.md
@@ -22,7 +22,7 @@ You'll need to do some initial setup prior to running these scripts.
 * `cd` to the directory you want to run in.  All files will be created as children of this directory.
 * Run `Set-ExecutionPolicy -ExecutionPolicy Unrestricted`.  You'll need this for PowerShell to allow you to run the scripts.
 * Get OpenSSL for Windows.  
-  * See [the official documentation] for places to download it.
+  * See [the official documentation] or https://www.openssl.org/docs/faq.html#MISC4 for places to download it or or https://www.openssl.org/source/ to build from source.
   * Set `$ENV:OPENSSL_CONF` to the openssl's openssl.cnf.
 * Run `. .\ca-certs.ps1` .  This is called dot-sourcing and brings the functions of the script into PowerShell's global namespace.
 * Run `Test-CACertsPrerequisites`.
--- a/tools/CACertificates/ca-certs.ps1
+++ b/tools/CACertificates/ca-certs.ps1
@@ -12,8 +12,9 @@ Set-StrictMode -Version 2.0
 # Errors in system routines will stop script execution
 $errorActionPreference    = "stop"
 
-$_rootCertSubject         = "CN=Azure IoT CA TestOnly Root CA"
-$_intermediateCertSubject = "CN=Azure IoT CA TestOnly Intermediate {0} CA"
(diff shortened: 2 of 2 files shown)


tools/CACertificates/CACertificateOverview.md
```diff
@@ -22,7 +22,7 @@ You'll need to do some initial setup prior to running these scripts.
 * `cd` to the directory you want to run in.  All files will be created as children of this directory.
 * Run `Set-ExecutionPolicy -ExecutionPolicy Unrestricted`.  You'll need this for PowerShell to allow you to run the scripts.
 * Get OpenSSL for Windows.  
-  * See [the official documentation] for places to download it.
+  * See [the official documentation] or https://www.openssl.org/docs/faq.html#MISC4 for places to download it or or https://www.openssl.org/source/ to build from source.
```
jspaith: Typo: "or or"

jspaith: Note that it's trivial so I'm just fixing it...

JimacoMS2: Two things here: 

1. The easiest way to install OpenSSL (especially for a newbie) that we have documented is the second bulletpoint here: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-security-x509-create-certificates#prerequisites . This is a link to download and install pre-built binaries from a third party. You can get there from the official documentation linked above, but you have to read through part of that doc, then click through to and read the pre-reqs of the article I just noted. I think it would be helpful to add something like this to this readme: "The quickest option might be download and install a [third-party OpenSSL binary](https://wiki.openssl.org/index.php/Binaries), for example, from [this project on SourceForge](https://sourceforge.net/projects/openssl/)."

2. I think the installing OpenSSL bulletpoint should be first in the list. If you don't set up OpenSSL and add it to your path before opening the PowerShell window, the script will fail. It's not a major thing, but we kind of set folks up to fail if they are not super knowledgeable and installing OpenSSL for the first time.

2.5 -- do we need to be explicit about telling folks to ensure that the script and any other files needed are resident in their working directory?

CIPop: > 1

Sure, that's a good place as well. The `official documentation` link is not pointing at that. By skimming through the docs I didn't understand how to install openssl... 
I'll update the paragraph shortly.

> 2 

👍 

>2.5

Good point: I've asked @JimacoMS2 to change the instructions so that users avoid cloning this repo if using C# or Java, etc. It's only required to download the scripts from this folder.
I'll add the clarification in this doc as well.

@jspaith , one more thing: It would be nice to have a readme.md as well in this folder - GitHub will automatically display the content when opening the folder instead of the user trying to find this .md file. I'll add a readme.md with a link for now - changing it would break all existing links...

jspaith: OK I've just got initial fix checked in (prior to this discussion) but obviously we're always open for new improvements.  I know for certain there are multiple links directly to the existing .MD so just having readme.md point to it is the best we can do.  Thanks

PR #372, opened 2018-02-01
Add explanatory comment to code

markrad:
# Checklist

# Description of the problem
It is easy to overlook the fact that the device twin callback functions are not invoked unless one comments out the call to IoTHubClient_SetDeviceTwinCallback.

# Description of the solution
No code changes. Added a comment above the callback declaration in DECLARE_MODEL indicating that the callback would be ignored unless the call to IoTHubClient_SetDeviceTwinCallback is commented out. I did this since I missed this, my colleague missed this and so did the customer.

### DIFF
--- a/serializer/samples/devicetwin_simplesample/devicetwin_simplesample.c
+++ b/serializer/samples/devicetwin_simplesample/devicetwin_simplesample.c
@@ -36,8 +36,8 @@ DECLARE_MODEL(CarState,
     WITH_REPORTED_PROPERTY(ascii_char_ptr, vanityPlate)
 );
 
-// NOTE: In order for the callback to be invoked you will need to comment out the
-//       call to IoTHubClient_SetDeviceTwinCallback further down.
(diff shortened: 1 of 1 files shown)


serializer/samples/devicetwin_simplesample/devicetwin_simplesample.c
```diff
@@ -36,6 +36,8 @@ DECLARE_MODEL(CarState,
     WITH_REPORTED_PROPERTY(ascii_char_ptr, vanityPlate)
 );
 
+// NOTE: In order for the callback to be invoked you will need to comment out the
```
ewertons: I would suggest this text:
For callbacks defined in the serializer model to be fired for desired properties, IoTHubClient_SetDeviceTwinCallback must not be invoked for the same device in the same application.
Please comment out the call to IoTHubClient_SetDeviceTwinCallback further down to enable the callbacks defined in the model below.

Issue #229, opened 2017-09-12
Issue using jansson json library with this SDK


karlbloedorn:
Using Latest available Ubuntu package on Ubuntu Linux VM in Azure

We have an existing project that uses jansson for json processing. When we link against this sdk, the json_object method for creating an empty json dictionary segfaults due to being linked against the parson library included in this SDK. So there are some symbol collisions that seem to only break things at link or runtime, without compile warnings or issues. I have created the minimum case for this in the sample below.

#include <jansson.h>

int main(){
   json_t *json;
   char *result;
   json = json_object();
   result = json_dumps(json, 0);
   if(result != NULL){
       printf("SUCCESS %s\n", result);
   }
   return 0;
}
karl@mytestmachine:~/testjansson$ gcc main.c -ljansson -lserializer -liothub_client -liothub_client_amqp_ws_transport -luamqp -laziotsharedutil -lcurl -ljansson -pthread -lm -lsqlite3 -lrt -lssl -lcrypto -luuid
karl@mytestmachine:~/testjansson$ ./a.out
SUCCESS {}
karl@mytestmachine:~/testjansson$ gcc main.c -lserializer -liothub_client -liothub_client_amqp_ws_transport -luamqp -laziotsharedutil -lcurl -ljansson -pthread -lm -lsqlite3 -lrt -lssl -lcrypto -luuid   karl@mytestmachine:~/testjansson$ ./a.out
Segmentation fault (core dumped)

Even changing the order of the include  ljansson  first or not, causes it to segfault or not. Within our program we can get it to compile, link, and run as long as we don't call IoTHubClient_CreateFromConnectionString. As soon as that is added in, it fails. I assume it is some dependency that then decides to actually link the dependent symbols, thus overriding the symbols from the jansson library. You can see below the segfault and it occurs in json_value_get_type  and json_value_get_object , which are parson symbols, not jansson, even though we are calling the jansson json_object() call and include jansson.h only and never parson.h

For the segfault:

==93311== Process terminating with default action of signal 11 (SIGSEGV)
==93311==  Access not within mapped region at address 0x9
==93311==    at 0x4020B5: json_value_get_type (in /home/karl/testjansson/a.out)
==93311==    by 0x4020F8: json_value_get_object (in /home/karl/testjansson/a.out)
==93311==    by 0x400E62: main (in /home/karl/testjansson/a.out)
==93311==  If you believe this happened as a result of a stack
[...]

PR #386, opened 2018-03-01
Hide parson library in .so. Version and install .so files properly

markrad:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/229

# Description of the problem
Use of the Parson JSON library can cause customer's issues. The exposed function names in the library are identical to those in other JSON libraries. This can cause the customer's code to erroneously link to the Parson library instead of the desired library.

# Description of the solution
This will only work on Linux and you must build against the shared object files. 
Linux has the ability to "hide" entry points that are only used within the shared object itself. This will resolve the the externs that the shared object has but will not expose them to the clients of the shared object. This change converts the Parson inline compilation into its own library. This is then included in the shared objects that require it. Along with this, I also made sure the shared objects are properly versioned and installed. Many were not built at all which would have rendered the Parson change moot.

### DIFF
--- a/iothub_client/CMakeLists.txt
+++ b/iothub_client/CMakeLists.txt
@@ -18,10 +18,31 @@ set(iothub_client_ll_transport_c_files
     ./src/iothub_client_diagnostic.c
  )
 
-if(NOT ${dont_use_uploadtoblob})
-    set(iothub_client_ll_transport_c_files 
+set(iothub_client_libs)
+
+set(install_staticlibs
+    iothub_client
+)
+
+add_library(parson
+    ../deps/parson/parson.c
+)
+
+if(MSVC)
+    set_source_files_properties(../deps/parson/parson.c PROPERTIES COMPILE_FLAGS "/wd4244 /wd4232")
+endif()
+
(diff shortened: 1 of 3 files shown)


hansdals:
Brilliant! I'm writing a library (libmylibrary.so) that uses ubox (openwrt/linux) and in turn it is using json-c. Which when I use the library with the AzureIoTHubSDK I get the same issue as described in the pull-request above, and when I'm using it with other daemons It does not fail.

It started causing problems after i upgraded the SDK from a 2017 version to 2018-11. In the upgrade I remember now that I had to include -lparson, the parson library in my makefile. libmylibrary.so works fine with other daemons, but only my mqtt client daemon statically linked with the AzureIoTHubSDK crashes. This led me here. I'm now in the process of making it dynamically linked.

My question is should I also hide the json-c symbols in libmylibrary.so ? Can it also cause problems with the AzureIoTHubSDK ? The trick is to add -Wl,--exlude-libs flags, right ?

markrad:
It shouldn't be necessary since the Azure IoT SDK has already resolved its externs by the time you link your code.

PR #439, opened 2018-03-29
Mrmmm3

markrad:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/229

# Description of the problem
The manner in which parson is included into our libraries causes the functions to be exposed to the linker. This caused one customer problems since they clashed with the JSON parser he was using. Switching to parson was not an option for him because he already had a lot of code using the other.
The parson functions can only be hidden on Linux and inside a shared object (.so) file. The customer was happy with that. 

# Description of the solution
Added Linux only code to the build to create parson as a library and have it linked where needed. In the shared objects though, a link flag is passed that tells the link to hide the functions exposed by the parson library. Customer confirmed this worked for him.
However, setting up the dynamic build revealed multiple problems both on Windows and Linux. 

- Windows dll pdbs and library pdbs had the same names and smashed each other
- I moved some source files from iothub_client into the transport libraries to resolve link errors in dynamic builds
- I removed the dynamic build of iothub_client completely because of a circular dependency between this dll and the transport dll. Resolving it via code would have been a significant change (passing pointers to functions instead of the transport directly calling them)
- Fixed the .def files for the .dlls since they did not expose everything they needed to. Created a template file that is the basis for the .def and modified it with cmake for each transport dll built. One only needs to update the template to have a new exported function exposed in all transport dlls. Added all these cmake generated .def files into the .gitignore
- Ensured all shared objects are properly versioned and installed
- Add install steps to the new provisioning code
- Fixed a problem in iothub_client_ll.c that caused a build failure when building with file upload turned off. Not really related to the rest of the fixes and I'll pull it out into a separate commit if asked.
## Issues
It is not clear to me what our #include install strategy should be. It seems that most of them go into a directory that reflects the name of the module such as azure_utpm_c for example. However, there a bunch of files that dumped directly into include. If one looks at prov_auth_client.h then you will see that it is included as azure_prov_client/prov_auth_client.h. However, files such as provisioning_sc_enrollment.h when included do not have a leading directory so one needs to either add an additional include path or they get dumped into the include directory (which I don't think is a good idea). I think all the files that reference the provisioning_sc*.h files should have a relative path added to them. I didn't do this because it affected many source files and felt it was best left to a second change. This only came up because I added the missing install step.


iothub_client/CMakeLists.txt
```diff
@@ -73,25 +56,38 @@ if (${use_prov_client})
     )
 endif()
 
-if(NOT ${dont_use_uploadtoblob})
-    set(iothub_client_ll_transport_h_files
-        ${iothub_client_ll_transport_h_files}
+if(NOT dont_use_uploadtoblob)
```
jebrando: >dont_use_uploadtoblob [](start = 7, length = 21)

Interesting that you removed the ${...} here.  Are you sure it works in all case?

provisioning_client/CMakeLists.txt
```diff
@@ -165,11 +165,23 @@ if (${build_as_dynamic})
         ${PROV_DEVICE_CLEINT_SOURCE_H_FILES}
         ./src/prov_device_client_dll.def
     )
-    target_link_libraries(prov_device_client_dll prov_device_ll_client)
-
+	linkSharedUtil(prov_device_client_dll)
```
jebrando: >	 [](start = 0, length = 1)

normalize tabs to spaces

markrad:
Addressed by a different pull request.

iothub_client/CMakeLists.txt
markrad: I believe that is actually how you are supposed to do it. I think, if by chance your variable contains the name of another variable, then that is the one it will test. Sometimes I've had cmake flag these as warnings.

PR #466, opened 2018-04-19
fix the uint32_t data_msg_id comparing if it is bigger than UINT32_MAX

mamokarz:
fix the uint32_t data_msg_id comparing if it is bigger than UINT32_MAX


jebrando:
Thanks Marcos for the PR.  I understand the issue with this and it needs to be done a little more elegantly.  I shall fix and submit a Code Review.

PR #472, opened 2018-04-25
Fix 3 bugs in ca-certs.ps1

nelsonmorais:
The file chnages are meant to fix:
- The global variable $_intermediateCertSubject was missing and is used as a default parameter value.
- When generating the full chain .pem file, a wrong filename was being picked for the RootCA. The CER file was used instead of the PEM file.

 

# Checklist
- [x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- If this is a modification that impacts the behavior of a public API
  - [ ] I edited the corresponding document in the `devdoc` folder and added or modified requirements.
- I submitted this PR against the correct branch: 
  - [ ] This pull-request is submitted against the `master` branch. 
  - [ ] I have merged the latest `master` branch prior to submission and re-merged as needed after I took any feedback.
  - [ ] I have squashed my changes into one with a clear description of the change.

# Reference/Link to the issue solved with this PR (if any)

# Description of the problem

# Description of the solution

PR #494, opened 2018-05-18
Fix ca-certs.ps1

damonbarry:
I hit this bug when following the steps in [Create an IoT Edge device that acts as a transparent gateway - preview](https://docs.microsoft.com/en-us/azure/iot-edge/how-to-create-transparent-gateway#powershell). Turns out it has already been reported by a customer at MicrosoftDocs/azure-docs#5474.

The function `New-CACertsEdgeDevice` was attempting to generate a default value for an argument from an orphaned variable. I updated the arg logic to use the correct variable, and confirmed that the get the certs I expect.


damonbarry:
@mrohera

jspaith:
I've taken fix for this from https://github.com/Azure/azure-iot-sdk-c/pull/472 so closing this out as change has merged in.  Thanks for help here Damon.

PR #506, opened 2018-06-01
Use portable string format

JetstreamRoySprowl:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/src/iothub_client_diagnostic.c
+++ b/iothub_client/src/iothub_client_diagnostic.c
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 #include <stdlib.h>
+#include <inttypes.h>
 #include <math.h>
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/gballoc.h"
@@ -22,7 +23,7 @@ static char* get_epoch_time(char* timeBuffer)
     char* result;
     time_t epochTime;
     int timeLen = sizeof(time_t);
-    
+
     if ((epochTime = get_time(NULL)) == INDEFINITE_TIME)
     {
         LogError("Failed getting current time");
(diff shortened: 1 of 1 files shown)


iothub_client/src/iothub_client_diagnostic.c
```diff
@@ -43,7 +44,7 @@ static char* get_epoch_time(char* timeBuffer)
     }
     else if (timeLen == sizeof(int32_t))
     {
```
dcristoloveanu: This is actually incorrect.
The standard says that time_t is an arithmetic type. It means it can actually be a float.
On what platform? I don't know that. But according to C99 it is incorrect to assume that because sizeof(time_t) == sizeof(int64_t) the type is actually int64_t.

Thus, if you are making a cast to something (and implicitly potentially the printed value can be completely wrong) I think you can simply cast to int64_t and print once, no need to have 2 branches one for int32_t and int64_t.

You could also put a comment that the prinout might be completely wrong if the type of time_t is actually double let's say.

Not a big deal obviously as this only impacts the printed value, but just saying :-)

dcristoloveanu: Peoples on the Interwebs + C99 standard might also suggest converting to struct tm and printing that, which would go through the proper channels of the C language.

---
In reply to: [192479429](https://github.com/Azure/azure-iot-sdk-c/pull/506#discussion_r192479429) [](ancestors = 192479429)

JetstreamRoySprowl: Yeah, right now I'm just trying to make things clear for the compiler so it won't gripe. If I were doing more of a rewrite I would use our own agenttime adapter, which has clear integer syntax.

JetstreamRoySprowl: Polished it up a bit more. There isn't a problem if the mythical floating point time_t shows up because of the casts to int types. Our get_time function is already specified to deliver epoch time per the agenttime spec, so it's guaranteed to be seconds (which C99 does not guarantee.)

PR #530, opened 2018-06-25
CA certificate bash script permissions fix

mrohera:
# Checklist

# Description of the problem

When generating the private key of the device CA its permission access bits were set to 400. This is generally fine if the user uses the key. However if a service running on the host machine were to access the device CA key, there is an access error.
 
# Description of the solution
 
The permissions for the device CA key ONLY were set to 444.


mrohera:
Seems like PRs are not the right way to push a fix into modules-preview. Abandoning this PR for now.

PR #531, opened 2018-06-25
Modifying permissions on device CA key so that it can be used by othe…

mrohera:
…r services running on the host

 

# Checklist

# Description of the problem

When generating the private key of the device CA its permission access bits were set to 400. This is generally fine if the user uses the key. However if a service running on the host machine were to access the device CA key, there is an access error.

# Description of the solution
 
The permissions for the device CA key ONLY were set to 444.


mrohera:
Will close this PR and create a new branch based off master per John's guidance.

PR #566, opened 2018-07-13
Export protocol functions

markrad:
# Checklist

# Description of the problem
Protocol functions are not exported in Windows dll.

# Description of the solution
Added functions to def file.

### DIFF
--- a/iothub_client/CMakeLists.txt
+++ b/iothub_client/CMakeLists.txt
@@ -378,6 +378,18 @@ if (${build_as_dynamic})
     if(NOT ${dont_use_uploadtoblob})
         set(iothub_def_file ${iothub_def_file} ./src/upload_to_blob.def)
     endif()
+	
+	if(use_amqp)
+		set(iothub_def_file ${iothub_def_file} ./src/iothub_transport_amqp.def)
+	endif()
+	
+	if(use_http)
+		set(iothub_def_file ${iothub_def_file} ./src/iothub_transport_http.def)
+	endif()
+	
+	if(use_mqtt)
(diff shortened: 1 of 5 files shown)


iothub_client/src/iothub_client_dll.def
```diff
@@ -145,6 +145,12 @@ EXPORTS
     IoTHubMessage_SetInputName
     IoTHubMessage_SetMessageId
     IoTHubMessage_SetProperty
+	
+	AMQP_Protocol
+	AMQP_Protocol_over_WebSocketsTls
+	HTTP_Protocol
+	MQTT_Protocol
+	MQTT_WebSocket_Protocol
```
jebrando: These will have to be broken up in 3 different files so if someone runs cmake without MQTT or AMQP or HTTP it doesn't break.  So you'll need to create 3 iothub_transport_XXXX.def files and then include each in the cmake depending on the transports that are being used.

iothub_client/CMakeLists.txt
```diff
@@ -378,6 +378,18 @@ if (${build_as_dynamic})
...
+		set(iothub_def_file ${iothub_def_file} ./src/iothub_transport_amqp.def)
+	endif()
+	
+	if(use_http)
+		set(iothub_def_file ${iothub_def_file} ./src/iothub_transport_http.def)
+	endif()
+	
+	if(use_mqtt)
+		set(iothub_def_file ${iothub_def_file} ./src/iothub_transport_mqtt.def)
+	endif()
```
jebrando: Can you please normalize tabs to spaces.

PR #570, opened 2018-07-20
Do not add /WX compile flag for MinGW

vkuzmichev:
# Checklist

# Description of the problem
Azure IoT SDK does not compile in Windows using MinGW-W64 compiler with the following error:
gcc.exe: error: /WX: No such file or directory

# Description of the solution
Amended CMakeLists.txt to not add /WX to compile flags for MinGW (which follows Unix-style command line options formatting), and instead use -Werror.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,20 +31,20 @@ endif()
 include (CTest)
 
 if (MSVC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /wd4232")
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /wd4232")
-endif()
-
-if (WIN32)
-    # windows needs this define
-    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /wd4232")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /wd4232")
     # Make warning as error
     add_definitions(/WX)
 else()
     # Make warning as error
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
-endif(WIN32)
+endif()
+
+IF(WIN32)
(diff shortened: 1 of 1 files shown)


jspaith:
Thanks for this contribution.  This generally looks good, except it's not clear to my why (pre-existing) we have a WIN32 line and an MSVC line a few lines above.  It seems like if we put all the WIN32'ness into MSVC if block then MinGW would work for free.

I need to followup on my side to see if there's history I'm unaware of here.

jspaith:
To my knowledge, /wx is Visual Studio specific.  In practice the only Windows compiler we've supported historically has been Visual Studio - as your experiments with MinGW are showing wasn't working.

For which compiler we support, visual studio/gcc/clang get by far the most amount of testing.  We're C99 compliant though so there are embedded compilers used against our code, too.  To my knowledge though they tend to have either a gcc frontend or are inspired by gcc/clang such that the `-Werror` has never been a problem for them.

I think correct fix here is to clarify compiler versus platform.  So I think we'd want something like below (I haven't verified this).
```
if (MSVC)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /wd4232")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /wd4232")
    add_definitions(/WX)
else()
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
endif()

if (WIN32)
    # windows needs this define
    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
endif()
```

If this works for you on MinGW and looks reasonable, if you want to create new iteration of this pull request (squash please :)) we can take this.

vkuzmichev:
@jspaith 
Ok, sounds reasonable. Done.
Thank you.

jspaith:
@vkuzmichev - Thank you for your contribution to our open-sourced project! Please help us improve by filling out [this 2-minute customer satisfaction survey](https://forms.office.com/Pages/DesignPage.aspx#FormId=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu&Token=af48ed4ed02842bea21e0d3933b78fac)

PR #594, opened 2018-08-14
Fix #587: avoid new GCC 8.2 warning (s/strncpy/memcpy/)

### LINKED ISSUE #587: GCC 8.2 generates warnings when compiling serializer/src/commanddecoder.c
Related issue: https://github.com/Azure/azure-c-shared-utility/issues/174

When compiling serializer/src/commanddecoder.c with 

    gcc (Debian 8.2.0-3) 8.2.0

I'm getting warnings (and thus errors (-Werror)) due the use of strncpy. GCC is unable to figure out that the length-arg is correctly guarded (or is it not)?

```
In function ‘DecodeAndExecuteModelAction.isra.0’,
    inlined from ‘ScanActionPathAndExecuteAction’ at azure-iot-sdk-c/serializer/src/commanddecoder.c:441:26,
    inlined from ‘DecodeCommand’ at azure-iot-sdk-c/serializer/src/commanddecoder.c:607:22,
    inlined from ‘CommandDecoder_ExecuteCommand’ at azure-iot-sdk-c/serializer/src/commanddecoder.c:679:26:
azure-iot-sdk-c/serializer/src/commanddecoder.c:194:13: error: ‘strncpy’ specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
         if (strncpy(tempStr, actionName, strLength - 1) == NULL)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
azure-iot-sdk-c/serializer/src/commanddecoder.c: In function ‘CommandDecoder_ExecuteCommand’:
azure-iot-sdk-c/serializer/src/commanddecoder.c:176:24: note: length computed here
     size_t strLength = strlen(actionName);
```

`strncpy()` could be replace with `memcpy()` here, I think.

pboettch:
GCC 8.2 is not able to determine that the length is OK for dst
and thus moans about the usage of strncpy.

How it is done, memcpy seems more appropriate to be used.

Maybe in the future this code can be even more simplified.

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/serializer/src/commanddecoder.c
+++ b/serializer/src/commanddecoder.c
@@ -191,7 +191,7 @@ static EXECUTE_COMMAND_RESULT DecodeAndExecuteModelAction(COMMAND_DECODER_HANDLE
 #ifdef _MSC_VER
 #pragma warning(suppress: 6324) /* We intentionally use here strncpy */ 
 #endif
-        if (strncpy(tempStr, actionName, strLength - 1) == NULL)
+        if (memcpy(tempStr, actionName, strLength - 1) == NULL)
         {
             /* Codes_SRS_COMMAND_DECODER_99_021:[ If the parsing of the command fails for any other reason the command shall not be dispatched.] */
             LogError("Invalid action name.");


serializer/src/commanddecoder.c
```diff
@@ -191,7 +191,7 @@ static EXECUTE_COMMAND_RESULT DecodeAndExecuteModelAction(COMMAND_DECODER_HANDLE
 #ifdef _MSC_VER
 #pragma warning(suppress: 6324) /* We intentionally use here strncpy */ 
 #endif
-        if (strncpy(tempStr, actionName, strLength - 1) == NULL)
+        if (memcpy(tempStr, actionName, strLength - 1) == NULL)
```
ewertons: Hi @pboettch ,
there was a bug in this part of the code that we had planned to fix, but with your change now it needs to be fixed in a different way.

To allow us to merge your change in and record your contribution, could you make the following changes please?

1. Change the memcpy call to: ```if (memcpy(tempStr, actionName, strLength + 1) == NULL)```

(note the strLength + 1 instead of strLength - 1)

and

2. Remove ```tempStr[strLength - 1] = 0;``` (on line 209 below).

jebrando: @pboettch I shall take your branch and apply the changes that Ewerton suggested.  I need to make the change as soon as possible.  I will merge this change into the SDK.

Thanks,
Jelani

pboettch: Just got up. I see you merged it into your memcpy-fix branch. Do not hesitate to request more changes. If I can be helpful, I will be.

PR #670, opened 2018-10-19
update certGen.sh

nnmer:
# Description of the problem
the command cannot execute successfully on windows environment

# Description of the solution
fix the openssl -subj parameter to make the script successfully run on both windows and linux os


tools/CACertificates/certGen.sh
```diff
@@ -28,6 +28,15 @@ intermediate_ca_password="1234"
 root_ca_prefix="azure-iot-test-only.root.ca"
 intermediate_ca_prefix="azure-iot-test-only.intermediate"
 
+function makeCNsubject()
+{
+    local result="/CN=${1}"
+    case $OSTYPE in
+        msys|cygwin|win32) result="/${result}"
```
YoDaMa: Can you explain the purpose of this exactly? Does the script not work in cygwin environment without the preceding double slash?

nnmer: yeah, you right. double checked and remove cygwin from the expression

nnmer:
> Still unsure what is the purpose of fixing this syntax for msys and win32. Are you running it on a specific platform that is failing? Are you specifically running it in a 32 Bit environment? certGen.sh is mainly for demonstrating X.509 certificates for a deployment, so running on a regular Linux or Windows box. Your initial comment that it doesn't work in windows doesn't seem to make sense currently.

My box is Win10, 64Bit.
The purpose of fixing is to have a bash script which is agnostic to platform. Personally I prefer to use bash, so I have 1 script for all platforms other than several for each. Thus I expect script to work and produce same result on different platform in the same way.

jspaith:
@nnmer - could you please provide more context (and ideally links if any) on why specifically this change is needed for Windows?  

I searched around a bit about why an extra "/" that's being added here would be required and was unable to find anything.

Thanks

nnmer:
@jspaith - I have a counterquestion, have you guys tried yourself?

steps to reproduce:
1 - Windows 10. (but i believe earlier versions may have same behavior)
2 - pull the code
3 - open cmd
4 - cd ....../azure-iot-sdk-c/tools/CACertificates
5 - sh certGen.sh create_root_and_intermediate
6 - output is Failed.... 'Subject does not start with "/"'


here is a picture:


and here is when we add extra '/'

jspaith:
@nnmer - makes sense.  I had initially tried to repro with your fix already included to see it working, I didn't realize that the problem happened so early on.  I've verified fixed .sh on both MinGW64 and no regressions on Ubuntu 18.04.

We'd never really tested the mix+match in 1st place, but as this is a good fix we'll get this in.

Thank you for getting it to us and clarification.

PR #692, opened 2018-10-24
Fix to set authorization mode as x509

ysripath:
Signed-off-by: Yuvaraj <yuvarajsripathi91@gmail.com>

 

# Checklist

# Description of the problem

# Description of the solution


jebrando:
@ysripath Thank you for the review, but we have this as `cred_type != IOTHUB_CREDENTIAL_TYPE_X509` because if the cred_type is not IOTHUB_CREDENTIAL_TYPE_X509 then we want to set it, if it is already set to IOTHUB_CREDENTIAL_TYPE_X509 then we don't need to go into the function and incur an unnecessary function call.

Issue #732, opened 2018-11-27
2018-11-20 release fails to link


larsonmpdx:
with a source-built library installed this way:
````
RUN git clone -b 2018-11-20 --recursive https://github.com/Azure/azure-iot-sdk-c.git \
  && cd azure-iot-sdk-c \
  && mkdir cmake-build \
  && cd cmake-build \
  && cmake .. \
  && make -j4 \
  && make install \
  && ldconfig
````
then building my own project with this cmake configuration:

````
target_link_libraries(myproject
        iothub_client
        iothub_client_mqtt_transport
        aziotsharedutil
        pthread
        curl
        ssl
        crypto
        m
        umqtt
          # shouldn't be needed. see issue https://github.com/Azure/azure-iot-sdk-c/issues/550
	  # prov_auth_client # causes conflicts on source-built sdk, nov 2018
	  # hsm_security_client # causes conflicts on source-built sdk, nov 2018
        uhttp
        uuid
        parson
        )
````
gives this error:
````
//usr/local/lib/libumqtt.a(mqtt_message.c.o): In function `mqttmessage_getTopicLevels':
mqtt_message.c:(.text+0x510): undefined reference to `StringToken_Split'
collect2: error: ld returned 1 exit status
````
stringtoken_split should come from aziotsharedutil but it's not working. things worked with this same build process with the 2018-09-11 release

the recommended linker options are given here:
https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/ubuntu_apt-get_sample_setup.md

see also #621, #550

larsonmpdx:
I tested with the 2018-10-03 release and it works with the above build steps

jebrando:
@larsonmpdx you hit a quirk with the gcc compiler, I'll try to explain (very poorly) below.  Moving the umqtt lib above aziotsharedutil will fix the issue and we shall fix our documentation to make it more clear.  Your target_link_libraries code should look like this:

target_link_libraries(myproject
        iothub_client
        iothub_client_mqtt_transport
        umqtt
        uhttp
        aziotsharedutil
        pthread
        curl
        ssl
        crypto
        m
          # shouldn't be needed. see issue https://github.com/Azure/azure-iot-sdk-c/issues/550
	  # prov_auth_client # causes conflicts on source-built sdk, nov 2018
	  # hsm_security_client # causes conflicts on source-built sdk, nov 2018
        uuid
        parson
        )

[...]

PR #735, opened 2018-11-27
demo link failure using a linux dockerfile and an independent cmakelists

larsonmpdx:
demonstrates link failure reported in #732 

also is a good example of an out of tree build which would be nice to include in tests or documentation so users can more easily start dependent projects


massand:
@larsonmpdx Closing this PR since this is being used as a demonstration of an actively investigated issue. Please continue to post any updates on the issue.

PR #737, opened 2018-11-28
Remove leading forward slash from UDS path on Windows

damonbarry:
The C SDK is already well-abstracted to communicate over UNIX domain sockets on Windows, except for one minor issue: `unix://` URLs with a Windows filesystem path will have a leading forward slash ('/') that needs to be stripped off before the "host" (it's not really a host, but it is the thing you connect to) is passed to socketio.

UNIX domain socket URLs don't have a hostname because the local host is assumed--it's an IPC mechanism. The socket file to connect to is found in the URL's path, and URL paths always begin with a '/'. On POSIX systems this isn't a problem because it aligns with the way filesystem paths work, but on Windows the '/' becomes purely URL syntax, and needs to be removed from the actual path.

### DIFF
--- a/provisioning_client/adapters/hsm_client_http_edge.c
+++ b/provisioning_client/adapters/hsm_client_http_edge.c
@@ -140,13 +140,25 @@ static int read_and_parse_edge_uri(HSM_CLIENT_HTTP_EDGE* hsm_client_http_edge)
         }
         else if (hsm_client_http_edge->workload_protocol_type == WORKLOAD_PROTOCOL_TYPE_UNIX_DOMAIN_SOCKET)
         {
-            // Make sure that there is an address set.
-            if (workload_uri[unix_domain_sockets_prefix_len] == 0)
+            const char* socket_path = workload_uri + unix_domain_sockets_prefix_len;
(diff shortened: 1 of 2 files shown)


provisioning_client/adapters/hsm_client_http_edge.c
```diff
@@ -140,13 +140,25 @@ static int read_and_parse_edge_uri(HSM_CLIENT_HTTP_EDGE* hsm_client_http_edge)
         }
         else if (hsm_client_http_edge->workload_protocol_type == WORKLOAD_PROTOCOL_TYPE_UNIX_DOMAIN_SOCKET)
         {
+            const char* socket_path = workload_uri + unix_domain_sockets_prefix_len;
+
+#ifdef WIN32
```
damonbarry: We're not doing it this way anywhere else in the C SDK, so I'm open to feedback about how best to make sure we only strip the forward slash on Windows devices.

damonbarry:
I opened this PR against my fork but the build system requires a branch on the main repo. I've opened #743; closing this one.

Issue #746, opened 2018-11-29
device client will never connect if started before networking is ready


larsonmpdx:
- **OS and version used:** <VERSION>  buildroot linux 4.1.15 on arm

- **SDK version used:** <VERSION>  Nov 20 2018 1.2.11

# Description of the issue:
If the iot device client application is started before network connectivity is ready (dns not yet working and connections can't be opened), it can get stuck and may never connect even with unlimited time and retries.  if it's started after network connectivity is ready it will work normally and can become disconnected through network faults and then reconnect later and so on, working as expected.

# Code sample exhibiting the issue:
an application more or less the same as https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/samples/iothub_ll_client_x509_sample/iothub_ll_client_x509_sample.c
but running in a loop and relying on the iot library's retry functionality

# Console log of the issue:
this is repeated forever, each time the iot library tries to reconnect (at some interval based on the azure iot reconnect policy):
````
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/c-utility/adapters/socketio_berkeley.c Func:wait_for_connection Line:386 Failure: connect failure 111.
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/c-utility/adapters/socketio_berkeley.c Func:socketio_open Line:770 wait_for_connection failed
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/c-utility/adapters/tlsio_openssl.c Func:on_underlying_io_open_complete Line:760 Invalid tlsio_state. Expected state is TLSIO_STATE_OPENING_UNDERLYING_IO.
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/c-utility/adapters/tlsio_openssl.c Func:tlsio_openssl_open Line:1258 Failed opening the underlying I/O.
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/umqtt/src/mqtt_client.c Func:mqtt_client_connect Line:1001 Error: io_open failed
Error: Time:Thu Nov 29 17:04:08 2018 File:/azure-iot-sdk-c/iothub_client/src/iothubtransport_mqtt_common.c Func:SendMqttConnectMsg Line:2141 failure connecting to address [address].
````

larsonmpdx:
[...]

PR #754, opened 2018-12-04
Demo loop connectivity problem

larsonmpdx:
demo for #746


larsonmpdx:
now that the res_init() call is in it's working

this would still be useful as a more full example, it shows how to detect queue growth while disconnected, how to correctly call dowork() while sleeping, and how to use a res_init() call which is frequently needed as the library doesn't handle this

ewertons:
Closed as the issue has been addressed.
Thanks for contributing to the Azure IoT SDKs!

PR #759, opened 2018-12-07
Edge certs from env

jspaith:
Add proper reading of environment variables 'EdgeModuleCACertificateFile' and 'EdgeModuleCACertificateFile'.

These will allow us to run the C SDK inside an EXE, instead of a Docker container, when talking to IoT Edge.  This will massively simplify developer story.

Needed special handling for both the IoTHub core of course, but also HTTP (for InvokeMethod) which needs the trusted certs sent down to curl/schannel.

NOTE: I will bring back the UT testing to retrieve_edge_certificates() which sadly got lost during Edge Ga prior to actually trying to merge this.


iothub_client/src/iothub_client_edge.c
```diff
@@ -368,6 +369,8 @@ static STRING_HANDLE createRelativePath(const char* deviceId, const char* module
     return result;
 }
 
+static const char* ENVIRONMENT_VAR_EDGEHUB_CACERTIFICATEFILE = "EdgeModuleCACertificateFile";
```
jebrando: Nitpick: Can you please move this up with the other const char*, just so they're all together

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -25,6 +25,8 @@
 #include "internal/iothub_client_diagnostic.h"
 #include "internal/iothubtransport.h"
 
+#include <errno.h>
```
bikamani: Nit: Can be moved along with standard headers at line #7

iothub_client/src/iothub_client_edge.c
jspaith: Fixed.

iothub_client/src/iothub_client_core_ll.c
jspaith: Fixed.

PR #762, opened 2018-12-07
Upload prov

jebrando:


ewertons:
>         json_value_free(json);

This can be avoided if this block gets moved to the else { } block bellow.

---
Refers to: iothub_client/src/iothub_client_ll_uploadtoblob.c:168 in 3071612. [](commit_id = 3071612246b68fab26dab7599132308276848522, deletion_comment = False)

Issue #758, opened 2018-12-06
SetOption for Sleep timeout in convenience layer


jspaith:
In https://github.com/Azure/azure-iot-sdk-c/issues/648 , we've got feedback about CPU usage in convenience layer of SDK.  The issue is that we have a loop running through all the pending items in IoT SDK and we're only sleeping for 1 millisecond between runs.

The workaround today is to use the _LL_ layer, in which the caller controls the loop.  

A better solution though would be to have a SetOption() to allow the caller to specify the sleep timeout.

YoDaMa:
To address this I've created a branch and submitted a [PR here](https://github.com/Azure/azure-iot-sdk-c/pull/813).

YoDaMa:
This is addressed in https://github.com/Azure/azure-iot-sdk-c/pull/813

az-iot-builder-01:
@jspaith, @YoDaMa, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #813, opened 2019-01-16
Setoption sleep timeout

YoDaMa:
# Checklist
- [NO] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

See Github #758 

# Description of the problem

Currently it is not possible to adjust the ThreadAPI_Sleep time for the convenience layer. This change proposes a adding an option to set this value, which is defaulted to 1ms.

# Description of the solution

### DIFF
--- a/iothub_client/inc/iothub_client_options.h
+++ b/iothub_client/inc/iothub_client_options.h
@@ -76,6 +76,8 @@ extern "C"
     //diagnostic sampling percentage value, [0-100]
     static STATIC_VAR_UNUSED const char* OPTION_DIAGNOSTIC_SAMPLING_PERCENTAGE = "diag_sampling_percentage";
 
+static STATIC_VAR_UNUSED const char* OPTION_CONVENIENCE_LOOP_TIME = "convenience_loop_timeout";
+
 #ifdef __cplusplus
 }
 #endif
--- a/iothub_client/src/iothub_client_core.c
+++ b/iothub_client/src/iothub_client_core.c
@@ -19,6 +19,9 @@
 #include "azure_c_shared_utility/xlogging.h"
 #include "azure_c_shared_utility/singlylinkedlist.h"
 #include "azure_c_shared_utility/vector.h"
+#include "iothub_client_options.h"
+
+#define LOOP_TIMEOUT_DEFAULT 1
 
 struct IOTHUB_QUEUE_CONTEXT_TAG;
 
@@ -43,6 +46,7 @@ typedef struct IOTHUB_CLIENT_CORE_INSTANCE_TAG
     struct IOTHUB_QUEUE_CONTEXT_TAG* connection_status_user_context;
     struct IOTHUB_QUEUE_CONTEXT_TAG* message_user_context;
     struct IOTHUB_QUEUE_CONTEXT_TAG* method_user_context;
+    uint16_t loop_timeout;
 } IOTHUB_CLIENT_CORE_INSTANCE;
 
 typedef enum HTTPWORKER_THREAD_TYPE_TAG
@@ -1691,13 +1695,20 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_SetOption(IOTHUB_CLIENT_CORE_HANDLE iotHub
         }
         else
         {
-            /*Codes_SRS_IOTHUBCLIENT_02_038: [If optionName doesn't match one of the options handled by this module then IoTHubClient_SetOption shall call IoTHubClientCore_LL_SetOption passing the same parameters and return what IoTHubClientCore_LL_SetOption returns.] */
-            result = IoTHubClientCore_LL_SetOption(iotHubClientInstance->IoTHubClientLLHandle, optionName, value);
-            if (result != IOTHUB_CLIENT_OK)
+            if (strcmp(OPTION_CONVENIENCE_LOOP_TIME, optionName) == 0)
             {
-                LogError("IoTHubClientCore_LL_SetOption failed");
+                iotHubClientInstance->loop_timeout = *((uint16_t *)value);
+                result = IOTHUB_CLIENT_OK;
             }
-
+            else
+            {
+                /*Codes_SRS_IOTHUBCLIENT_02_038: [If optionName doesn't match one of the options handled by this module then IoTHubClient_SetOption shall call IoTHubClientCore_LL_SetOption passing the same parameters and return what IoTHubClientCore_LL_SetOption returns.] */
+                result = IoTHubClientCore_LL_SetOption(iotHubClientInstance->IoTHubClientLLHandle, optionName, value);
+                if (result != IOTHUB_CLIENT_OK)
+                {
+                    LogError("IoTHubClientCore_LL_SetOption failed");
+                }
+            }
             (void)Unlock(iotHubClientInstance->LockHandle);
         }
     }
--- a/iothub_client/src/iothub_client_core.c
+++ b/iothub_client/src/iothub_client_core.c
@@ -817,7 +817,7 @@ static int ScheduleWork_Thread(void* threadArgument)
             /*Codes_SRS_IOTHUBCLIENT_01_040: [If acquiring the lock fails, IoTHubClientCore_LL_DoWork shall not be called.]*/
             /*no code, shall retry*/
         }
-        (void)ThreadAPI_Sleep(1);
+        (void)ThreadAPI_Sleep(iotHubClientInstance->loop_timeout);
     }
 
     ThreadAPI_Exit(0);


iothub_client/inc/iothub_client_options.h
```diff
@@ -76,6 +76,8 @@ extern "C"
     //diagnostic sampling percentage value, [0-100]
     static STATIC_VAR_UNUSED const char* OPTION_DIAGNOSTIC_SAMPLING_PERCENTAGE = "diag_sampling_percentage";
 
+static STATIC_VAR_UNUSED const char* OPTION_CONVENIENCE_LOOP_TIME = "convenience_loop_timeout";
```
ewertons: I would suggest using another name.
Something like OPTION_DO_WORK_FREQUENCY_IN_MS.
I haven't heard the term CONVENIENCE_LOOP_TIME before when referring to the DOWork frequency, so it might be a little foreign to the customers if introduced as such.

ewertons: You should send an email to the team asking their opinion I think.

iothub_client/src/iothub_client_core.c
```diff
@@ -19,6 +19,9 @@
 #include "azure_c_shared_utility/xlogging.h"
 #include "azure_c_shared_utility/singlylinkedlist.h"
 #include "azure_c_shared_utility/vector.h"
+#include "iothub_client_options.h"
+
+#define LOOP_TIMEOUT_DEFAULT 1
```
ewertons: A good name will have the terms FREQUENCY + DEFAULT + IN_MS (or MILLISEC).
Maybe DEFAULT_DO_WORK_FREQUENCY_MILLISEC ?

iothub_client/src/iothub_client_core.c
```diff
@@ -813,7 +817,7 @@ static int ScheduleWork_Thread(void* threadArgument)
             /*Codes_SRS_IOTHUBCLIENT_01_040: [If acquiring the lock fails, IoTHubClientCore_LL_DoWork shall not be called.]*/
             /*no code, shall retry*/
         }
-        (void)ThreadAPI_Sleep(1);
+        (void)ThreadAPI_Sleep(iotHubClientInstance->loop_timeout);
```
ewertons: again, loop_timeout is quite different from what we usually call it. Should be something that matches the other names  commented above.

iothub_client/src/iothub_client_core.c
```diff
@@ -1691,13 +1695,20 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_SetOption(IOTHUB_CLIENT_CORE_HANDLE iotHub
         }
         else
         {
-            /*Codes_SRS_IOTHUBCLIENT_02_038: [If optionName doesn't match one of the options handled by this module then IoTHubClient_SetOption shall call IoTHubClientCore_LL_SetOption passing the same parameters and return what IoTHubClientCore_LL_SetOption returns.] */
-            result = IoTHubClientCore_LL_SetOption(iotHubClientInstance->IoTHubClientLLHandle, optionName, value);
-            if (result != IOTHUB_CLIENT_OK)
+            if (strcmp(OPTION_CONVENIENCE_LOOP_TIME, optionName) == 0)
             {
```
ewertons: You should define if there is a range of valid values, and verify it here (and fail setoption of the number is out of range).
TODO: discuss with team what a good range would be .

iothub_client/src/iothub_client_core.c
```diff
@@ -1691,13 +1699,21 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_SetOption(IOTHUB_CLIENT_CORE_HANDLE iotHub
...
         else
         {
-            /*Codes_SRS_IOTHUBCLIENT_02_038: [If optionName doesn't match one of the options handled by this module then IoTHubClient_SetOption shall call IoTHubClientCore_LL_SetOption passing the same parameters and return what IoTHubClientCore_LL_SetOption returns.] */
-            result = IoTHubClientCore_LL_SetOption(iotHubClientInstance->IoTHubClientLLHandle, optionName, value);
-            if (result != IOTHUB_CLIENT_OK)
+            /* Codes_SRS_IOTHUBCLIENT_41_001 [ If parameter `optionName` is `OPTION_CONVENIENCE_LOOP_TIME` then `IoTHubClient_SetOption` shall set `loop_timeout` parameter of `IoTHubClientInstance` ]*/ 
+            if (strcmp(OPTION_CONVENIENCE_LOOP_TIME, optionName) == 0)
             {
-                LogError("IoTHubClientCore_LL_SetOption failed");
+                iotHubClientInstance->loop_timeout = *((uint16_t *)value);
```
massand: >iotHubClientInstance->loop_timeout = *((uint16_t *)value); [](start = 16, length = 58)

This set cannot happen here. It needs to occur within the LL DoWork itself because a race condition can occur in ScheduleWorkThread function around the read on low-end devices. There is no lock around the Sleep but there is one around the LL DoWork itself.

iothub_client/inc/iothub_client_options.h
massand: >OPTION_CONVENIENCE_LOOP_TIME [](start = 37, length = 28)

Sleep interval is more appropriate naming here than loop time. We need to document some guidance around the optimal values through CPU usage analysis on , say, Raspberry Pi, MXChip, Azure Sphere, etc. Please put out some measurements along with this commit in an article on our docs folder.

iothub_client/src/iothub_client_core.c
```diff
@@ -1691,13 +1699,21 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_SetOption(IOTHUB_CLIENT_CORE_HANDLE iotHub
         }
         else
         {
-            /*Codes_SRS_IOTHUBCLIENT_02_038: [If optionName doesn't match one of the options handled by this module then IoTHubClient_SetOption shall call IoTHubClientCore_LL_SetOption passing the same parameters and return what IoTHubClientCore_LL_SetOption returns.] */
-            result = IoTHubClientCore_LL_SetOption(iotHubClientInstance->IoTHubClientLLHandle, optionName, value);
-            if (result != IOTHUB_CLIENT_OK)
+            /* Codes_SRS_IOTHUBCLIENT_41_001 [ If parameter `optionName` is `OPTION_CONVENIENCE_LOOP_TIME` then `IoTHubClient_SetOption` shall set `loop_timeout` parameter of `IoTHubClientInstance` ]*/ 
+            if (strcmp(OPTION_CONVENIENCE_LOOP_TIME, optionName) == 0)
```
massand: >OPTION_CONVENIENCE_LOOP_TIME [](start = 23, length = 28)

Is there any timeout within our system that might conflict with this? For instance, can I set message timeout < loop time and never send a message? We should help avoid these scenarios. Please change the telemetry sample and try this out. This is why a design within the Github issue is preferred over design discussions within PRs :) 
For reference, please see the Current Configuration Options section of this article - [https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/connection_and_messaging_reliability.md](https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/connection_and_messaging_reliability.md)

iothub_client/inc/iothub_client_options.h
YoDaMa: After team consensus I've updated the names to center around OPTION_DO_WORK_FREQUENCY_IN_MS

iothub_client/src/iothub_client_core.c
YoDaMa: addressed

iothub_client/src/iothub_client_core.c
YoDaMa: addressed and removed race condition

iothub_client/src/iothub_client_core.c
YoDaMa: This comment is not relevant in this scenario since we are threadlocked here, but makes sense for when I retrieve the value for ThreadAPI_Sleep. I've fixed that.

iothub_client/inc/iothub_client_options.h
YoDaMa: i've updated the naming around the code.

iothub_client/src/iothub_client_core.c
YoDaMa: setting the upperlimit to this to be 100ms

iothub_client/src/iothub_client_core.c
YoDaMa: there was actually a bug in the message timeout (it would essentially never be called :) ) so I've also fixed that. I'll add validation for message timeout and for this to make sure the two options don't step on each other's toes.

iothub_client/src/iothub_client_core.c
```diff
@@ -43,6 +46,8 @@ typedef struct IOTHUB_CLIENT_CORE_INSTANCE_TAG
     struct IOTHUB_QUEUE_CONTEXT_TAG* connection_status_user_context;
     struct IOTHUB_QUEUE_CONTEXT_TAG* message_user_context;
     struct IOTHUB_QUEUE_CONTEXT_TAG* method_user_context;
+    uint16_t do_work_freq_ms;
+    tickcounter_ms_t currentMessageTimeout;
```
jebrando: >tickcounter_ms_t [](start = 4, length = 16)

This should probably be just a uint32_t instead of a tickcounter_ms_t.  

As a side note if you bring in a value you need to make sure you add the #include header at the top.

PR #835, opened 2019-01-31
Public preview master merge

massand:
Merge from master to public-preview. 
This PR contains two types of changes - 
1) New changes from master since the public-preview release
2) Merge fixes for some core pieces that made its way into public-preview


iothub_client/src/iothub_client_core_ll.c
```diff
@@ -487,36 +487,6 @@ static STRING_HANDLE make_product_info(const char* product)
     return result;
 }
 
-static void destroy_iothub_client(IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* client_handle)
-{
```
ewertons: I had added this function from the Device Streaming branch.
It is a good improvement to have.
Although, I understand it should have been brought to master in separate. 

So there are two ways to fixing this:
1. Keep the improvement (so you won't need to fix unit tests as well)
2. Take it out, update unit tests

And after that we should bring the change to master anyway, so we don't have these merge conflicts every time we update public-preview branch.

PR #843, opened 2019-02-01
Update http in preparation for code changes

jebrando:


iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+                LogError("Failed getting header %lu", (unsigned long)index);
+                result = __FAILURE__;
+                break;
+            }
+            else
+            {
+                char* colon_pos;
+                if (strncmp(IOTHUB_APP_PREFIX, completeHeader, strlen(IOTHUB_APP_PREFIX)) == 0)
+                {
+                    /*looks like a property headers*/
```
jspaith: >/* [](start = 20, length = 2)

C++ style comments?

jebrando: Yes, these are littered throughout this code, I would like to change them all, but not now.

jebrando: Existing comment, will change this to look better.

---
In reply to: [253209664](https://github.com/Azure/azure-iot-sdk-c/pull/843#discussion_r253209664) [](ancestors = 253209664)

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+            if (HTTPHeaders_GetHeader(resp_header, index, &completeHeader) != HTTP_HEADERS_OK)
+            {
+                LogError("Failed getting header %lu", (unsigned long)index);
+                result = __FAILURE__;
+                break;
+            }
+            else
+            {
+                char* colon_pos;
+                if (strncmp(IOTHUB_APP_PREFIX, completeHeader, strlen(IOTHUB_APP_PREFIX)) == 0)
```
jspaith: >strlen [](start = 63, length = 6)

Nit - strlen(const) should be something more like

const size_t IOTHUB_APP_PREFIX_LEN = sizeof(IOTHUB_APP_PREFIX) - 1; // on top

and then use constant instead throought.

iothub_client/src/iothubtransporthttp.c
jspaith: >/*looks like a property headers*/ [](start = 20, length = 33)

"headers" should be "header"; not sure we need comment...

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+                {
+                    /*looks like a property headers*/
+                    /*there's a guaranteed ':' in the completeHeader, by HTTP_HEADERS module*/
+                    colon_pos = strchr(completeHeader, ':');
+                    if (colon_pos != NULL)
+                    {
+                        *colon_pos = '\0'; /*cut it down*/
```
jspaith: >/*cut it down*/ [](start = 43, length = 15)

Nit - I'm not sure we need the /*cut it down*/ coments all over place; this is fairly standard C string traversal pattern.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+                        if (Map_AddOrUpdate(properties, completeHeader + strlen(IOTHUB_APP_PREFIX), colon_pos + 2) != MAP_OK) /*whereIsColon+1 is a space because HTTPEHADERS outputs a ": " between name and value*/
```
jspaith: >colon_pos + 2 [](start = 100, length = 13)

On colon_pos+2, what happens if the string is of form

IoTHub_appct_whatever:\0.  Looks like we'd inject a 0 len string into these values, which I think would pass but not sure we'd want it to.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+        for (size_t index = 0; index < count; index++)
+        {
+            /*Codes_SRS_TRANSPORTMULTITHTTP_17_074: [Every property name shall add  to the message size the length of the property name + the length of the property value + 16 bytes.] */
+            *msg_size += (strlen(values[index]) + strlen(keys[index]) + MAXIMUM_PROPERTY_OVERHEAD);
+            if (*msg_size > MAXIMUM_MESSAGE_SIZE)
+            {
+                /*Codes_SRS_TRANSPORTMULTITHTTP_17_072: [The message size shall be limited to 255KB -1 bytes.] */
+                PDLIST_ENTRY head = DList_RemoveHeadList(deviceData->waitingToSend); /*actually this is the same as "actual", but now it is removed*/
+                DList_InsertTailList(&(deviceData->eventConfirmations), head);
+                handleData->transport_callbacks.send_complete_cb(&(deviceData->eventConfirmations), IOTHUB_CLIENT_CONFIRMATION_ERROR, deviceData->device_transport_ctx); // takes care of emptying the list too
```
jspaith: >handleData->transport_callbacks.send_complete_cb [](start = 16, length = 48)

I don't understand why this error path has the cleanup & callback logic (removing from list, IoTHubMessage_SetMessageId) but the error paths immediately below don't.  

Beyond the tactical level of this funciton, are there other places callbacks should be called?

In my opinion these should be as high up as they can in call stack, so just one place doing all the work on a callee error instead of sprinkling this logic all over.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+        result = __LINE__;
+    }
+    else
+    {
+        result = 0;
+    }
+    return result;
+}
+
+static bool set_message_properties(IOTHUB_MESSAGE_LIST* message, size_t* msg_size, HTTP_HEADERS_HANDLE headers, HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERDEVICE_DATA* deviceData)
```
jspaith: >set_message_properties [](start = 12, length = 22)

Nit - should this be "set_app_message_properties" or something to give sense it's not ALL possible stuff & especially not system?

jspaith:
> #define IOTHUB_APP_PREFIX "iothub-app-"

Nit - this should really be a "static const char*" unless its used with #def concats someplace else.  (And other #def's here too since we have hood open).


---
Refers to: iothub_client/src/iothubtransporthttp.c:29 in a8d93f7. [](commit_id = a8d93f710184c871e26c84cc6ae78468747101be, deletion_comment = False)

iothub_client/src/iothubtransporthttp.c
```diff
@@ -1540,199 +1744,76 @@ static void DoEvent(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERDEVI
...
-                            MAP_HANDLE map = IoTHubMessage_Properties(message->messageHandle);
-                            const char*const* keys;
-                            const char*const* values;
-                            size_t count;
-                            if (Map_GetInternals(map, &keys, &values, &count) != MAP_OK)
+                            // set_message_properties returning false does not necessarily mean the the function failed, it just means
+                            // the the adding of messages should not continue and should try the next time.  So you should not log if this 
+                            // returns false
+                            if (set_message_properties(message, &messageSize, clonedEventHTTPrequestHeaders, handleData, deviceData))
                             {
```
jspaith: >{ [](start = 28, length = 1)

Nit - extra space after this "{".

iothub_client/src/iothubtransporthttp.c
```diff
@@ -136,6 +138,208 @@ static bool create_eventHTTPrelativePath(HTTPTRANSPORT_PERDEVICE_DATA* handleDat
...
+                    }
+                }
+                free(completeHeader);
+            }
+        }
+    }
+    return result;
+}
+
+static int set_system_properties(IOTHUB_MESSAGE_LIST* message, HTTP_HEADERS_HANDLE headers)
```
jspaith: >set_system_properties [](start = 11, length = 21)

Why does the set_message_properties() have to respect the msg_size but the set_system_properties doesn't?

iothub_client/src/iothubtransporthttp.c
```diff
@@ -1540,199 +1744,76 @@ static void DoEvent(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERDEVI
...
-                                {
-                                    if (HTTPHeaders_ReplaceHeaderNameValuePair(clonedEventHTTPrequestHeaders, IOTHUB_CONTENT_ENCODING_D2C, contentEncoding) != HTTP_HEADERS_OK)
-                                    {
-                                        LogError("unable to HTTPHeaders_ReplaceHeaderNameValuePair (content-encoding)");
-                                        goOn = false;
-                                    }
-                                }
-
-                                if (!goOn)
+                                if (set_system_properties(message, clonedEventHTTPrequestHeaders) != 0)
```
jspaith: >set_system_properties [](start = 36, length = 21)

Nit - and I get I'm adding a lot of these given this is a good change as is :).  but this and some of other blocks could be moved to a

if (set_messag_propes() == fail)
{  LogFailure() ; }
else if (set_system_prorites == )
{...
}

iothub_client/src/iothubtransporthttp.c
```diff
@@ -1540,199 +1744,76 @@ static void DoEvent(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERDEVI
                         }
                         else
                         {
-                            /*Codes_SRS_TRANSPORTMULTITHTTP_17_078: [Every message property "property":"value" shall be added to the HTTP headers as an individual header "iothub-app-property":"value".] */
-                            MAP_HANDLE map = IoTHubMessage_Properties(message->messageHandle);
-                            const char*const* keys;
-                            const char*const* values;
-                            size_t count;
-                            if (Map_GetInternals(map, &keys, &values, &count) != MAP_OK)
+                            // set_message_properties returning false does not necessarily mean the the function failed, it just means
```
jspaith: >// set_message_properties returning false does not necessarily mean the the function failed, it just means [](start = 28, length = 106)

Hmm... if any of these calls fail, it feels we should just give up and log.  I don't necesarily think we retry on OOMs and other errors, like the message too long, aren't going to be helped by going through loop again.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -1540,199 +1744,76 @@ static void DoEvent(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERDEVI
...
-
-                                if (!goOn)
+                                if (set_system_properties(message, clonedEventHTTPrequestHeaders) != 0)
                                 {
                                     /*Codes_SRS_TRANSPORTMULTITHTTP_17_079: [If any HTTP header operation fails, _DoWork shall advance to the next action.] */
                                 }
                                 else
                                 {
-                                    BUFFER_HANDLE toBeSend = BUFFER_new();
+                                    BUFFER_HANDLE toBeSend = BUFFER_create(messageContent, originalMessageSize);
```
jspaith: >originalMessageSize [](start = 91, length = 19)

Not necessairly code you've touch, but I'm having trouble getting head around the MAXIMUM_PAYLOAD_OVERHEAD business.  It seems weird to me that we're coupling the payload and a max size here with anything HTTP header related, unless there's some sort of limit on server??

iothub_client/src/iothubtransporthttp.c
jebrando: This will have to be done in a separate CR

---
In reply to: [253702980](https://github.com/Azure/azure-iot-sdk-c/pull/843#discussion_r253702980) [](ancestors = 253702980)

iothub_client/src/iothubtransporthttp.c
jebrando: This whole file can be looked at.

---
In reply to: [253703651](https://github.com/Azure/azure-iot-sdk-c/pull/843#discussion_r253703651) [](ancestors = 253703651)

PR #857, opened 2019-02-08
Checking that the sas token is not set to an invalid value

jebrando:

### DIFF
--- a/iothub_client/src/iothub_client_authorization.c
+++ b/iothub_client/src/iothub_client_authorization.c
@@ -674,7 +674,7 @@ int IoTHubClient_Auth_Set_SasToken_Expiry(IOTHUB_AUTHORIZATION_HANDLE handle, si
     // Validate the expiry_time in seconds
     else if (expiry_time_seconds < MIN_SAS_EXPIRY_TIME)
     {
-        LogError("Failure setting expiry time to value %lu min value is %lu", (unsigned long)expiry_time_seconds, MIN_SAS_EXPIRY_TIME);
+        LogError("Failure setting expiry time to value %lu min value is %d", (unsigned long)expiry_time_seconds, MIN_SAS_EXPIRY_TIME);
         result = __FAILURE__;
     }
     else
--- a/iothub_client/src/iothub_client_authorization.c
+++ b/iothub_client/src/iothub_client_authorization.c
@@ -20,7 +20,7 @@
 
 #define DEFAULT_SAS_TOKEN_EXPIRY_TIME_SECS          3600
 #define INDEFINITE_TIME                             ((time_t)(-1))
-#define MIN_SAS_EXPIRY_TIME                         30  // 30 seconds
+#define MIN_SAS_EXPIRY_TIME                         5  // 5 seconds
 
 typedef struct IOTHUB_AUTHORIZATION_DATA_TAG
 {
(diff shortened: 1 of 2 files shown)


massand:
>             }

Can we move the initialization of IOTHUB_AUTHORIZATION_DATA to a common function? Due to the duplication of code for initializing device_id, we landed up with the Edge specific bugs here. 

---
Refers to: iothub_client/src/iothub_client_authorization.c:172 in 7024c55. [](commit_id = 7024c55bb9060f6f3a0677b7c4288507c6f2da8f, deletion_comment = False)

massand:
>             }

Again, the refactor can be done in a separate checkin. In the same refactor, we should address the fact that whenever we build the provisioning client, we also build all the module provisioning stuff. In the way we structure iothub_client,, the deviceclient and moduleclient code have clear API surfaces and there is a clear clientcore common obj to separate common code. Here we are using #Ifdefs, which is horrible way to manage this. We also don't appear to have any way to test this e2e without a massive Edge setup. Bert's framework sort of help us catch this. So, please revert Yoseph's test change once this is through. 

---
In reply to: [461970204](https://github.com/Azure/azure-iot-sdk-c/pull/857#issuecomment-461970204) [](ancestors = 461970204)

---
Refers to: iothub_client/src/iothub_client_authorization.c:172 in 7024c55. [](commit_id = 7024c55bb9060f6f3a0677b7c4288507c6f2da8f, deletion_comment = False)

iothub_client/src/iothub_client_authorization.c
```diff
@@ -20,6 +20,7 @@
 
 #define DEFAULT_SAS_TOKEN_EXPIRY_TIME_SECS          3600
 #define INDEFINITE_TIME                             ((time_t)(-1))
+#define MIN_SAS_EXPIRY_TIME                         30  // 30 seconds
```
massand: >MIN_SAS_EXPIRY_TIME [](start = 8, length = 19)

Why not 5 seconds? Or 10 seconds? 
Do we document the default of this SAS token refresh someplace in our header comments or devdocs?

iothub_client/src/iothub_client_authorization.c
```diff
@@ -676,6 +678,12 @@ int IoTHubClient_Auth_Set_SasToken_Expiry(IOTHUB_AUTHORIZATION_HANDLE handle, si
         LogError("Invalid handle value handle: NULL");
         result = __FAILURE__;
     }
+    // Validate the expiry_time in seconds
+    else if (expiry_time_seconds < MIN_SAS_EXPIRY_TIME)
+    {
+        LogError("Failure setting expiry time below the min value of %lu", (unsigned long)expiry_time_seconds);
```
massand: >lu [](start = 70, length = 2)

You need to change this to say "this is actual; value" and "you need to set it to something greater than MIN_SAS_EXPIRY_TIME"

iothub_client/src/iothub_client_authorization.c
massand: As soon as I typed this, our e2e tests failed because they set minimum 10 seconds for fault injection :) Please set this to 5 seconds. It should be fine. 

---
In reply to: [255260962](https://github.com/Azure/azure-iot-sdk-c/pull/857#discussion_r255260962) [](ancestors = 255260962)

jebrando: 30 seconds is arbitrary.  I had to pick a number.  I changed it to 5 seconds.

---
In reply to: [255263674](https://github.com/Azure/azure-iot-sdk-c/pull/857#discussion_r255263674) [](ancestors = 255263674,255260962)

jebrando:
>             }

I added this revision.

---
In reply to: [461972747](https://github.com/Azure/azure-iot-sdk-c/pull/857#issuecomment-461972747) [](ancestors = 461972747,461970204)

---
Refers to: iothub_client/src/iothub_client_authorization.c:172 in 7024c55. [](commit_id = 7024c55bb9060f6f3a0677b7c4288507c6f2da8f, deletion_comment = False)

PR #860, opened 2019-02-11
Add device ID to platform info when not using HTTP

jadeiceman:
This change is dependent on this PR: https://github.com/Azure/azure-c-shared-utility/pull/268

# Checklist

# Description of the problem
Adding the SQM ID telemetry to the product info string for non HTTP protocol only 

# Description of the solution
SQM ID is a Windows-only ID, so the change is only for Win32 and only when the transport protocol is not HTTP.


iothub_client/src/iothub_client_core_ll.c
```diff
@@ -27,6 +27,10 @@
 #include "internal/iothub_client_diagnostic.h"
 #include "internal/iothubtransport.h"
 
+#ifdef WIN32
+#include "iothubtransporthttp.h"
+#endif
+
```
jebrando: The transport code can not live in the client core.

PR #862, opened 2019-02-12
add use_mbedtls option

alexmrtn:
# Checklist
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down). -> not sure if this is applicable in this case
  - [ ] I edited the corresponding document in the `devdoc` folder and added or modified requirements. -> N/A

This PR has a dependency on the PR on the c-utility repo https://github.com/Azure/azure-c-shared-utility/pull/270

# Description of the problem
This add the ability to use mbedTLS as tls layer for all platforms 

# Description of the solution
This only propagate a flag on the underlying c-utility submodule.


massand:
Thanks for your patience on this PR. I was trying to debug a shutdown error before moving this in, but we are good here from the code changes point of view. So, approved! 

Here is the issue for reference - https://github.com/Azure/azure-c-shared-utility/issues/317

PR #874, opened 2019-02-20
Invalid token warnings

jspaith:
Add a `LogError` when an unknown token is provided.  Not making an error presently or doing stricmp instead of strcmp pending compat review.


massand:
@jspaith Please make the case-insensitive check a separate PR. This is definitely required in our logs regardless of the Edge connection string issue. Thanks!

Issue #651, opened 2018-09-26
segmentation fault on Azure SDK - name conflicts (device_create/device_destroy)


syedelec:
- **OS and version used:** <VERSION> 
Ubuntu 18.04

- **SDK version used:** <VERSION> 
master

# Description of the issue:

I simplified a sample of iothub_client to a simple Azure SDK call and added calls to function `device_create` which produces a Segmentation Fault. 

In the code, I am not calling the function, however it seems that it's called internally since I have the output : `inside device_create`

# Code sample exhibiting the issue:

```
#include <stdio.h>

#include <azureiot/iothub_client.h>
#include <azure_c_shared_utility/platform.h>
#include <azureiot/iothubtransportamqp_websockets.h>

void device_create(void);

void device_create(void) {
    printf("inside device_create\n");
}

int main(void)
{
    IOTHUB_CLIENT_HANDLE handle;
    const char *connection_string = "[connection_string]";

    platform_init();

    if ( (handle = IoTHubClient_CreateFromConnectionString(connection_string, AMQP_Protocol_over_WebSocketsTls)) != NULL )
    {
        printf("iot client handle worked\n");
    }
    else
    {
        printf("ERROR: handle is NULL!\r\n");
    }

    platform_deinit();
    return 0;
}
```

# Console log of the issue:

```
inside device_create
Segmentation fault (core dumped)
```

jebrando:
@syedelec We have a poorly named function in the SDK, we will look to change to something a little more descriptive.  If you are using c you should/could have the static keyword in front of the function to decrease the change of name collisions.

```C
static void device_create(void) {
    printf("inside device_create\n");
}
```

syedelec:
@jebrando Thank you, however this is a simple reproducible program to show the issue. In our case, we cannot consider using `static` functions because our functions are called across a larger project.

YoDaMa:
The PR #925 addresses this, so I’m going to go ahead and close this issue.

az-iot-builder-01:
@jebrando, @syedelec, @YoDaMa, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #925, opened 2019-03-26
Amqp refactor

YoDaMa:
In the AMQP Transport we have some generically named public functions that can easily conflict with public functions created by a customer. To prevent any confusing behavior, I have renamed all these functions with the prefix "iothubtransport_amqp_".

In reference to https://github.com/Azure/azure-iot-sdk-c/issues/651

### DIFF
--- a/iothub_client/inc/internal/iothubtransport_amqp_device.h
+++ b/iothub_client/inc/internal/iothubtransport_amqp_device.h
@@ -17,7 +17,7 @@ extern "C"
 {
 #endif
 
-// @brief    name of option to apply the instance obtained using device_retrieve_options
+// @brief    name of option to apply the instance obtained using amqp_device_retrieve_options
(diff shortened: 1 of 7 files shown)


iothub_client/inc/internal/iothubtransport_amqp_device.h
```diff
@@ -106,23 +106,23 @@ typedef struct DEVICE_CONFIG_TAG
```
ewertons: I would suggest renaming DEVICE_CONFIG to IOTHUBTRANSPORT_AMQP_DEVICE_CONFIG. Same for the _TAG above (rename DEVICE_CONFIG_TAG to IOTHUBTRANSPORT_AMQP_DEVICE_CONFIG_TAG).
Everything else is good. Thanks for this.

PR #946, opened 2019-04-10
Prov payload change

jebrando:
# Checklist

# Description of the problem

# Description of the solution


provisioning_client/src/prov_transport_amqp_common.c
```diff
@@ -360,22 +363,79 @@ static int set_amqp_msg_property(AMQP_VALUE uamqp_prop_map, const char* key, con
     return result;
 }
 
+static char* construct_json_data(PROV_TRANSPORT_AMQP_INFO* amqp_info)
```
ewertons: This function seems duplicated on http transport, changing only the way ek and srk are accessed. Maybe it could be put on a common module?

PR #950, opened 2019-04-12
Add extended platform info to product info string

jadeiceman:
# Checklist

# Description of the problem

# Description of the solution


iothub_client/inc/internal/iothub_transport_ll_private.h
```diff
@@ -90,6 +91,7 @@ extern "C"
     typedef int(*pfIoTHubTransport_Subscribe_InputQueue)(IOTHUB_DEVICE_HANDLE handle);
     typedef void(*pfIoTHubTransport_Unsubscribe_InputQueue)(IOTHUB_DEVICE_HANDLE handle);
     typedef int(*pfIoTHubTransport_SetCallbackContext)(TRANSPORT_LL_HANDLE handle, void* ctx);
+    typedef bool(*pfIoTHubTransport_IsExtraPlatformInfoRequired)(TRANSPORT_LL_HANDLE handle);
```
ewertons: @jadeiceman ,

first of all great job with the whole implementation, it is really well done and fits great with all our design guidelines.

I have one point though, and that will change the code quite a bit.
I do believe that a single function that says yes/no if a specific extra info is required per transport is, well, too specific. If later we want add more flags, we would then need to change this function or add new ones.

Instead, since you already defined an ENUM in c-utility for this data (SQM ID), I would suggest this function to be defined as :
```
typedef int(*pfIoTHubTransport_GetSupportedPlatformInfo)(TRANSPORT_LL_HANDLE handle, PLATFORM_INFO_OPTION* info);
``` 

Args:
handle: the transport handle
info: a pointer to the enum var; this is set by the function as a combination of enum values (flags).

Returns: 0 if the handle is not null, 1 (use MU_FAILURE internally) otherwise.

usage:
```c
PLATFORM_INFO_OPTION supportedPlatformInfo;
if (someTransport_GetSupportedPlatformInfo(transport_handle, &info) != 0)
{
    // Log and handle error.
}
else if ((supportedPlatformInfo & PLATFORM_INFO_OPTION_RETRIEVE_SQM) == PLATFORM_INFO_OPTION_RETRIEVE_SQM)
{
    // get the option like you did in the code below.
}
```

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -462,12 +463,19 @@ static bool invoke_message_callback(IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* handleDat
 
 /*Codes_SRS_IOTHUBCLIENT_LL_10_032: ["product_info" - takes a char string as an argument to specify the product information(e.g. `"ProductName/ProductVersion"`). ]*/
 /*Codes_SRS_IOTHUBCLIENT_LL_10_034: ["product_info" - shall store the given string concatenated with the sdk information and the platform information in the form(ProductInfo DeviceSDKName / DeviceSDKVersion(OSName OSVersion; Architecture). ]*/
-static STRING_HANDLE make_product_info(const char* product)
+static STRING_HANDLE make_product_info(const char* product, bool IsExtraPlatformInfoRequired)
```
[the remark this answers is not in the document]
jadeiceman: Whoops, I think this was a result of a find + replace.  Good catch.

iothub_client/inc/internal/iothub_transport_ll_private.h
jadeiceman: I was actually thinking this too, so I'll make the changes.

jadeiceman: I made the changes, but I didn't make the GetSupportedPlatformInfo function return 1 when the handle is null since I'm not using the handle for anything - having the handle as a parameter was to keep it consistent with the rest of the transport functions and have it in case we needed in the future.  I also updated the tests, but you might want to double check I updated them correctly.  I ran them locally and they passed though.

massand:
> All looks good.
> Just one thing: the GetSupportedPlatformInfo should be implemented in the iothubtransport_amqp_common and iothubtransport_mqtt_common and called from the specific transport implementations (iothubtransportamqp, iothubtransportamqp_websockets, iothubtransportmqtt, iothubtransportmqtt_websockets).
> 
> But that's alright, I'll make those changes.

I haven't seen your changes yet. are we waiting on you?

PR #965, opened 2019-04-22
Fix for gh issue 775 (twin/amqp not resubscribed after reconnection)

ewertons:
# Checklist

# Description of the problem

The IoT Hub client uses a module called iothubtransport_amqp_twin_messenger for handling Device Twin over AMQP. This module has, among other functions, a _start() and _stop() functions that are used by the iothub client AMQP transport to restart the iothubtransport_amqp_twin_messenger when a reconnection is necessary.
However, when iothubtransport_amqp_twin_messenger is started again after a reconnection, it should resubscribe automatically to receive Twin updates, if it was previously subscribed by the user.
There is a bug in that module that prevented the re-subscription.

# Description of the solution
 

The internal state of the messenger needs to be moved back to TWIN_SUBSCRIPTION_STATE_GET_COMPLETE_TWIN... when _stop is called, so when _start is called the re-subscription occurs.


iothub_client/src/iothubtransport_amqp_twin_messenger.c
```diff
@@ -2115,6 +2115,11 @@ int twin_messenger_stop(TWIN_MESSENGER_HANDLE twin_msgr_handle)
         }
         else
         {
+            if (twin_msgr->subscription_state != TWIN_SUBSCRIPTION_STATE_UNSUBSCRIBE)
```
YoDaMa: There is no comment added in this PR clarifying the purpose of this check, and it's not initially clear without knowing the GH Issue in reference. Can you add a comment specifying why this if statement and set is necessary?

ewertons: Sure, please read the description, I added the details there.

YoDaMa: yes but the description will not be in the code :) this is for future developer sanity. One line will do:

// if previously subscribed, twin_messenger_stop moves the subscription state to resubscribe automatically to receive twin updates

ewertons: I'll add the requirement docs.

massand:
@ewertons - do we have an e2e test for this disconnect-reconnect scenario? If not, please add it before checking in. Thanks.

ewertons:
@massand , no there is no e2e test for this.
I can add the test, but since this is a new test implementation it might take some time (it will be in the fault-injection tests).

PR #1004, opened 2019-05-14
Fix amqp property value that was being release early

jebrando:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/provisioning_client/inc/azure_prov_client/prov_client_const.h
+++ b/provisioning_client/inc/azure_prov_client/prov_client_const.h
@@ -6,7 +6,7 @@
 
 #define PROV_DEVICE_CLIENT_VERSION "1.3.2"
 #define PROV_GET_THROTTLE_TIME      1
-#define MAX_PROV_GET_THROTTLE_TIME  2*60
+#define MAX_PROV_GET_THROTTLE_TIME  5*60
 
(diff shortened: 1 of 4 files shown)


massand:
>                     else if (type == AMQP_TYPE_STRING)

Please copy the "when this is a string vs int" comment here 

---
Refers to: provisioning_client/src/prov_transport_amqp_common.c:364 in 06c122a. [](commit_id = 06c122aacaba082cf65243dfcfbdcb9596a6aece, deletion_comment = False)

massand:
>                     if (desc != NULL && amqpvalue_get_type(desc) == AMQP_TYPE_ULONG)

Questioning why this is a ULONG on error? Please add comment here and clarify with service. 

---
Refers to: provisioning_client/src/prov_transport_amqp_common.c:254 in 06c122a. [](commit_id = 06c122aacaba082cf65243dfcfbdcb9596a6aece, deletion_comment = False)

massand:
>                             amqp_info->retry_after_value = (uint32_t)desc_value;

I think its prudent to set this to UINT32_MAX here if there is a value larger here. 

---
Refers to: provisioning_client/src/prov_transport_amqp_common.c:259 in 06c122a. [](commit_id = 06c122aacaba082cf65243dfcfbdcb9596a6aece, deletion_comment = False)

PR #1006, opened 2019-05-15
Horton patch

YoDaMa:
We need to point to a single version of Horton so that the build does not break in the event of a Horton change.


massand:
@YoDaMa - I need more context. Who will manage this snap branch we are pointing to? Is this temporary or permanent?

YoDaMa:
I'm going to close this PR. This was going to be a temporary fix so that folks could get their merges in, but we ended up getting the permanent fix in sooner.

PR #1035, opened 2019-06-03
(fix) change string options to CACHE STRING

YoDaMa:
## Description

The CMakeLists.txt file has three options that are defined in BOOL format but anticipate input as strings. This change converts the CMakeCache expected value from BOOL to STRING. 

## Reasoning

CMake defines the `option` module as a way to default BOOL values to OFF. Just BOOL values. Not strings. For the string options being used, they should not be defined as `option` modules. 

More information on this can be found in this SO post: https://stackoverflow.com/questions/36358217/what-is-the-difference-between-option-and-set-cache-bool-for-a-cmake-variabl

**Thank you for letting me contribute to your open source project as a lowly Node.JS developer!!**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,9 +74,9 @@ option(use_prov_client "Enable provisioning client" OFF)
 option(use_tpm_simulator "tpm simulator type of hsm used with the provisioning client" OFF)
 option(use_edge_modules "Enable support for running modules against Azure IoT Edge" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
-set(compileOption_C CACHE STRING "passes a string to the command line of the C compiler")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -77,6 +74,9 @@ option(use_prov_client "Enable provisioning client" OFF)
 option(use_tpm_simulator "tpm simulator type of hsm used with the provisioning client" OFF)
 option(use_edge_modules "Enable support for running modules against Azure IoT Edge" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
+set(compileOption_C "" CACHE STRING "passes a string to the command line of the C compiler")
```
ewertons: Be noted that looks like we don't have any build scripts that actually test this option. Please reply if that's not correct.

PR #1041, opened 2019-06-06
Cap the maximum load of make build to 80% of number of CORES

ewertons:
# Checklist

# Description of the problem
The gate has been failing on Debian builds, likely because we tell make to use a number of processes that takes all cores, resulting in exhaustion of resources.

# Description of the solution
The solution is to cap the maximum load of make to a number less than the max number of cores. Used 80% as an arbitrary number on this change.

### DIFF
--- a/jenkins/debian_c.sh
+++ b/jenkins/debian_c.sh
@@ -18,11 +18,18 @@ build_folder=$build_root"/cmake/debian_linux"
 # Set the default cores
 CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
 
+if [ $CORES -eq 1 ]
+then
+    LOADLIMIT=$CORES
+else
+    LOADLIMIT=$(perl -e "print int($CORES * 0.8)")
+fi
+
 rm -r -f $build_folder
 mkdir -p $build_folder
 pushd $build_folder
 cmake -Drun_valgrind:BOOL=ON $build_root -Drun_unittests:BOOL=ON
-make --jobs=$CORES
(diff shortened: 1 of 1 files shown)


jenkins/debian_c.sh
```diff
@@ -18,11 +18,18 @@ build_folder=$build_root"/cmake/debian_linux"
...
+else
+    LOADLIMIT=$(perl -e "print int($CORES * 0.8)")
+fi
+
 rm -r -f $build_folder
 mkdir -p $build_folder
 pushd $build_folder
 cmake -Drun_valgrind:BOOL=ON $build_root -Drun_unittests:BOOL=ON
-make --jobs=$CORES
+make --jobs=$CORES --max-load=$LOADLIMIT
```
jspaith: Shouldn't we put this on all our jenkins/*.sh files, just for uniformity?  Ideally with a helper .sh that had a function() to calculate that though we could take an IOU on that one given need to get this in.

jenkins/debian_c.sh
```diff
@@ -18,11 +18,18 @@ build_folder=$build_root"/cmake/debian_linux"
 # Set the default cores
 CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
 
+if [ $CORES -eq 1 ]
+then
+    LOADLIMIT=$CORES
+else
+    LOADLIMIT=$(perl -e "print int($CORES * 0.8)")
```
massand: >LOADLIMIT [](start = 4, length = 9)

Please print this value and CORES value for reference. We will need it for debugging on Azure DevOps hosted VMs which have variance in I/O and cores. We have no control on that.

Issue #1022, opened 2019-05-21
Support for private keys that live on crypto devices


OYTIS:
Greetings,

we've got a device that can perform security operations with private keys never leaving the device, and we'd like to have it supported with Azure IoT SDK. The problem is in the Azure SDK private key is abstracted as a string, and all existing tlsio adapters interpret this string as plain text PEM key, which is not what we'd like to have.

There was a [similar request](https://github.com/Azure/azure-iot-sdk-c/issues/509) about a year ago, and the proposed solution was to implement new tlsio adapter that would interpret the string it gets in a special way, and interact with the security device.

The problem I see with this solution is that existing adapters already contain a lot of code not related to private key management, and creating a duplicate with changed private key logic would make poor architecture and increase the maintenance burden. Moreover, tlsio adapters abstract TLS libraries, not crypto devices (which are in turn abstracted by `HSM_CLIENT_X509_INTERFACE`).

What I would propose is adding private key abstraction to tlsio, that would represent a private key in a more generic way, e.g.

```
struct tlsio_private_key {
  int (*sign)(const char*, int, char*, int*, void*); // (hash, length, output, out_length, priv) -> result
  int (*decrypt)(const char*, int, char*, int*, void*); // (data, length, output, out_length, priv) -> result
  union {
    struct rsa_params {...};
    struct ecc_params {...};
  };
};
```

It could then be converted by the adapters to a library-specific private key representation.

It's definitely quite a few changes to multiple repos (c-utility, uhttp, c SDK), but will also improve the security dramatically.

Would be grateful for any feedback, guidance and help.

CIPop:
Thanks @OYTIS. I have recently joined the C SDK group (please bare with me as I ramp-up). We have discussed the proposal and came up with the following functional requirements:
1. No new public APIs should be added (if possible).
2. Our SDK repo will not contain the custom HSM code. (We cannot maintain drivers specific to certain hardware but we can implement some of the existing standards such as PKCS#11, TPM etc.)
3. A new 3rd party custom HSM module must be easy to integrate with our SDK. This should be achievable without having access to source-code of the driver.
4. The HSM custom module be usable with or without DPS (e.g. X509 auth with the IoT Hub).

The plan for the current SDK is to offer a way to implement a custom TLS xio that is configurable by the application and can take full advantage of the HSM features. Some (if not all the configuration) should be passed through existing APIs that can be repurposed (e.g. the cert key could be a slot# in the HSM).

[...]

PR #1058, opened 2019-06-21
Support DPS provisioning with non-extractable private keys

OYTIS:
Related to https://github.com/Azure/azure-iot-sdk-c/issues/1022

Depends on https://github.com/Azure/azure-c-shared-utility/pull/327 and https://github.com/Azure/azure-uhttp-c/pull/33

No unit tests yet, waiting for dependencies to be reviewed.


CIPop:
Thank you for the contribution @OYTIS ! We acknowledge this is missing from our current SDK implementation.

We have decided to investigate a different solution to the problem of passing information to custom HSMs that would not require a new API such as `IoTHubClient_Auth_Set_HSM_Data` but instead try to reuse existing `iothub_client_authorization` APIs such as `IoTHubClient_Auth_Set_xio_Certificate`. We will also expose a new compilation mode to allow 3rd parties to use their custom xio TLS module implementation from their own repos via CMake flags.
Given the high variability of HSM configuration modes, the custom xio TLS module can be configured outside of the SDK if necessary, or through the SDK passing data from the existing APIs. 
This new HSM mode of operation should allow non-DPS scenarios (directly connecting to IoT Hub using an HSM).

PR #1076, opened 2019-07-05
Move real parson

jspaith:
Move `real_parson.*` to common location, instead of buried in serializer UT directory.


serializer/tests/datamarshaller_ut/CMakeLists.txt
```diff
@@ -11,10 +11,10 @@ set(${theseTestsName}_test_files
...
 
-include_directories(${SHARED_UTIL_REAL_TEST_FOLDER})
+include_directories(${SHARED_UTIL_REAL_TEST_FOLDER} ${IOTHUB_TEST_REAL_INC_FOLDER})
 
 set(${theseTestsName}_c_files
-    real_parson.c
+    ${IOTHUB_TEST_REAL_SRC_FOLDER}/real_parson.c
     ${SHARED_UTIL_REAL_TEST_FOLDER}/real_vector.c
     ${SHARED_UTIL_REAL_TEST_FOLDER}/real_crt_abstractions.c
     ${SHARED_UTIL_REAL_TEST_FOLDER}/real_strings.c
```
massand: These dependencies have now been moved up directly into azure-sdk-iot-c. We should move the corresponding real test files of these dependencies into the same IOTHUB_TEST_REAL_SRC_FOLDER.

massand: On second thoughts, Jelani convinced me that the real_component.c and real_component.h files belong closer to the component.c code itself. NOT closer to the test. I am convinced. Signing off!

---
In reply to: [302683688](https://github.com/Azure/azure-iot-sdk-c/pull/1076#discussion_r302683688) [](ancestors = 302683688)

PR #1094, opened 2019-08-02
Remove hardcoded key name usage from iothub_messaging_ll.c

ewertons:
# Checklist

# Description of the problem

The service client for cloud-to-device messaging constructed a token authentication id using a hardcoded name of the policy owner.
Only policies with owner "iothubowner" would then work with this client.

# Description of the solution
 

This change removes this hard-coded name, consuming the policy owner name as provided by the user through the IoT Hub connection string.

The service client has no tests in general, so no automated tests were added. 
This fix has been tested manually though with different IoT Hub policies (different owner names).

### DIFF
--- a/iothub_service_client/src/iothub_messaging_ll.c
+++ b/iothub_service_client/src/iothub_messaging_ll.c
@@ -457,15 +457,16 @@ static char* createAuthCid(IOTHUB_MESSAGING_HANDLE messagingHandle)
     }
     else
     {
-        const char* AMQP_SEND_AUTHCID_FMT = "iothubowner@sas.root.%s";
-        size_t authCidLen = strlen(AMQP_SEND_AUTHCID_FMT) + strlen(messagingHandle->iothubName);
+        const char* AMQP_SEND_AUTHCID_FMT = "%s@sas.root.%s";
+        const int AMQP_SEND_AUTHCID_FMT_LENGTH = 10;
+        size_t authCidLen = strlen(messagingHandle->keyName) + AMQP_SEND_AUTHCID_FMT_LENGTH + strlen(messagingHandle->iothubName);
 
         if ((buffer = (char*)malloc(authCidLen + 1)) == NULL)
         {
             LogError("Malloc failed for authCid.");
             result = NULL;
         }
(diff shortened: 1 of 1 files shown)


iothub_service_client/src/iothub_messaging_ll.c
```diff
@@ -457,15 +457,16 @@ static char* createAuthCid(IOTHUB_MESSAGING_HANDLE messagingHandle)
     }
     else
     {
-        const char* AMQP_SEND_AUTHCID_FMT = "iothubowner@sas.root.%s";
-        size_t authCidLen = strlen(AMQP_SEND_AUTHCID_FMT) + strlen(messagingHandle->iothubName);
+        const char* AMQP_SEND_AUTHCID_FMT = "%s@sas.root.%s";
+        const int AMQP_SEND_AUTHCID_FMT_LENGTH = 10;
```
jebrando: >10 [](start = 49, length = 2)

Should this be changed to strlen(AMQP_SEND_AUTHCID_FMT)?

jebrando: >int [](start = 14, length = 3)

Should this be size_t

iothub_service_client/src/iothub_messaging_ll.c
```diff
@@ -457,15 +457,16 @@ static char* createAuthCid(IOTHUB_MESSAGING_HANDLE messagingHandle)
     }
     else
     {
-        const char* AMQP_SEND_AUTHCID_FMT = "iothubowner@sas.root.%s";
-        size_t authCidLen = strlen(AMQP_SEND_AUTHCID_FMT) + strlen(messagingHandle->iothubName);
+        const char* AMQP_SEND_AUTHCID_FMT = "%s@sas.root.%s";
```
danewalton: >const char* AMQP_SEND_AUTHCID_FMT [](start = 8, length = 33)

i usually prefer using `const char[]`  because it allows the next line to be a `sizeof()` instead of needing to count the characters by hand.

iothub_service_client/src/iothub_messaging_ll.c
danewalton: better yet sizeof() :)

---
In reply to: [310311464](https://github.com/Azure/azure-iot-sdk-c/pull/1094#discussion_r310311464) [](ancestors = 310311464)

Issue #1109, opened 2019-08-12
iothubmessage-getstring needs better documentation


ericwol-msft:
**Repro:**
https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-message-h/iothubmessage-getstring

**Issue:**
Need to document who owns freeing the string.

gregman-msft:
Proposed PR to header from which this document is generated: https://github.com/Azure/azure-iot-sdk-c/pull/1121

danewalton:
The PR has been merged

az-iot-builder-01:
@ericwol-msft, @gregman-msft, @danewalton, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1121, opened 2019-08-16
clarifying memory responsibilities for _getstring

gregman-msft:
in response to github issue #1109

@ewertons @ericwol-msft


ericwol-msft:
what about documenting other methods that return a string as well??

IoTHubMessage_GetContentTypeSystemProperty
IoTHubMessage_GetContentEncodingSystemProperty
IoTHubMessage_GetProperty

Issue #1111, opened 2019-08-12
 iothub-client-core-common-h enum values lack any documentation


ericwol-msft:
**Repro:**
https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-client-core-common-h
What is IOTHUB_CLIENT_CONFIRMATION_BECAUSE_DESTROY???

**Issue:**
All these enum values need code/public documentation.

gregman-msft:
@ericwol-msft I added definitions for the ones that are not self explanatory, please see the PR

jspaith:
Doing this correctly requires fixing the MU_DEFINE_ENUM stuff, since doxygen won't play nicely with what we have.  We have a draft PR of this up at https://github.com/Azure/azure-iot-sdk-c/pull/1342 which may or may not come in but it's blocked until then.

az-iot-builder-01:
@ericwol-msft, @gregman-msft, @jspaith, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1123, opened 2019-08-16
clarified complex enumerations

gregman-msft:
in response to github issue #1111

### DIFF
--- a/iothub_client/inc/iothub_client_core_common.h
+++ b/iothub_client/inc/iothub_client_core_common.h
@@ -38,20 +38,19 @@ extern "C"
 
     typedef void(*IOTHUB_METHOD_INVOKE_CALLBACK)(IOTHUB_CLIENT_RESULT result, int responseStatus, unsigned char* responsePayload, size_t responsePayloadSize, void* context);
 
-#define IOTHUB_CLIENT_RETRY_POLICY_VALUES     \
-    IOTHUB_CLIENT_RETRY_NONE,                   \
-    IOTHUB_CLIENT_RETRY_IMMEDIATE,                  \
-    IOTHUB_CLIENT_RETRY_INTERVAL,      \
-    IOTHUB_CLIENT_RETRY_LINEAR_BACKOFF,      \
-    IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF,                 \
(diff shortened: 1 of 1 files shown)


iothub_client/inc/iothub_client_core_common.h
```diff
@@ -38,20 +38,19 @@ extern "C"
...
+*/
+MU_DEFINE_ENUM(IOTHUB_CLIENT_RETRY_POLICY, 
+    IOTHUB_CLIENT_RETRY_NONE,                               /**<No re-connections are ever attempted*/
+    IOTHUB_CLIENT_RETRY_IMMEDIATE,                          /**<Re-connections shall be tried immediately, no wait time*/          
+    IOTHUB_CLIENT_RETRY_INTERVAL,                           /**<First re-connection attempted immediately, fixed-interval between subsequent retries*/
+    IOTHUB_CLIENT_RETRY_LINEAR_BACKOFF,                     /**<First re-connection attempted immediately, linear increase between subsequent retries*/
+    IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF,                /**<First re-connection attempted immediately, delay doubles between subsequent retries*/
+    IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF_WITH_JITTER,    /**<First re-connection attempted immediately, delay doubles with random jitter deduction between subsequent retries
+    IOTHUB_CLIENT_RETRY_RANDOM                              /**<First re-connection attempted immediately, random delay between 0 and 5 seconds between subsequent retries*/
+);
```
rafilho: I don't think this will work. We should just add those comments on the defines above.

danewalton: the main problem with your suggestion is that doxygen won't pick those up which i believe is where the original problem stems from. we will have to figure out some other way to rectify this. we can talk about it but i believe the reason this is generated this way is so that the enums can be printed as strings when the errors are thrown. there is a possibility that we could simplify the documentation by doing it this way but the side effect would be errors on the output would be numbers instead of strings. will have to double check with jelani

---
In reply to: [322363358](https://github.com/Azure/azure-iot-sdk-c/pull/1123#discussion_r322363358) [](ancestors = 322363358)

Issue #1136, opened 2019-08-21
Unable to connect to IoT Hub with RetryPolicy disabled


alexeystrakh:
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

Build Machine: Mac OS Mojave
Build for: iPhone X - iOS 12
Build Tools: Xcode 10.2

**SDK Version (Please Give Commit SHA if Manually Compiling)**

PODS:
  - AzureIoTHubClient (1.2.4):
  - AzureIoTuAmqp (1.2.4):
  - AzureIoTuMqtt (1.1.4):
  - AzureIoTUtility (1.1.4)

SPEC CHECKSUMS:
  AzureIoTHubClient: 343127ac6edc62af3fe62221243a2ee8e906b5fb
  AzureIoTuAmqp: 093aceeea0dc17e38e9fbd785ce2c3a7a7346065
  AzureIoTuMqtt: 99005699e0c575d47a2c3878e95769562991f83b
  AzureIoTUtility: cf0e15e2342f7c28f9798f3330ccb5d60950fede

**Protocol**

MQTT

**Describe the Bug**

Our solution works fine with the default configuration but when we try to disable retry policy we are unable to connect to the IoT Hub and send/receive messages. To disable retry policy we use the following code right after the hub client is created:

```
(void)IoTHubClient_LL_SetRetryPolicy(_iotHubClientHandle, IOTHUB_CLIENT_RETRY_NONE, 0);
```

**Console Logs**

```2019-08-19 23:31:18.402 [DEBUG] (9:Thread Pool Worker) [Cloud+IoTHub] : Calling IoT Hub connect from SendAndPurgeIoTHubMessages
2019-08-19 23:31:18.404 [DEBUG] (9:Thread Pool Worker) [Cloud+IoTHub] : Subscribed to IoT hub events
2019-08-19 23:31:18.405 [DEBUG] (9:Thread Pool Worker) [Cloud+IoTHub] : Connecting to IoT Hub [Hostname= ConnectionString= DeviceID= DeviceKey=
2019-08-19 23:31:18.405 [INFO] (9:Thread Pool Worker) [IoTHub] : IoT Hub connection string is null
2019-08-19 23:31:22.638 [DEBUG] (11:Thread Pool Worker) [Cloud+IoTHub] : Connecting to IoT Hub ...
[...]

PR #1151, opened 2019-08-28
Fix #924; always delete xio in MQTT cleanup

### LINKED ISSUE #924: Version 1.2.14 Memory Leak when client handle is released while MQTT trying to reconnect
**Development Machine, OS, Compiler**

STM32F405, FreeRTOS, GCC

**SDK Version**

1.2.14

**Protocol**

MQTT

**Describe the Bug**

So, this has been kicking my butt for the past week and a half in regards to Azure IoT Hub C SDK Version 1.2.14 (latest recommended by Microsoft).  A 2K leak when MQTT is in constant reconnect state and IoT Hub client handle is destroyed.  /*This*/ code was originally in the if statement and I replaced it with this code (transport_data->xioTransport).  In this scenario, the currPacketState can be DISCONNECT_TYPE, with transport still allocated and needing to be freed.  The file is Mesh_FwLib_STM32CubeExpansion_Cloud_AZURE\Middlewares\Third_Party\azure-iot-sdk-c\iothub_client\src\iothubtransport_mqtt_common.c

```
static void DisconnectFromClient(PMQTTTRANSPORT_HANDLE_DATA transport_data)
{
    if (transport_data->xioTransport/*transport_data->currPacketState != DISCONNECT_TYPE*/)
    {
        if (!transport_data->isDestroyCalled)
        {
            OPTIONHANDLER_HANDLE options = xio_retrieveoptions(transport_data->xioTransport);
            set_saved_tls_options(transport_data, options);
        }
        // Ensure the disconnect message is sent
        if (transport_data->mqttClientStatus == MQTT_CLIENT_STATUS_CONNECTED)
        {
            transport_data->disconnect_recv_flag = 0;
            (void)mqtt_client_disconnect(transport_data->mqttClient, mqtt_disconnect_cb, &transport_data->disconnect_recv_flag);
            size_t disconnect_ctr = 0;
            do
            {
                mqtt_client_dowork(transport_data->mqttClient);
                disconnect_ctr++;
                ThreadAPI_Sleep(50);
            } while ((disconnect_ctr < MAX_DISCONNECT_VALUE) && (transport_data->disconnect_recv_flag == 0));
        }
        xio_destroy(transport_data->xioTransport);
        transport_data->xioTransport = NULL;

        transport_data->device_twin_get_sent = false;
        transport_data->mqttClientStatus = MQTT_CLIENT_STATUS_NOT_CONNECTED;
        transport_data->currPacketState = DISCONNECT_TYPE;
    }
}
```

jspaith:
Fixes #924 

At issue is that there exists a path on MQTT destruction where `xio_destroy` is not invoked, causing a memory leak:
* MQTT enters a disconnected state and stays there.  E.G. `currPacketState` = `DISCONNECT_TYPE`.
* App calls through to a `IoTHubTransport_MQTT_Common_Destroy`. 
* The `DisconnectFromClient` call we're relying on will NOT free up the xio, because it's in `DISCONNECT_TYPE`.

The fix here is to make `free_transport_handle_data` (which is effectively the destructor of this handle) *always* destroy the xio.  This gets us an unconditional free so we're not dependent on other state to free up data.

I did further code review around this and saw that `conn_attempted` was set to TRUE but never reset on the xio going down.  Since the `conn_attempted` feels implicitly joined to a given xio instance, make it so that destroying an xio will reset.

Also adding a UT here.  I was able to repro @maskimthedog issue pre-fix (Valgrind showed a leak running this regression specific UT) and verified via Valgrind that it's addressed post fix.


jspaith:
@danewalton as a heads up, #1136 had a conflict here where you started using `conn_attempted` for retry logic whereas I had been resetting it always on xio destroy.

I've clarified that `conn_attempted` needs to stay set after initial connect attempt (since otherwise it seemed like it should've been set false on a destroy) in comments and removed my over-aggressive clearing.

PR #1156, opened 2019-08-30
Additional PnP bug fixes

jspaith:
Fixes that had been blocked on required changes going into master first.

* Have SDKInformation upload have its own callback for SDKInfo specific log messages.
* Re-enable fault injection tests.

### DIFF
--- a/digitaltwin_client/src/dt_client_core.c
+++ b/digitaltwin_client/src/dt_client_core.c
@@ -666,6 +666,22 @@ static DT_SEND_TELEMETRY_CALLBACK_CONTEXT* CreateDTSendTelemetryCallbackContext(
     return dtSendTelemetryCallbackContext;
 }
 
+// ReportedSdkInfo_Callback is invoked when SendSdkInformation property is acknowledged. Because this is best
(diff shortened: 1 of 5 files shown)


digitaltwin_client/src/dt_client_core.c
```diff
@@ -666,6 +666,22 @@ static DT_SEND_TELEMETRY_CALLBACK_CONTEXT* CreateDTSendTelemetryCallbackContext(
     return dtSendTelemetryCallbackContext;
 }
 
+// ReportedSdkInfo_Callback is invoked when SendSdkInformation property is acknowledged. Because this is best
+// effort only, we simply log the result for diagnostics purposes but take no further action.
+static void ReportedSdkInfo_Callback(int status_code, void* userContextCallback)
```
massand: I would like to nit pick on the added storage cost for this one. Can we pass some enum (with interface type = SDKINFORMATION) or boolean (isSDKInfo) with the context parameter to indicate that this is a "special" interface?

jspaith: Per offline discussion, it could go either way but wanting to keep the standard property update path as free from extra complexity as possible drove this decision.

PR #1169, opened 2019-09-09
Public preview pnp fixes

jspaith:
Another batch of small/medium sized features.  This one changes API contract.  In contact with codegen specifically before we take this into public-preview branch they base off of.

Changes:
* Make it so that UpdateAsyncCommandAsync itself has a callback function, following conventions that all *Async need a cb.
* No longer overload command/property context pointers in callback off of that which was created at handle create time.  Got feedback this was confusing, and it certainly breaks the normal standard other places.  Instead have the command callback function pointer set take it its context, and same with properties.
* Check for interface instance names being duplicated.  As comments indicate we normally don't do policy checks like this on client side, but given how unpleasant it'd be for partners to debug errors it's far better in this case to do so.
* Fix up many spots where UT should have had form like ASSERT(expected, actual) instead were coded up as ASSERT(actual, expected).  
This doesn't make a functional difference, but created obnoxiousness where the output on a fail was "Expected Value=<what was actual> but actual=<what was the expected>.


digitaltwin_client/inc/digitaltwin_interface_client.h
```diff
@@ -204,21 +204,31 @@ typedef void(*DIGITALTWIN_REPORTED_PROPERTY_UPDATED_CALLBACK)(DIGITALTWIN_CLIENT
...
-           status changes.  It is also used when commands and properties are invoked on this interface, assuming the interface accepts such callbacks.  This allows
-           the context to store state for the device application that spans the life of the interface and is not tied to a single call.
-
   @param[in]  interfaceId                       The interfaceId of interface to be registered.  For example, urn:contoso:com:EnvironmentalSensor:1.
   @param[in]  interfaceInstanceName             The interface instance name associated with this interface Id.  For example, environmentalSensor.
   @param[in]  dtInterfaceRegisteredCallback     (Optional) Function pointer to be invoked when the interface status (e.g. is registered or fails to register) changes.
-  @param[in]  userInterfaceContext              (Optional) Context pointer passed to dtInterfaceRegisteredCallback function on interface change, command updates, and property updates.
+  @param[in]  userInterfaceContext              (Optional) Context pointer passed to dtInterfaceRegisteredCallback function on interface changes.
```
massand: >userInterfaceContext [](start = 14, length = 20)

Nit: can we keep this consistent with other callbacks by naming it userContextCallback

digitaltwin_client/src/dt_client_core.c
```diff
@@ -799,6 +799,58 @@ static DIGITALTWIN_CLIENT_RESULT SendInterfacesToRegisterMessage(DT_CLIENT_CORE*
     return result;
 }
 
+// VerifyInterfaceInstancesUnique makes sure that the interfaceInstanceName of each handle is unique.  E.G. one connection cannot have
```
massand: >E.G. [](start = 103, length = 4)

nit: e.g. instead of E.G.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -799,6 +799,58 @@ static DIGITALTWIN_CLIENT_RESULT SendInterfacesToRegisterMessage(DT_CLIENT_CORE*
...
+// to instances both named "frontCamera".  We check here, instead of solely relying on server policy, because if this fails on server
```
massand: >to [](start = 3, length = 2)

spelling nit: two

digitaltwin_client/src/dt_client_core.c
```diff
@@ -799,6 +799,58 @@ static DIGITALTWIN_CLIENT_RESULT SendInterfacesToRegisterMessage(DT_CLIENT_CORE*
...
+// side for MQTT then the connection will be dropped and it will be hard for application developer to debug.
```
massand: >the connection will be dropped [](start = 22, length = 30)

Is it worth documenting the error handling protocol from service here? If it is, I recommend changing verbiage from WILL to MAY?

digitaltwin_client/src/dt_client_core.c
```diff
@@ -799,6 +799,58 @@ static DIGITALTWIN_CLIENT_RESULT SendInterfacesToRegisterMessage(DT_CLIENT_CORE*
...
+        {
+            const char* interfaceInstanceName2 = DT_InterfaceClient_GetInterfaceInstanceName(dtInterfaces[j]);
+            if (interfaceInstanceName2 == NULL)
+            {   
+                LogError("DT_InterfaceClient_GetInterfaceInstanceName failed on element %d in handle list", j);
+                result = DIGITALTWIN_CLIENT_ERROR;
+                break;
+            }
+
+            if (strcmp(interfaceInstanceName1, interfaceInstanceName2) == 0)
```
massand: >strcmp [](start = 16, length = 6)

Please double check server case-sensitivity standard for instance names. Worth adding a comment here to specify server protocols. This is one area I miss our requirements comments :)

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -1465,12 +1504,13 @@ static DIGITALTWIN_CLIENT_RESULT DT_InterfaceClient_SetAsyncResponseProperties(I
...
 // DigitalTwin_InterfaceClient_UpdateAsyncCommandStatusAsync is invoked by application when an async DigitalTwin command is executing, updating
 // DigitalTwin server with status information provided by caller in a specially formatted telemetry message.
-DIGITALTWIN_CLIENT_RESULT DigitalTwin_InterfaceClient_UpdateAsyncCommandStatusAsync(DIGITALTWIN_INTERFACE_CLIENT_HANDLE dtInterfaceClientHandle, const DIGITALTWIN_CLIENT_ASYNC_COMMAND_UPDATE* dtClientAsyncCommandUpdate)
+DIGITALTWIN_CLIENT_RESULT DigitalTwin_InterfaceClient_UpdateAsyncCommandStatusAsync(DIGITALTWIN_INTERFACE_CLIENT_HANDLE dtInterfaceClientHandle, const DIGITALTWIN_CLIENT_ASYNC_COMMAND_UPDATE* dtClientAsyncCommandUpdate, DIGITALTWIN_CLIENT_UPDATE_ASYNC_COMMAND_CALLBACK dtUpdateAsyncCommandCallback, void* userContextCallback)
 {
     DT_INTERFACE_CLIENT* dtInterfaceClient = (DT_INTERFACE_CLIENT*)dtInterfaceClientHandle;
     DIGITALTWIN_CLIENT_RESULT result;
     STRING_HANDLE jsonToSend = NULL;
     IOTHUB_MESSAGE_HANDLE telemetryMessageHandle = NULL;
+    DT_INTERFACE_SEND_TELEMETRY_CALLBACK_CONTEXT* sendTelemetryCallbackContext = NULL;
```
massand: >DT_INTERFACE_SEND_TELEMETRY_CALLBACK_CONTEXT [](start = 4, length = 44)

I notice that we are repurposing the DT_INTERFACE_SEND_TELEMETRY_CALLBACK_CONTEXT structure here for both async commands and sending telemetry. I understand the logical structure is the same, but I would rather typedef the DT_INTERFACE_SEND_TELEMETRY_CALLBACK_CONTEXT to something like DT_INTERFACE_ASYNC_COMMAND_CALLBACK_CONTEXT. This way it is apparent to SDK devs that it isn't in error.

digitaltwin_client/src/dt_client_core.c
jspaith: I double checked before coding this :).  But agree a comment here is nice.

---
In reply to: [322953947](https://github.com/Azure/azure-iot-sdk-c/pull/1169#discussion_r322953947) [](ancestors = 322953947)

digitaltwin_client/src/dt_interface_client.c
jspaith: DT_INTERFACE_SEND_TELEMETRY_CALLBACK_CONTEXT is an internal only structure in this file.  If you look at API shape for app developer, they are different function callbacks exactly so that it's clear to SDK which to use.

I'd noodled over naming on this and the issue is at end of day this is a send telemetry req at protocol level.  I have to have the structure be the same (without much bigger changes, including to dt_client_core.c) because it blindly shoots back telemetry ACK's via DT_InterfaceClient_ProcessTelemetryCallback.  The intf client then handles de-muxing.

---
In reply to: [322957430](https://github.com/Azure/azure-iot-sdk-c/pull/1169#discussion_r322957430) [](ancestors = 322957430)

digitaltwin_client/inc/digitaltwin_interface_client.h
jspaith: I need to sleep on this one and will check code in as it is for now.  It's a good point, but my thinking had been this is quite different than other cb's so was trying to name specify.  But "user interface" business could be problems.

---
In reply to: [322951649](https://github.com/Azure/azure-iot-sdk-c/pull/1169#discussion_r322951649) [](ancestors = 322951649)

PR #1173, opened 2019-09-12
Public preview pnp block on destroy

jspaith:
If an application calls `DigitalTwin_InterfaceClient_Destroy`, we need to block return of the destroy until all callbacks have completed processing.  It doesn't add that much code and protects device application from having to synchronize this on their own.

Fix is analogous to what the `client_core` layer already does with device level handles.  Record that we're in a callback and then if destroy happens at same time, poll on destroy.  

Because we have only a single dispatcher thread to process callbacks on, a simple bool to monitor callbacks being processed is sufficient complexity here.


digitaltwin_client/src/dt_interface_client.c
```diff
@@ -629,11 +700,19 @@ void DT_InterfaceClient_RegistrationCompleteCallback(DIGITALTWIN_INTERFACE_CLIEN
         {
             SetInterfaceState(dtInterfaceClient, DT_INTERFACE_STATE_REGISTERED);
         }
-        (void)InvokeBindingInterfaceUnlock(dtInterfaceClient, &lockHeld);
 
         if (dtInterfaceClient->dtInterfaceRegisteredCallback != NULL)
         {
+            // Mark the fact we're in middle of a callback so that if a call to destroy on this handle arrives at same time,
+            // we'll block destroy call until we exit.  This is basically what BeginInterfaceCallbackProcessing does,
```
massand: >This is basically what BeginInterfaceCallbackProcessing does [](start = 56, length = 60)

Nit: I think following consistent pattern of using our own BeginCallback and EndCallback throughout is more important than a tiny optimization here. A comment helps though if this optimization is really needed.

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -19,6 +19,10 @@
 
 #define DT_INTERFACE_PREFIX "$iotin:"
 
+// When we're poll active callbacks to drain while shutting down DigitalTwin Client Core,
```
massand: >we're [](start = 8, length = 5)

nit: "we"

massand:
> // Invokes Lock() (for convenience layer based handles) or else a no-op (for _LL_)

I think this comment is intended for the lock binding statement below. We got rid of interface client LL

---
Refers to: digitaltwin_client/src/dt_interface_client.c:129 in 8b30eca. [](commit_id = 8b30eca778915e8e60373e7671609aff56f97fc8, deletion_comment = False)

jspaith:
> // Invokes Lock() (for convenience layer based handles) or else a no-op (for _LL_)

It actually covers the whole function.  For _LL_ we still call dtBindingLock(), which is where no-op ness happens.

Way locking is setup in interface layer means though that thes dtBindingLockHandle could be NULL in some of the error cases which is why there's the extra check.

---
In reply to: [531373758](https://github.com/Azure/azure-iot-sdk-c/pull/1173#issuecomment-531373758) [](ancestors = 531373758)

---
Refers to: digitaltwin_client/src/dt_interface_client.c:129 in 8b30eca. [](commit_id = 8b30eca778915e8e60373e7671609aff56f97fc8, deletion_comment = False)

digitaltwin_client/src/dt_interface_client.c
jspaith: I'd have preferred to route everything through BeginInterfaceCallbackProcessing().  However, Begin...() assumes that the lock is NOT held already.  And the way our locking works, I can't call Lock() on the same lock multiple times a la EnterCriticalSection().

So... rather than having an 
InterfaceUnlock();
BeginInterfaceCallbackProcessing(). // which locks & unlocks
dtInterfaceRegisteredCallback(...);
EndInterfaceClient();

Beyond just optimization, it's safer once we go into the set state DT_INTERFACE_STATE_REGISTERED to do the processingCallback=TRUE under the same lock instead of releasing the lock (which has opening for a destroy to come in before processing when locks are released).

---
In reply to: [324348889](https://github.com/Azure/azure-iot-sdk-c/pull/1173#discussion_r324348889) [](ancestors = 324348889)

Issue #791, opened 2019-01-03
concurency issue with blob upload 


alexmrtn:
- **OS and version used:** Ubuntu 18.04

- **SDK version used:** lts_10_2018

# Description of the issue:
Hi,

From the [iothubclient_requirements](https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/devdoc/requirement_docs/iothubclient_requirements.md#overview), the high-level API should be thread-safe (unlike the low-level one). However, the blob upload interface does not respect this. 

When the `IoTHubClientCore_UploadToBlobAsync` function is called, a dedicated thread is spawned. the low-level function `IoTHubClientCore_LL_UploadToBlob` is then called within this thread, triggering the process of blob upload. 

If the function `IoTHubClientCore_UploadToBlobAsync` is called multiple times, nothing prevent threads to be launched. On Linux platform using curl adapter, `HTTPAPI_Init` function will be called multiple times. This may result in a concurrent access of the `nUsersOfHTTPAPI` variable, thus interfering with the calls to `curl_global_init` and `curl_global_cleanup`.

How can we prevent this? Is allowing only one blob upload at a time the correct behavior?

Regards,
Alexis.

ewertons:
Hi @alexmrtn ,
apologies for the long delay.

We will need to investigate the implications of multiple simultaneous calls to IoTHubClientCore_UploadToBlobAsync.

To mitigate your scenario though, please restrict to calling it only once at a time, until we investigate this further.

rafilho:
@mamokarz: please investigate/confirm and if necessary, let's proceed with a fix for this one. Thanks.

mamokarz:
Hi @alexmrtn,

I've just push the fix for this issue, I hope it solve your problem, if not, feel free to reopen it.

Thanks,
Marcos

az-iot-builder-01:
@ewertons, @mamokarz, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1174, opened 2019-09-12
Force single initialization of HTTPAPIEX for upload to blob. (#791)

mamokarz:


iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
```diff
@@ -340,6 +358,7 @@ void e2e_uploadtoblob_multiblock_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol,
 
 void e2e_uploadtoblob_test_multiple_simultaneous_uploads(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol)
 {
+    static int fileCounter = 0;
```
danewalton: >fileCounter [](start = 15, length = 11)

this doesn't seem to be doing anything. am i missing something here?

mamokarz: It is static, so it will get 0 only on the first call.

---
In reply to: [324815469](https://github.com/Azure/azure-iot-sdk-c/pull/1174#discussion_r324815469) [](ancestors = 324815469)

PR #1183, opened 2019-09-18
Adding srand() calls

### LINKED ISSUE #749: Add srand() call per-thread
Many of our samples have legitimate need to invoke `srand()` for seeding because they're generating random data, like temps, for test.

However `iothub_client_sample_module_filter` doesn't need to do this and has caused some confusion RE whether it's required for callers to do this on behalf of the C SDK (they do not).  The module_filter was just a copy/paste issue, which we should cleanup.

Should review other samples while at it to see if other issues like this.

jspaith:
Fixes #749.

Whether `srand()` sets a seed for an entire program or thread local is per-platform specific.  E.G. on Windows the random is thread-local [reference](https://docs.microsoft.com/en-us/previous-versions/f0d4wb4t(v=vs.140)?redirectedfrom=MSDN) whereas POSIX I believe is global [discussion here](https://stackoverflow.com/questions/10159728/does-one-need-to-call-srand-c-function-per-thread-or-per-process-to-seed-the-r).

Since we don't know at the Platform layer which OS we'll be running on - this is abstracted - make `srand` calls on worker threads that SDK creates.  Particularly for scenarios where we're doing UUID generation where OS doesn't provide a facility (non Windows/Linux), we want a random seed regardless of thread.


danewalton:
> static int uploadMethodInvoke_thread(void* data)

is srand needed here?

---
Refers to: iothub_client/src/iothub_client_core.c:2587 in cda979d. [](commit_id = cda979df928714dfb2b76c90bb74da31cccd063e, deletion_comment = False)

PR #1217, opened 2019-09-27
Mqtt twin timeout cleanup

### LINKED ISSUE #1210: Reported twin info can enter no-free state on MQTT
Acknowledgements to @markrad who did great investigation here and whose issue I'm basically copy/pasting below.

I've put a preliminary fix up https://github.com/Azure/azure-iot-sdk-c/tree/mqtt-twin-timeout-cleanup.  Cleaned up some of code so it's more common and of course removing reported properties.  UT's are passing but we clearly need more to cover this case.  I'll create PR after I sleep on this some and get UT.

--
Each time a device twin packet is sent to the IoT hub, the SDK will queue an item to the list ack_waiting_queue in iothubtransport_mqtt_common.c. This item will remain on the queue until the IoT hub sends a acknowledgment for the twin update.

At issue though is if there's a network disconnect (and I believe specifically that MQTT layer hasn't been alerted by underlying network stack) these packets have never been sent. Since these ack items do not have any kind of time out mechanism they will continue to collect on the ack queue and will never been removed.

There is a function in iothubtransport_mqtt_common.c called removeExpiredGetTwinRequestPendingAck that will clear the ack_waiting_queue of expired items but it specifically only processes items that are flagged as RETRIEVE_PROPERTIES. The end result is that the ack_waiting_queue will continue to grow every time there is a failure of the connection to the IoT hub.

jspaith:
Fixes #1210 which has additional details.

If a reported property ended up timing out, we got into states where we weren't cleaning it up in a timely manner.  It would be destroyed on closing the client handle, but this could be well off into the future.

After this fix make it so in our timer queue, remove message after too much time has elapsed.

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -1053,27 +1053,20 @@ static void removeExpiredTwinRequestsFromList(PMQTTTRANSPORT_HANDLE_DATA transpo
             (((current_ms - msg_entry->msgEnqueueTime) / 1000) >= ON_DEMAND_GET_TWIN_REQUEST_TIMEOUT_SECS))
         {
             item_timed_out = true;
+            if (msg_entry->userCallback != NULL)
+            {
+                msg_entry->userCallback(DEVICE_TWIN_UPDATE_COMPLETE, NULL, 0, msg_entry->userContext);
+            }
         }
         else if ((msg_entry->device_twin_msg_type == REPORTED_STATE) &&
                  (((current_ms - msg_entry->msgEnqueueTime) / 1000) >= TWIN_REPORT_UPDATE_TIMEOUT_SECS))
         {
             item_timed_out = true;
+            transport_data->transport_callbacks.twin_rpt_state_complete_cb(msg_entry->iothub_msg_id, STATUS_CODE_TIMEOUT_VALUE, transport_data->transport_ctx);
         }
 
         if (item_timed_out)
         {
-            if (msg_entry->device_twin_msg_type == RETRIEVE_PROPERTIES)
-            {
-                if (msg_entry->userCallback != NULL)
-                {
-                    msg_entry->userCallback(DEVICE_TWIN_UPDATE_COMPLETE, NULL, 0, msg_entry->userContext);
-                }
-            }
-            else if (msg_entry->device_twin_msg_type == REPORTED_STATE)
-            {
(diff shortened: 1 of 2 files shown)


iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1036,59 +1037,60 @@ static void sendPendingGetTwinRequests(PMQTTTRANSPORT_HANDLE_DATA transportData)
...
     }
+
 }
 
-static void removeExpiredGetTwinRequestsPendingAck(PMQTTTRANSPORT_HANDLE_DATA transport_data)
+static void removeExpiredTwinRequests(PMQTTTRANSPORT_HANDLE_DATA transport_data)
 {
     tickcounter_ms_t current_ms;
 
     if (tickcounter_get_current_ms(transport_data->msgTickCounter, &current_ms) == 0)
```
ewertons: This is smart. I like that and have been trying to use that way (get the current time once, and use that on the sub-sequent functions).

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -6200,6 +6202,117 @@ TEST_FUNCTION(IoTHubTransportMqtt_MessageRecv_device_twin_fail)
     umock_c_negative_tests_deinit();
 }
 
+// Subscribe for a twin notification, which implicitly initiates a GetTwin, and have the request timeout.
+TEST_FUNCTION(IoTHubTransportMqtt_implicit_gettwin_timeout)
```
ewertons: Pretty good. Thanks for adding all these tests.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1036,59 +1037,60 @@ static void sendPendingGetTwinRequests(PMQTTTRANSPORT_HANDLE_DATA transportData)
...
+            item_timed_out = true;
+        }
+        else if ((msg_entry->device_twin_msg_type == REPORTED_STATE) &&
+                 (((current_ms - msg_entry->msgEnqueueTime) / 1000) >= TWIN_REPORT_UPDATE_TIMEOUT_SECS))
+        {
+            item_timed_out = true;
+        }
 
-            if (((current_ms - msg_entry->msgEnqueueTime) / 1000) >= ON_DEMAND_GET_TWIN_REQUEST_TIMEOUT_SECS)
+        if (item_timed_out)
```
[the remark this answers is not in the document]
jspaith: Good catch on the double if() checking of state.  I'm using the optimization suggested where I'll call the function where I detect timeout, but have the freeing of the list be done in the `is_timed_out` common if check.

PR #1244, opened 2019-10-09
Fix segfault in the function message_queue_move_all_back_to_pending a…

mamokarz:


iothub_client/src/message_queue.c
```diff
@@ -386,32 +386,27 @@ static int move_messages_between_lists(SINGLYLINKEDLIST_HANDLE from_list, SINGLY
 
     while ((list_item = singlylinkedlist_get_head_item(from_list)) != NULL)
     {
+        MESSAGE_QUEUE_ITEM* mq_item = (MESSAGE_QUEUE_ITEM*)singlylinkedlist_item_get_value(list_item);
```
danewalton: >MESSAGE_QUEUE_ITEM* mq_item = (MESSAGE_QUEUE_ITEM*)singlylinkedlist_item_get_value(list_item); [](start = 8, length = 94)

if this is successful but the `singlylinkedlist_remove` fails, this seems like it is doing unnecessary work or worst case advancing this linked list forward one. i'm not sure how this works but might it be better to put this in an `else` after `singlylinkedlist_remove` and then checking if `mq_item is NULL before adding it.

mamokarz: That was exactly the cause of the segfault. The original code tries to get the content of the list_item after remove it. :)

---
In reply to: [333103165](https://github.com/Azure/azure-iot-sdk-c/pull/1244#discussion_r333103165) [](ancestors = 333103165)

danewalton: i guess if it is just getting the value it won't advance anything but still seems like unnecessary work.

---
In reply to: [333103165](https://github.com/Azure/azure-iot-sdk-c/pull/1244#discussion_r333103165) [](ancestors = 333103165)

danewalton: oh interesting okay sounds good then

---
In reply to: [333107769](https://github.com/Azure/azure-iot-sdk-c/pull/1244#discussion_r333107769) [](ancestors = 333107769,333103165)

iothub_client/src/message_queue.c
```diff
@@ -422,53 +417,35 @@ int message_queue_move_all_back_to_pending(MESSAGE_QUEUE_HANDLE message_queue)
...
     else
     {
-        SINGLYLINKEDLIST_HANDLE temp_list;
-
-        if ((temp_list = singlylinkedlist_create()) == NULL)
+        // Codes_SRS_MESSAGE_QUEUE_21_070: [The message_queue_move_all_back_to_pending shall add all in_progress message in front of the pending messages.]
+        if (move_messages_between_lists(message_queue->pending, message_queue->in_progress) != 0)
         {
-            LogError("failed creating temporary list");
+            LogError("failed moving in-progress message to temporary list");
```
danewalton: >temporary list [](start = 59, length = 14)

this no longer seems to be true

iothub_client/src/message_queue.c
```diff
@@ -422,53 +417,35 @@ int message_queue_move_all_back_to_pending(MESSAGE_QUEUE_HANDLE message_queue)
...
+            result = MU_FAILURE;
+        }
+        else if (move_messages_between_lists(message_queue->in_progress, message_queue->pending) != 0)
+        {
+            LogError("failed moving pending message to temporary list");
```
danewalton: >moving pending message to temporary list [](start = 29, length = 40)

might need to update this one as well

iothub_client/tests/message_queue_ut/message_queue_ut.c
```diff
@@ -1603,4 +1603,191 @@ TEST_FUNCTION(do_work_in_progress_queue_timeout)
     message_queue_destroy(mq);
 }
 
+// Tests_SRS_MESSAGE_QUEUE_21_070: [The message_queue_move_all_back_to_pending shall add all in_progress message in front of the pending messages.]
```
danewalton: >add all in_progress message in front of the pending messages [](start = 85, length = 60)

might be overkill but do we have a test which makes sure this is true (the ordering of the messages)

mamokarz: Short answer is no.

There is no good way to test it. We should be able to inspect the queue orders, and there is no API for that.

---
In reply to: [333112087](https://github.com/Azure/azure-iot-sdk-c/pull/1244#discussion_r333112087) [](ancestors = 333112087)

Issue #1209, opened 2019-09-25
Remove duplicate STRING_construct_sprintf call in dt_interface_client.c


jrdouceur:
dt_interface_client.c contains the following lines:

    if ((desiredPathForInterface = STRING_construct_sprintf("%s%s", DT_INTERFACE_PREFIX, dtInterfaceClient->componentName)) == NULL)
    {   
        LogError("Cannot build desiredPathForInterface string");
    }
    else if ((reportedPathForInterface = STRING_construct_sprintf("%s%s", DT_INTERFACE_PREFIX, dtInterfaceClient->componentName)) == NULL)
    {
        LogError("Cannot build reportedPathForInterface string");
    }

The call that constructs desiredPathForInterface is identical to the call that constructs reportedPathForInterface.  This should really just be a single call and a single variable, perhaps named pathForInterface.

No functional impact, but an unnecessary duplicate call, including memory allocation.

az-iot-builder-01:
@jrdouceur, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1252, opened 2019-10-10
Remove duplicate STRING_construct_sprintf call #1209

jasmineymlo:


digitaltwin_client/src/dt_interface_client.c
```diff
@@ -1200,27 +1200,22 @@ static void ProcessPropertiesForTwin(DT_INTERFACE_CLIENT* dtInterfaceClient, JSO
     JSON_Object* dtInterfaceDesiredJson;
     JSON_Object* dtInterfaceReportedJson;
 
-    STRING_HANDLE desiredPathForInterface = NULL;
-    STRING_HANDLE reportedPathForInterface = NULL;
+    STRING_HANDLE pathForInterface = NULL;
 
     // If there are property updates for this interface - which there may or may not be - the will in the passed in JSON
```
jspaith: >the [](start = 90, length = 3)

Old typo but please fix while you're in here

"the will in the passed" should be
"they will be in the passed"

PR #1259, opened 2019-10-14
Fix the cancellation of the longhaul jobs in the gate for Windows platforms

ewertons:
The longhaul tests on Windows have been cancelled by devops after 12h, because not all Windows LH tests run in parallel.
There are 10 tests, but on our scripts we restrict the max number of parallel tests to 8.
We could fix this in different ways, but this is the cheapest and most straightforward.
Also, on Linux we run up to 16 tests in parallel, so increasing by that much on Windows should be acceptable.

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/build_all/windows/build.cmd
+++ b/build_all/windows/build.cmd
@@ -220,7 +220,7 @@ if %MAKE_NUGET_PKG% == yes (
         if !ERRORLEVEL! neq 0 exit /b !ERRORLEVEL!
 
         if %build-platform% neq arm (
-            ctest -T test --no-compress-output -C "debug" -V -j 8
+            ctest -T test --no-compress-output -C "debug" -V -j 16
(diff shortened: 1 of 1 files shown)


build_all/windows/build.cmd
```diff
@@ -220,7 +220,7 @@ if %MAKE_NUGET_PKG% == yes (
         if !ERRORLEVEL! neq 0 exit /b !ERRORLEVEL!
 
         if %build-platform% neq arm (
-            ctest -T test --no-compress-output -C "debug" -V -j 8
+            ctest -T test --no-compress-output -C "debug" -V -j 10
```
jspaith: >10 [](start = 64, length = 2)

Nit: Recommend we bump this to 16 so that we have more headroom, and won't hurt in meantime.

PR #1262, opened 2019-10-15
Uhttp blob fix, stress sample enhancements [uhttp_upload]

CIPop:
1. Fixing uhttp_upload bug where requests were pipelined but not awaited causing the last step, hub notification, to fail as the blob was not yet ready.
1. Fixing proxy bug for step 3.
1. Adding CMake options to turn on logging and the stress test.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,18 +74,16 @@ option(use_prov_client "Enable provisioning client" OFF)
 option(use_tpm_simulator "tpm simulator type of hsm used with the provisioning client" OFF)
 option(use_edge_modules "Enable support for running modules against Azure IoT Edge" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
-option(use_uhttp_upload "set use_uhttp_upload to ON to use uhttp for upload to blob" OFF)
+option(use_uhttp_upload "set use_uhttp_upload to ON to use uhttp for upload to blob" ON)
 option(use_uhttp_upload_logging "set use_uhttp_upload_logging to ON to enable uhttp blob logging (only when use_uhttp_upload is also ON)" OFF)
+option(use_upload_to_blob_stress "set use_upload_to_blob_stress to run a blob stress test." OFF)
 set(compileOption_C "" CACHE STRING "passes a string to the command line of the C compiler")
 set(compileOption_CXX "" CACHE STRING "passes a string to the command line of the C++ compiler")
 set(linkerOption "" CACHE STRING "passes a string to the shared and exe linker options of the C compiler")
 
 set(use_prov_client_core OFF)
 
-# Temporary for now
-set(${use_uhttp_upload} ON)
-
-if(${use_uhttp_upload_logging} ON)
+if(${use_uhttp_upload_logging})
     add_definitions(-DUSE_UHTTP_UPLOAD_LOGGING)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,7 @@ option(use_tpm_simulator "tpm simulator type of hsm used with the provisioning c
 option(use_edge_modules "Enable support for running modules against Azure IoT Edge" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
 option(use_uhttp_upload "set use_uhttp_upload to ON to use uhttp for upload to blob" OFF)
+option(use_uhttp_upload_logging "set use_uhttp_upload_logging to ON to enable uhttp blob logging (only when use_uhttp_upload is also ON)" OFF)
 set(compileOption_C "" CACHE STRING "passes a string to the command line of the C compiler")
 set(compileOption_CXX "" CACHE STRING "passes a string to the command line of the C++ compiler")
 set(linkerOption "" CACHE STRING "passes a string to the shared and exe linker options of the C compiler")
@@ -83,6 +84,10 @@ set(use_prov_client_core OFF)
 
 # Temporary for now
 set(${use_uhttp_upload} ON)
+
+if(${use_uhttp_upload_logging} ON)
+    add_definitions(-DUSE_UHTTP_UPLOAD_LOGGING)
+endif()
 
 if(${use_custom_heap})
     add_definitions(-DGB_USE_CUSTOM_HEAP)
--- a/iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
+++ b/iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
@@ -29,7 +29,7 @@ and removing calls to _DoWork will yield the same results. */
 #define ITERATIONS    1
 #define MAX_BLOCK_SIZE_BYTES    4194304
 
-/*Optional string with http proxy host and integer for http proxy port. (Both Win and Lin when uhttp is used.)*/
+/*Optional string with http proxy host and integer for http proxy port.*/
 static const char* proxyHost = NULL;
 static int proxyPort = 0;
 
--- a/iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
+++ b/iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
@@ -29,10 +29,13 @@ and removing calls to _DoWork will yield the same results. */
 #define ITERATIONS    1
 #define MAX_BLOCK_SIZE_BYTES    4194304
 
-/*Optional string with http proxy host and integer for http proxy port (Linux only)         */
-static const char* proxyHost = "127.0.0.1";
-static int proxyPort = 81;
-static int enableVerboseLogging = 1;
+/*Optional string with http proxy host and integer for http proxy port. (Both Win and Lin when uhttp is used.)*/
+static const char* proxyHost = NULL;
+static int proxyPort = 0;
+
+// Enables verbose logging for IoT Hub communications. 
+// Use CMake flag -Duse_uhttp_upload_logging=ON to enable BLOB verbose logging.
+static int enableVerboseLogging = 0;
 
 static char data_to_upload[MAX_BLOCK_SIZE_BYTES] = { 0 };
 static int blocks_sent = 0;
@@ -55,6 +58,7 @@ static IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_RESULT getDataCallback(IOTHUB_CLIENT_F
                 *data = (const unsigned char*)data_to_upload;
                 *size = block_size;
                 blocks_sent++;
+                (void)printf(".");
             }
             else
             {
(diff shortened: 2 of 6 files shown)


CIPop:
Also requires 
https://github.com/Azure/azure-uhttp-c/pull/47
https://github.com/Azure/azure-c-shared-utility/pull/397

CMakeLists.txt
```diff
@@ -74,15 +74,18 @@ option(use_prov_client "Enable provisioning client" OFF)
 option(use_tpm_simulator "tpm simulator type of hsm used with the provisioning client" OFF)
 option(use_edge_modules "Enable support for running modules against Azure IoT Edge" OFF)
 option(use_custom_heap "use externally defined heap functions instead of the malloc family" OFF)
-option(use_uhttp_upload "set use_uhttp_upload to ON to use uhttp for upload to blob" OFF)
+option(use_uhttp_upload "set use_uhttp_upload to ON to use uhttp for upload to blob" ON)
```
ewertons: Is this really going to be the default going forward?

CMakeLists.txt
```diff
@@ -74,15 +74,18 @@ option(use_prov_client "Enable provisioning client" OFF)
...
+option(use_uhttp_upload_logging "set use_uhttp_upload_logging to ON to enable uhttp blob logging (only when use_uhttp_upload is also ON)" OFF)
```
ewertons: This seems a little off. There shouldn't be an cmake option for uhttp in the azure-iot-SDK-c

CMakeLists.txt
```diff
@@ -74,15 +74,18 @@ option(use_prov_client "Enable provisioning client" OFF)
...
+option(use_upload_to_blob_stress "set use_upload_to_blob_stress to run a blob stress test." OFF)
 set(compileOption_C "" CACHE STRING "passes a string to the command line of the C compiler")
 set(compileOption_CXX "" CACHE STRING "passes a string to the command line of the C++ compiler")
 set(linkerOption "" CACHE STRING "passes a string to the shared and exe linker options of the C compiler")
 
 set(use_prov_client_core OFF)
 
-# Temporary for now
-set(${use_uhttp_upload} ON)
+if(${use_uhttp_upload_logging})
```
ewertons: This should be in uhttp repo, AFAIK

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
```diff
@@ -29,10 +29,14 @@ and removing calls to _DoWork will yield the same results. */
 #define ITERATIONS    1
 #define MAX_BLOCK_SIZE_BYTES    4194304
 
-/*Optional string with http proxy host and integer for http proxy port (Linux only)         */
+/*Optional string with http proxy host and integer for http proxy port. (Both Win and Lin when uhttp is used.)*/
```
ewertons: This is not very clear I would say.
Either it needs to be more explanatory, or don't mention the "Both Win ..."

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
```diff
@@ -63,20 +68,20 @@ static IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_RESULT getDataCallback(IOTHUB_CLIENT_F
...
+                (void)printf("\nIndicating upload is complete (%d blocks uploaded)\r\n", blocks_sent);
             }
         }
         else
         {
             // The last call to this callback is to indicate the result of uploading the previous data block provided.
             // Note: In this last call, data and size pointers are NULL.
 
-            (void)printf("Last call to getDataCallback (result for %dth block uploaded: %s)\r\n", blocks_sent, MU_ENUM_TO_STRING(IOTHUB_CLIENT_FILE_UPLOAD_RESULT, result));
+            (void)printf("\nLast call to getDataCallback (result for %dth block uploaded: %s)\r\n", blocks_sent, MU_ENUM_TO_STRING(IOTHUB_CLIENT_FILE_UPLOAD_RESULT, result));
```
ewertons: What is the issue with all the \n? This is diverging from the usual style on the samples.

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
```diff
@@ -54,6 +58,7 @@ static IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_RESULT getDataCallback(IOTHUB_CLIENT_F
                 *data = (const unsigned char*)data_to_upload;
                 *size = block_size;
                 blocks_sent++;
+                (void)printf("."); fflush(stdout);
```
ewertons: Too much decoration. The sample is supposed to be as lean as possible, unless the directive has changed.

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
```diff
@@ -54,6 +58,7 @@ static IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_RESULT getDataCallback(IOTHUB_CLIENT_F
...
             }
```
ewertons: Oh wow. We have a stress test in the samples directory  now? hmm.

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
```diff
@@ -254,7 +254,7 @@ static HTTP_CLIENT_HANDLE create_http_client(IOTHUB_CLIENT_LL_UPLOADTOBLOB_HANDL
         uhttp_client_destroy(result);
         result = NULL;
     }
-    else if (upload_data->curl_verbosity_level != UPOADTOBLOB_CURL_VERBOSITY_UNSET && uhttp_client_set_trace(result, true, true) != HTTP_CLIENT_OK)
+    else if (upload_data->curl_verbosity_level == UPOADTOBLOB_CURL_VERBOSITY_ON && uhttp_client_set_trace(result, true, true) != HTTP_CLIENT_OK)
```
ewertons: Will this break our customers applications?

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
```diff
@@ -590,7 +584,13 @@ static int IoTHubClient_LL_UploadToBlob_step3(IOTHUB_CLIENT_LL_UPLOADTOBLOB_HAND
             case IOTHUB_CREDENTIAL_TYPE_X509_ECC:
             case IOTHUB_CREDENTIAL_TYPE_DEVICE_AUTH:
             {
-                if (send_http_request(upload_data, http_client, STRING_c_str(relativePathNotification), request_header, messageBody) != 0)
+                // Codes_SRS_IOTHUBCLIENT_LL_02_065: [ If creating the HTTPAPIEX_HANDLE fails then IoTHubClient_LL_UploadMultipleBlocksToBlob(Ex) shall fail and return IOTHUB_CLIENT_ERROR. ]
+                if ((http_client = create_http_client(upload_data)) == NULL)
```
ewertons: Isn't the client being created in another step already? Why does it need to be created here?

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
```diff
@@ -740,7 +743,13 @@ static int initiate_blob_upload(IOTHUB_CLIENT_LL_UPLOADTOBLOB_HANDLE_DATA* uploa
 
                         if ((hub_content = STRING_construct(FILE_UPLOAD_ABORTED_BODY)) != NULL)
                         {
-                            if (IoTHubClient_LL_UploadToBlob_step3(upload_data, STRING_c_str(correlation_id), http_client, request_header, hub_content) != 0)
+                            if ((http_client = create_http_client(upload_data)) == NULL)
```
ewertons: I'd like to get some idea f2f on what was the reasoning for adding create_http_client here and other places.

CMakeLists.txt
CIPop: Yes, for this branch.

CMakeLists.txt
CIPop: None of these are for uhttp - both are for the SDKs:
- first one is for using uhttp for blob upload (IoT SDK + BLOB storage SDK).
- second one is to control logging for the BLOB SDK.

CMakeLists.txt
CIPop: See above: this is for the SDK _using_ uhttp not the uhttp client.

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
CIPop: This is a stress test / sample that will only be available in this branch for a particular customer.
The \n is to add a new line after the progress done using "." characters.

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
CIPop: There are no existing apps - this implementation is for uhttp only and exists only in this branch.

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
CIPop: This branch was required by a customer that had problems with stress. We need to seriously think about a stress-test solution.
For now, this is a compromise for the uhttp_blob branch.

iothub_client/samples/iothub_client_sample_upload_to_blob_mb/iothub_client_sample_upload_to_blob_mb_stress.c
CIPop: I don't see how the sample clarity is affected. Actually, not having progress is a huge problem here as uploading 2000 chunks can take minutes and the sample appears as hanging.

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
CIPop: Good catch! No, it shouldn't need to be created again here.

iothub_client/src/iothub_client_ll_uploadtoblob_uhttp.c
CIPop: Sure: this is due to how uhttp handles xio when a proxy is configured. 
@jebrando may have more context as he was the original designer of the BLOB over uhttp implementation.

CIPop:
> Cris, the repo can only be updated to submodule commits that are in the master tree of the respective submodule.

git allowed it :laughing: !

If you mean that I _should_ update these again after addressing the PR comments in the respective branches, I completely agree. 
This though is a live branch shared with the 3rd party customer and so it must work with minimal friction for their testing.

CIPop:
Thanks @ewertons . Per our offline discussion merging these changes and will come back with another round of fixes after the perf investigation.

PR #1274, opened 2019-10-18
added a define that can be used as an option to reduce memory footpri…

jbobotek:
…nt on embedded devices Note: esp8266 may run out of memory without this define.

 

# Checklist

# Description of the problem

# Description of the solution


certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
+//Note: for the ESP8266, other certs are defined out to save on space.
```
danewalton: >the ESP8266 [](start = 12, length = 11)

this should likely be more generic like "devices with limited resources" as an indication and potential option to set for other boards/developers.

certs/certs.c
```diff
@@ -77,4 +80,6 @@ const char certificates[] =
 "PIWmawomDeCTmGCufsYkl4phX5GOZpIJhzbNi5stPvZR1FDUWSi9g/LMKHtThm3Y\r\n"
 "Johw1+qRzT65ysCQblrGXnRl11z+o+I=\r\n"
 "-----END CERTIFICATE-----\r\n"
+#endif
```
danewalton: usually good practice to end the `#endif` with a comment describing what it's closing. ex:

`#endif  // LEAN_CERTS

certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
+//Note: for devices with limited resources, other certs are defined out to save on space.
```
danewalton: >on [](start = 80, length = 2)

nit: delete

certs/certs.c
```diff
@@ -51,7 +54,7 @@ const char certificates[] =
 "YSEY1QSteDwsOoBrp+uvFRTp2InBuThs4pFsiv9kuXclVzDAGySj4dzp30d8tbQk\r\n"
 "CAUw7C29C79Fv1C5qfPrmAESrciIxpg0X40KPMbp1ZWVbd4=\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*D-TRUST Root Class 3 CA 2 2009*/
+/*D-TRUST Root Class 3 CA 2 2009 --Europe--*/
```
danewalton: >Europe [](start = 35, length = 6)

i don't think this is exactly true. i think this is particular to Germany and not necessarily all of Europe. we might need to double check though. [Link](https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud)

certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
+//Note: for devices with limited resources, other certs are defined out to save on space. 
+//For Europe and China regions, please change which certs are defined out to the appropriate ones.
```
danewalton: >Europe [](start = 6, length = 6)

same here about below comment

jbobotek: Fixed.

certs/certs.c
jbobotek: Fixed. And yeah, it does appear to be Germany specific.

certs/certs.c
jbobotek: Done.

certs/certs.c
```diff
@@ -6,7 +6,7 @@
 #include "certs.h"
 
 const char certificates[] =
-/* DigiCert Baltimore Root */
+/* DigiCert Baltimore Root --America--*/
```
danewalton: >America [](start = 29, length = 7)

i don't think this is necessarily true either. this one i believe is essentially used if `!China && !Germany`. More specifically hubs with a tail endpoint of `.azure-devices.net`

certs/certs.c
jbobotek: Done.

certs/certs.c
jbobotek: Done.

certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
+//Note: for devices with limited resources, other certs are defined out to save space. 
+//For Germany, China, and other regions, please change which certs are defined out to the appropriate ones.
```
danewalton: > and other regions [](start = 21, length = 18)

i don't believe there are other regions to consider

jbobotek: Fixed.

certs/certs.c
jbobotek: Yeah, a closer read of that article you linked me seemed to indicate that those other certs are mostly for servers geolocal to those regions that are seperate for compliance etc reasons. So !China and !Germany seems fair.

jbobotek: I put Global though to be more succinct.

certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
+//Note: for devices with limited resources, other certs are defined out to save space.
```
danewalton: >other certs are defined out to save space [](start = 44, length = 41)

this can be "developers can choose to remove certs not relevant to their region. Note that removing these might cause issues if the device were to be moved into/from Germany or China". Since they aren't automatically defined out.

certs/certs.c
```diff
@@ -28,7 +28,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#ifndef LEAN_CERTS
```
danewalton: This should likely be documented somewhere if this gets pushed. Otherwise people will have to stumble upon this to use it.

danewalton: This should also be tested at the gate

---
In reply to: [337152560](https://github.com/Azure/azure-iot-sdk-c/pull/1274#discussion_r337152560) [](ancestors = 337152560)

ewertons:
Also it would be good to squash the commits before sending the PR. There are seven commits to change just one file?

jbobotek:
> Also it would be good to squash the commits before sending the PR. There are seven commits to change just one file?

Dane was specific about language :P

certs/certs.c
rafilho: I'd suggest being even more specific: something like Azure Germany Cloud (codiname Black Forest).

The same for Azure China Cloud (codiname Mooncake).

---
In reply to: [336702649](https://github.com/Azure/azure-iot-sdk-c/pull/1274#discussion_r336702649) [](ancestors = 336702649)

certs/certs.c
```diff
@@ -4,9 +4,12 @@
 /* This file contains certs needed to communicate with Azure (IoT) */
 
 #include "certs.h"
+//Note: for devices with limited resources, only one cert should be loaded. #defines are used to reduce memory footprint of certificates.
+//For Germany and China regions, please build with -DUSE_GERMANY_CERT_ONLY or -DUSE_CHINA_CERT_ONLY if you wish to load ONLY those certs.
 
 const char certificates[] =
-/* DigiCert Baltimore Root */
+#if !(defined(USE_CHINA_CERT_ONLY) || defined(USE_GERMANY_CERT_ONLY))
```
rafilho: >USE_CHINA_CERT_ONLY [](start = 14, length = 19)

Do we need CMake updates for those defines? Or are you planning to have them in another PR?

certs/certs.c
```diff
@@ -28,7 +31,10 @@ const char certificates[] =
 "ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS\r\n"
 "R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp\r\n"
 "-----END CERTIFICATE-----\r\n"
-/*DigiCert Global Root CA*/
+#endif /* BALTIMORE_CERT */
+
+#if !(defined(USE_BALTIMORE_CERT_ONLY) || defined(USE_GERMANY_CERT_ONLY))
+/*DigiCert Global Root CA --China--*/
```
rafilho: >--China--*/ [](start = 26, length = 11)

Can we also be more descriptive here with the comment? I would add a comment in the following line instead saying something like: "This cert should be used when connecting to Azure IoT on the Azure China Cloud (a.k.a. Mooncake)." And of course, something similar for the other certs including Baltimore. It would help customers to understand why 3 certs.

danewalton: are the code names public?

certs/certs.c
jbobotek: We do not, if you use none of the defines, the file will contain all certs.

certs/certs.c
jbobotek: Updated with comments below each cert name.

jbobotek:
> I would hold on this, circulate it through the team to make sure its done solid.
> There are no tests in this PR, so I cannot validate it's correct.
> Also the naming should be reviewed. I tend to suggest more specific names,  (e.g.) like "ENABLE_CERT_FOR_REGION_X" instead.

Updated.

rafilho:
@jspaith is added to the review. #Closed

jenkins/linux_c_option_test.sh
```diff
@@ -56,6 +56,7 @@ declare -a arr=(
     "-Drun_longhaul_tests=ON"
     "-Duse_prov_client=ON -Dhsm_custom_lib=$custom_hsm_lib"
     "-Drun_e2e_tests=ON -Drun_sfc_tests=ON -Duse_edge_modules=ON"
+    "-Drun_e2e_tests=ON -DUSE_BALTIMORE_CERT=ON"
```
rafilho: Do we need this?

rafilho:
Yes, please squash. :)

---
In reply to: [544631573](https://github.com/Azure/azure-iot-sdk-c/pull/1274#issuecomment-544631573) [](ancestors = 544631573)

jenkins/linux_c_option_test.sh
jbobotek: This was added to address Dane's concern that new #defines should be tested.

ewertons: When running cmake from the command line, it is possible to specify command line options to cmake that will set values in the cache. This is done with a -DVARIABLE:TYPE=VALUE syntax on the command line. This is useful for non-interactive nightly test builds.
​
https://cmake.org/runningcmake/
​
-D defines variables for cmake, not #defines in the project

that is done with add_definition()

so passing that actually gets ignored by cmake, since it's not used anywhere

CMakeLists.txt
```diff
@@ -28,6 +28,18 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     add_definitions(-DMACOSX)
 endif()
 
+if (USE_BALTIMORE_CERT)
```
jspaith: If we want these exposed up via cmake, I think we should have option()'s for them grouped with the list below.  ANd then these if() checks should come after the option() stuff.

Everything else in this file that has in if(X) *without* the explicit option(), the X is a system thing like the compiler name or Linux.

jbobotek: Rgr.

CMakeLists.txt
```diff
@@ -28,6 +28,18 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
...
+endif()
+
+if (USE_MICROSOFTAZURE_DE_CERT)
+    add_definitions(-DUSE_MICROSOFTAZURE_DE_CERT)
+endif()
+
+if (USE_PORTAL_AZURE_CN_CERT)
+    add_definitions(-DUSE_PORTAL_AZURE_CN_CERT)
+endif()
+
```
ewertons: It's close, but …

Look inside the CMakeLists.txt for "use_uamqp".

The option should be defined like that.

option(use_baltimore_cert "set use_baltimore_cert to ON if the Baltimore cert is to be used, set to OFF to not use it" OFF)

And on the if, it should be checked if its value is ON.

if (${use_baltimore_cert}) 
    add_definitions(-DUSE_BALTIMORE_CERT) 
endif()

------

The same above needs to be done for all 3 new variables.

-----

And down where you call cmake, use -Duse_baltimore_cert=ON (notice the lowercase).

jbobotek: Thanks for the specificity!

PR #1289, opened 2019-11-01
Move send sdk information to sample

jasmineymlo:
refactor sending sdk information related code out of the SDK and implement it as a sample component.


digitaltwin_client/samples/digitaltwin_sample_sdk_info/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+set(digitaltwin_sample_sdk_info_c_files
+    ./digitaltwin_sample_sdk_info.c
+)
+
+set(digitaltwin_sample_sdk_info_h_files
+    ./digitaltwin_sample_sdk_info.h
+)
+
+set(DIGITALTWIN_SAMPLE_SDK_INFO_INC_FOLDER ${CMAKE_CURRENT_LIST_DIR} CACHE INTERNAL "this is what needs to be included if using digitaltwin_client lib" FORCE)
```
jspaith: >"this is what needs to be included if using digitaltwin_client lib" [](start = 84, length = 67)

The "this is what..." comment I think we should remove from here and various other sample interfaces it was pre-existing.

I don't think as-is it adds much clarity and not sure if we need explicit in-line doc in 1st place.

digitaltwin_client/samples/digitaltwin_sample_sdk_info/digitaltwin_sample_sdk_info.c
```diff
@@ -0,0 +1,194 @@
...
+//
+// DigitalTwinSampleSdkInfo_ReportPropertyAsync is a helper function to report a SdkInfo's properties.
+// It invokes underlying DigitalTwin API for reporting properties and sets up its callback on completion.
+//
+static DIGITALTWIN_CLIENT_RESULT DigitalTwinSampleSdkInfo_ReportPropertyAsync(DIGITALTWIN_INTERFACE_CLIENT_HANDLE interfaceHandle, const char* propertyName, const char* propertyData)
+{
+    char* serializedPropertyValue = NULL;
+    DIGITALTWIN_CLIENT_RESULT result = DIGITALTWIN_CLIENT_ERROR;
+
+    size_t propertyDataSize = snprintf(NULL, 0, "\"%s\"", propertyData);
```
jspaith: >size_t propertyDataSize = snprintf(NULL, 0, "\"%s\"", propertyData); [](start = 4, length = 68)

I think we can remove the snprintf() and calloc().  What if instead we had DT_SdkLanguage and DT_SdkVendor just hard-code in the escaping, with a comment about the fact it had to be a legal jvalue.  For DigitalTwin_Client_GetVersionString() we could also use implicit string concat.  So something like DT_Version = "\"" DIGITALTWIN_CLIENT_SDK_VERSION "\"".

jspaith: I'm signing off momentarily, assuming you're in agreement with this comment.  If I'm missing something please let's talk when you're in prior to checkin.

---
In reply to: [341720075](https://github.com/Azure/azure-iot-sdk-c/pull/1289#discussion_r341720075) [](ancestors = 341720075)

digitaltwin_client/samples/digitaltwin_sample_sdk_info/CMakeLists.txt
jasmineymlo: Should we remove "CACHE INTERNAL "this is what needs to be included if using digitaltwin_client lib" FORCE" completely so that the variable is not cached?  The comment is required if we want to cache the variable.

jspaith: Reading the cmake docs on [set](https://cmake.org/cmake/help/v3.0/command/set.html), I don't think we need the CACHE stuff here.  See `variable types in CMake` for details.

@danewalton - you've spent more time with cmake than I have.  Can you see any reason we'd want this?

danewalton: to promote best CMake practices this should not be included. if you wish to use an installed version of a lib there are other more modern ways to do it. [Example](https://github.com/Azure/azure-macro-utils-c/blob/5926caf4e42e98e730e6d03395788205649a3ada/CMakeLists.txt#L32) that I pushed to `macro-utils` which differentiates between the install directory and build directory.

PR #1302, opened 2019-11-07
Add comments about network status callbacks

jspaith:
Clarify how applications should take network status callbacks.

### DIFF
--- a/digitaltwin_client/doc/connection_setup.md
+++ b/digitaltwin_client/doc/connection_setup.md
@@ -32,7 +32,7 @@ The first step in using the Digital Twin SDK is to create the appropriate `IOTHU
 
 **NOTE: The only protocols supported for Digital Twin for the C SDK are MQTT and MQTT over Web Sockets.**  Attempts to use AMQP, AMQP over Web Sockets, or HTTP for C will fail.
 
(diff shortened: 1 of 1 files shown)


digitaltwin_client/doc/connection_setup.md
```diff
@@ -32,7 +32,7 @@ The first step in using the Digital Twin SDK is to create the appropriate `IOTHU
 
 **NOTE: The only protocols supported for Digital Twin for the C SDK are MQTT and MQTT over Web Sockets.**  Attempts to use AMQP, AMQP over Web Sockets, or HTTP for C will fail.
 
-Once the handle is created, you may need to [set options on it](../../doc/Iothub_sdk_options.md).  Some of the options you may need to set include enabling web proxies, enabling verbose logging at the transport layer, configuring client certificates, and more.  All options that need to be set should be *before* the next step.
+Once the handle is created, you may need to [set options on it](../../doc/Iothub_sdk_options.md).  Some of the options you may need to set include enabling web proxies, enabling verbose logging at the transport layer, configuring client certificates, and more.  If you want to receive notifications of the status of the connection to IoTHub, this is also the time to do so via the `IoTHubDeviceClient_SetConnectionStatusCallback` (or analogous function).  All options that need to be set should be *before* the next step.
```
jasmineymlo: Do we need to mention about setting retry policy/interval?

jspaith: I believe it's implicit in the "Some of the options..." section, since we don't want to re-doc everything.  I'm putting the SetConnectionStatusCallback explicitly called out here though since it's different than how a SetOption() would behave.

PR #1307, opened 2019-11-12
Option build fix

jasmineymlo:
- fix command in .sh to build with options array.
- fix "-Drun_e2e_tests=ON -Drun_sfc_tests=ON -Duse_edge_modules=ON" build.
- fix build with options for various hsm types.


jenkins/linux_c_option_test.sh
```diff
@@ -59,7 +59,7 @@ declare -a arr=(
     "-Drun_e2e_tests=ON -Duse_baltimore_cert=ON"
     "-Duse_prov_client:BOOL=ON -Dhsm_type_symm_key:BOOL=ON"
     "-Duse_prov_client:BOOL=ON -Dhsm_type_x509:BOOL=ON"
-    "-Duse_prov_client:BOOL=ON -hsm_type_sastoken:BOOL=ON"
+    "-Duse_prov_client:BOOL=ON -Dhsm_type_sastoken:BOOL=ON"
```
CIPop: I'm a bit worried this worked before... I'm still ramping up to CMake but if anyone has an idea on how to actually _test_ this works, please do add your comments.

CIPop: If there is a "treat warnings as errors" in CMake we should enable it. Alternatively, in the future, we can grep the output and check for warnings that shouldn't be there.
For now, we can manually verify if the expected things actually got built whenever we touch this file.

provisioning_client/tools/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ function(add_tool_directory whatIsBuilding)
 endfunction()
 
 if (NOT ${hsm_type_custom} AND NOT ${run_e2e_tests})
-    add_tool_directory(dice_device_enrollment)
-    add_tool_directory(tpm_device_provision)
-    add_tool_directory(symm_key_provision)
+	if (${hsm_type_x509})
```
CIPop: NIT: why 6 char indentation?

jenkins/linux_c_option_test.sh
jebrando: it worked because of the way CMake accepts anything that you send it.  by have "" around the below item  it took the whole item as the input which cmake will happily accepts.  Removing the "" will make each variable be used individually.

---
In reply to: [345464418](https://github.com/Azure/azure-iot-sdk-c/pull/1307#discussion_r345464418) [](ancestors = 345464418)

provisioning_client/CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ target_link_libraries(prov_auth_client hsm_security_client)
 set(provisioning_libs ${provisioning_libs} prov_auth_client)
 set(provisioning_headers ${provisioning_headers} ${AUTH_CLIENT_H_FILES})
 
-if(${use_prov_client})
+if(${use_prov_client} OR (${use_prov_client_core} AND ${run_e2e_tests}))
```
jebrando: >AND ${run_e2e_tests} [](start = 50, length = 20)

I'm pretty sure you don't want run_e2e_tests here.  You want some of this stuff done even if run_e2e_test is OFF>

CMakeLists.txt
```diff
@@ -366,7 +366,7 @@ if (${use_prov_client_core})
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_PROV_MODULE")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_PROV_MODULE")
 
-    if (${build_provisioning_service_client} AND ${use_prov_client})
+    if ((${build_provisioning_service_client} AND ${use_prov_client}) OR ${run_e2e_tests})
```
jebrando: > OR ${run_e2e_tests} [](start = 69, length = 20)

This is interesting logic here.  So if you don't have e2e test you you won't have provisioning service client.  I like the idea, but I'm not sure if others will go with it.  I just want to make sure you meant to have that.

jasmineymlo: Yes, this is for the case of building for Edge where ${use_prov_client_core} is ON and not ${use_prov_client}.  We would only bring provisioning_service_client for the case of e2e tests

provisioning_client/CMakeLists.txt
jasmineymlo: discussed offline.  It's ok for now to unblock as we do not want to bring in extras for the case of building for Edge.  The caveat is that we are bringing in more bits for e2e tests only meaning that the bits that we are testing vs in production "COULD" be different.  This is not ideal and we are going to create a backlog item to track the work for investigation/fixing it.

Issue #1270, opened 2019-10-18
PnP should provide reported properties to app always


jspaith:
Currently for C SDK, when we get a full twin we report properties to application *only* if there are desired properties in the list.  If there is a reported property but not a desired property, we just swallow it.

This isn't right.  We should always provide reported properties.  This will help bandwidth conscious interfaces to be thoughtful about whether they report back a given reported property again or whether they realize the server has what they were going to send up anyway and save the round-trip.

az-iot-builder-01:
@jspaith, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1315, opened 2019-11-13
Provide reported properties to components on receipt of full twin. issue #1270.

jasmineymlo:


digitaltwin_client/samples/digitaltwin_sample_environmental_sensor/digitaltwin_sample_environmental_sensor.c
```diff
@@ -524,6 +524,11 @@ static void DigitalTwinSampleEnvironmentalSensor_ProcessPropertyUpdate(const DIG
     {
         DigitalTwinSampleEnvironmentalSensor_BrightnessCallback(dtClientPropertyUpdate, propertyCallbackContext);
     }
+    else if (strcmp(dtClientPropertyUpdate->propertyName, digitaltwinSample_DeviceStateProperty) == 0)
+    {
+        LogInfo("ENVIRONMENTAL_SENSOR_INTERFACE: Property name <%s>, last reported value=<%.*s>", 
+            dtClientPropertyUpdate->propertyName, (int)dtClientPropertyUpdate->propertyReportedLen, dtClientPropertyUpdate->propertyReported);
```
jspaith: >     [](start = 8, length = 4)

Nit - please make the dtClientPropertyUpdate line up under the " of the LogInfo line.

jasmineymlo: This is one statement broken into 2 lines...  We are only doing LogInfo in the else if case.

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -1163,6 +1163,7 @@ static void ProcessPropertyUpdateIfNeededFromDesired(DT_INTERFACE_CLIENT* dtInte
             if (dtInterfaceReportedJson)
             {
                 payloadForInitialProperty = GetPayloadFromProperty(dtInterfaceReportedJson, propertyName);
+                json_object_remove(dtInterfaceReportedJson, propertyName);
```
jspaith: >json_object_remove [](start = 16, length = 18)

Could you please put a comment explaining the intent of this.  Namely so that it signals to caller that we have invoked the callback for this reported property and reported-only-properties should be skipped.

I don't see how else to solve this problem more cleanly than this, but we typically don't remove items like this so we should call out why.

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -1188,6 +1189,47 @@ static void ProcessPropertyUpdateIfNeededFromDesired(DT_INTERFACE_CLIENT* dtInte
     }
 }
 
+// For each high-level property associated with this interface, determine if there is a callback function associated with it.
+static void ProcessPropertyUpdateIfNeededFromReportedOnly(DT_INTERFACE_CLIENT* dtInterfaceClient, JSON_Object* dtInterfaceReportedJson, size_t propertyNumber, int jsonVersion)
+{
+    const char* propertyName;
+    if ((propertyName = json_object_get_name(dtInterfaceReportedJson, propertyNumber)) == NULL)
```
jspaith: >json_object_get_name [](start = 24, length = 20)

If we remove the reported value from a previous call to desired/reported, I believe this will cause a LogError() to be output correct?  We shouldn't do that as it'd be a false positive for customer.

jasmineymlo: No.  Refer to line 1280, after ProcessPropertyUpdateIfNeededFromDesired(), we do a json_object_get_count() to get the number of reported property in the json before going into another for loop for processing the reported-only properties.  The index is restarted.  The LogError should not be false positive in this case.

Issue #1311, opened 2019-11-13
Don't directly reference `DIGITALTWIN_CLIENT_SDK_VERSION` in SDK Information


jspaith:
PnP's sample of SDKInformation has a hardcoded reference of `DIGITALTWIN_CLIENT_SDK_VERSION` in it to specify the version.

If the dev updates the SDK and statically links the code in, then we'll be fine because `DIGITALTWIN_CLIENT_SDK_VERSION` will be updated and they'll recompile.

If they're using pnp_sdk.dll and not recompiling, directly referencing `DIGITALTWIN_CLIENT_SDK_VERSION` is a problem.

Instead we should:
* Fix SDKInformation to use `DigitalTwin_Client_GetVersionString`.
* Remove `DIGITALTWIN_CLIENT_SDK_VERSION` from the header so that other people don't make the same mistake.

az-iot-builder-01:
@jspaith, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1329, opened 2019-11-18
use DigitalTwin_Client_GetVersionString() to get the sdk version in s…

jasmineymlo:
fix issue #1311

### DIFF
--- a/digitaltwin_client/src/dt_version.c
+++ b/digitaltwin_client/src/dt_version.c
@@ -3,6 +3,8 @@
 
 #include "digitaltwin_client_version.h"
 
+#define DIGITALTWIN_CLIENT_SDK_VERSION "0.9.0"
+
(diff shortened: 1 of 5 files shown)


jspaith:
@jasmineymlo - I think you should also remove `DIGITALTWIN_CLIENT_SDK_VERSION` from the public header file itself.  If I could make this mistake it's likely customers would also.

Thanks

digitaltwin_client/src/dt_version.c
```diff
@@ -3,6 +3,9 @@
 
 #include "digitaltwin_client_version.h"
 
+/** @brief Pre-processor friendly representation of Digital Twin Client version. */
```
jspaith: Please remove this comment as it's for Doxygen / external facing docs, which doesn't apply as this is internal facing now.

PR #1336, opened 2019-11-19
define defaults value for device info sample. #1325

### LINKED ISSUE #1325: device info sample should utilize a header where values are set
The values provided for the device info interface sample are in the c file, these should instead be #defines in a header, with default values provided if not defined. I would also recommend following the following guidelines for the default values: A.B.C.D for version where each is an integer and 1-64 chars, alphanumeric, dot and dash only for other fields.

jasmineymlo:
fix #1325

### DIFF
--- a/digitaltwin_client/samples/digitaltwin_sample_device_info/ops/deviceinfo_ops_sample.c
+++ b/digitaltwin_client/samples/digitaltwin_sample_device_info/ops/deviceinfo_ops_sample.c
@@ -1,75 +1,77 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#include <stdio.h>
-#include <stdlib.h>
-#include <stdbool.h>
-#include "azure_c_shared_utility/xlogging.h"
 #include "../digitaltwin_sample_device_info.h"
 #include "deviceinfo_ops.h"
 
-// Sample values
-static const char digitaltwinSample_SoftwareVersionData[] = "software-revision-1.0.2A";
-static const char digitaltwinSample_ManufacturerData[] = "Sample Manufacturer for Azure IoT C SDK enabled device";
-static const char digitaltwinSample_ModelData[] = "Sample Model 123";
-static const char digitaltwinSample_OsNameData[] = "sample-OperatingSystem-name";
-static const char digitaltwinSample_ProcessorArchitectureData[] = "Contoso-Arch-64bit";
-static const char digitaltwinSample_ProcessorManufacturerData[] = "Processor Manufacturer(TM)";
-static const int  digitaltwinSample_TotalStorageData = 83886080; // 80GB
-static const int  digitaltwinSample_TotalMemoryData = 32768; // 32 MB
+#ifndef DIGITALTWIN_SAMPLE_SOFTWARE_VERSION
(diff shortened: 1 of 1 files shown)


digitaltwin_client/samples/digitaltwin_sample_device_info/ops/deviceinfo_ops_sample.c
```diff
@@ -1,75 +1,77 @@
...
-static const char digitaltwinSample_ProcessorArchitectureData[] = "Contoso-Arch-64bit";
-static const char digitaltwinSample_ProcessorManufacturerData[] = "Processor Manufacturer(TM)";
-static const int  digitaltwinSample_TotalStorageData = 83886080; // 80GB
-static const int  digitaltwinSample_TotalMemoryData = 32768; // 32 MB
+#ifndef DIGITALTWIN_SAMPLE_SOFTWARE_VERSION
+#define DIGITALTWIN_SAMPLE_SOFTWARE_VERSION "1.0.0.0"
+#endif
+
+#ifndef DIGITALTWIN_SAMPLE_MANUFACTURER
+#define DIGITALTWIN_SAMPLE_MANUFACTURER "Sample Manufacturer for Azure IoT C SDK enabled device"
```
gregman-msft: We need manufacturer to follow these rules: 1-64 chars, alphanumeric, dot and dash.

EG: "Sample-Manufacturer"

digitaltwin_client/samples/digitaltwin_sample_device_info/ops/deviceinfo_ops_sample.c
```diff
@@ -1,75 +1,77 @@
...
+#endif
+
+#ifndef DIGITALTWIN_SAMPLE_MODEL
+#define DIGITALTWIN_SAMPLE_MODEL "Sample Model 123"
```
gregman-msft: We also need model to follow these rules: 1-64 chars, alphanumeric, dot and dash.

EG: "Sample-Model-123"

digitaltwin_client/samples/digitaltwin_sample_device_info/ops/deviceinfo_ops_sample.c
```diff
@@ -1,75 +1,77 @@
...
+#ifndef DIGITALTWIN_SAMPLE_PROCESSOR_ARCHITECTURE
+#define DIGITALTWIN_SAMPLE_PROCESSOR_ARCHITECTURE "Contoso-Arch-64bit"
+#endif
+
+#ifndef DIGITALTWIN_SAMPLE_PROCESSOR_MANUFACTURER
+#define DIGITALTWIN_SAMPLE_PROCESSOR_MANUFACTURER "Processor Manufacturer(TM)"
+#endif
+
+#ifndef DIGITALTWIN_SAMPLE_TOTAL_STORAGE
+#define DIGITALTWIN_SAMPLE_TOTAL_STORAGE 83886080 // 80GB
```
jspaith: Nit: Could we reduce this from 80GB to something smaller, like into the 1-2MB maybe so we're looking a bit more embedded.  Our SDK doesn't take 80GB to run :).  Same feedback with the 32MB of memory in DIGITALTWIN_SAMPLE_TOTAL_MEMORY  below.

PR #1338, opened 2019-11-20
Remove DIGITALTWIN_LOGGING_ENABLED and use standard log. 

### LINKED ISSUE #1326: Remove DIGITALTWIN_LOGGING_ENABLED and move to standard log
Today DigitalTwin has a special macro for logging out info, though it falls directly to the C SDK `LogInfo`.  There's no equivalent `DigitalTwinLogError` - the code in DT uses `LogError` directly - so the lack of symmetry is one problem.

We should remove `DigitalTwinLogInfo` and replace it with `LogInfo` like everything else across C SDK.  The intent was to allow DT logs to be not included but C SDK to be, but don't think this is worth pattern mismatch.  Also need to remove references to `DIGITALTWIN_LOGGING_ENABLED` in docs / .c files a few places.

```c
#ifdef DIGITALTWIN_LOGGING_ENABLED
#define DigitalTwinLogInfo LogInfo
#else
#define DigitalTwinLogInfo(...)
#endif
```

jasmineymlo:
fix #1326

### DIFF
--- a/digitaltwin_client/CMakeLists.txt
+++ b/digitaltwin_client/CMakeLists.txt
@@ -65,8 +65,6 @@ IF(WIN32)
     add_definitions(-DGB_MEASURE_MEMORY_FOR_THIS -DGB_DEBUG_ALLOC)
 ENDIF(WIN32)
 
-add_definitions(-DDIGITALTWIN_LOGGING_ENABLED)
-
 
 add_library(digitaltwin_client
     ${digitaltwin_client_c_files}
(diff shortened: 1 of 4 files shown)


jspaith:
@dooriya - if codegen is using any DigitalTwin specific logging, early heads up that we're removing it and just going to LogInfo().  See also Jasmine's removal of the -DENABLE logic in the cmakelists.txt.

This will NOT impact your coming release as it's on a different branch.

PR #1342, opened 2019-11-22
Remove MU_DEFINE_ENUM_* from headers, use standard C enums

jspaith:
Using `MU_DEFINE_ENUM_WITHOUT_INVALID` and `MU_DEFINE_ENUM` has proven to be problematic.  We've got breaks from underlying macro utils library into public headers, it's not standard C, and doesn't play well with doxygen.

This change removes MU_DEFINE_ENUM* out of the IoT C SDK's public headers and goes to a straight enum.  (We are not touching the serializer directory since it's otherwise deprecated and this isn't a security/crash level for it.)

We're going to not break existing code - both code dynamically linking against exports that MU_DEFINE_ENUM* produces (see `IOTHUB_CLIENT_IOTHUB_METHOD_STATUSStrings` as one example of many).  And as icing on cake, we're going to do this so anyone doing an `MU_ENUM_TO_STRING` is also going to keep working.

We do this by:
* In the .c files that have `MU_DEFINE_ENUM_STRINGS_WITHOUT_INVALID` that build the export tables, leave them alone so everything still gets generated.
* Put a `XYZ_RESULT_VALUES` with the list of enum names into the .c file.  See `iothub_client_core_ll.c` and `IOTHUB_CLIENT_RESULT_VALUES` for how it'd work.  
* In the header files, explicitly declare a function named e.g. `IOTHUB_CLIENT_FILE_UPLOAD_RESULTStrings`.  This is what the MU_ENUM_TO_STRING ultimately wants to call, and thanks to leaving MU_DEFINE_ENUM_STRINGS_WITHOUT_INVALID around it's still there.  And with the string table to boot.
* In the UT, I can't do ASSERT against the MU_DEFINE_ENUM* types anymore.  The ASSERT turns into a fairly complex set of calls that use the name of the type and assume funnctions are there for like printing out values to stdout.  Rather than hand-roll these all over the place, it's fine using an int.

There's a few fit & finish places I need to work on, at minimum putting in all the XYZ_RESULT_VALUES and also making sure that the *ResultStrings() functions have proper spacing (there's a lot of `const char *` instead of `const char*`.

This PR is out as draft (a) for initial high-level feedback and to get a proper gate run.


iothub_client/inc/iothub_client_core_common.h
```diff
@@ -17,137 +17,138 @@ extern "C"
 {
 #endif
 
-#define IOTHUB_CLIENT_FILE_UPLOAD_RESULT_VALUES \
-    FILE_UPLOAD_OK, \
-    FILE_UPLOAD_ERROR
+typedef enum IOTHUB_CLIENT_FILE_UPLOAD_RESULT_TAG
+{
+    FILE_UPLOAD_OK = 0,
```
danewalton: just as a general comment...not to be fixed in this PR because it would take a while. But if and after this gets merged we should go through these enums and properly document them with doxygen. EG:
`/**<  */`

iothub_client/inc/iothub_client_core_common.h
```diff
@@ -17,137 +17,138 @@ extern "C"
...
+    FILE_UPLOAD_ERROR = 1
+} IOTHUB_CLIENT_FILE_UPLOAD_RESULT;
 
-    MU_DEFINE_ENUM_WITHOUT_INVALID(IOTHUB_CLIENT_FILE_UPLOAD_RESULT, IOTHUB_CLIENT_FILE_UPLOAD_RESULT_VALUES)
-        typedef void(*IOTHUB_CLIENT_FILE_UPLOAD_CALLBACK)(IOTHUB_CLIENT_FILE_UPLOAD_RESULT result, void* userContextCallback);
+const char* IOTHUB_CLIENT_FILE_UPLOAD_RESULTStrings(IOTHUB_CLIENT_FILE_UPLOAD_RESULT value);
```
danewalton: so these functions just become NOP's?

iothub_client/inc/iothub_client_core_common.h
```diff
@@ -17,137 +17,138 @@ extern "C"
...
-    IOTHUB_CLIENT_CONNECTION_COMMUNICATION_ERROR,          \
-    IOTHUB_CLIENT_CONNECTION_NO_PING_RESPONSE,             \
-    IOTHUB_CLIENT_CONNECTION_OK                            \
-
-    /** @brief Enumeration passed in by the IoT Hub when the connection status
-    *           callback is invoked to indicate status of the connection in
-    *           the hub.
-    */
-    MU_DEFINE_ENUM_WITHOUT_INVALID(IOTHUB_CLIENT_CONNECTION_STATUS_REASON, IOTHUB_CLIENT_CONNECTION_STATUS_REASON_VALUES);
+} IOTHUBMESSAGE_DISPOSITION_RESULT;
```
danewalton: this one doesn't seem to have a `Strings()` function. by design? do we only now put them in if the compiler complains it doesn't exist? Basically only if the `to_string()` call is made for that enum?

iothub_client/tests/common_device_method_e2e/iothubclient_common_device_method_e2e.c
```diff
@@ -216,13 +212,13 @@ void test_invoke_device_method(const char* deviceId, const char* moduleId, const
     {
         LogInfo("IoTHubDeviceMethod_InvokeModule deviceId='%s', moduleId='%s'", deviceId, moduleId);
         IOTHUB_DEVICE_METHOD_RESULT invokeResult = IoTHubDeviceMethod_InvokeModule(serviceClientDeviceMethodHandle, deviceId, moduleId, METHOD_NAME, payload, TIMEOUT, &responseStatus, &responsePayload, &responsePayloadSize);
-        ASSERT_ARE_EQUAL(IOTHUB_DEVICE_METHOD_RESULT, IOTHUB_DEVICE_METHOD_OK, invokeResult, "Service Client IoTHubDeviceMethod_InvokeModule failed");
+        ASSERT_ARE_EQUAL(int, IOTHUB_DEVICE_METHOD_OK, invokeResult, "Service Client IoTHubDeviceMethod_InvokeModule failed");
```
danewalton: Just curious. How useful are these additional error message strings? For ulib we don't have them and if a test fails it will say something like "Expected X but was Y on line 215". That seems to be sufficient to see where the problem is. These error messages only add to the strings that are required to be stored by the device and potentially further limit platforms that could run them. I could be nitpicking here.

iothub_client/tests/iothubclient_ll_u2b_ut/iothub_client_ll_u2b_ut.c
```diff
@@ -1013,7 +1007,7 @@ TEST_FUNCTION(IoTHubClient_LL_UploadToBlob_Impl_with_proxy_succeeds)
 
     //assert
     (void)result;
-    //ASSERT_ARE_EQUAL(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_OK, result);
+    //ASSERT_ARE_EQUAL(int, IOTHUB_CLIENT_OK, result);
```
danewalton: nit: should this just be removed

iothub_client/tests/iothubmessage_ut/iothubmessage_ut.c
```diff
@@ -1079,7 +1073,7 @@ TEST_FUNCTION(IoTHubMessage_GetMessageId_SUCCEED)
     // arrange
     IOTHUB_MESSAGE_HANDLE h = IoTHubMessage_CreateFromString(TEST_STRING_VALUE);
     (void)IoTHubMessage_SetMessageId(h, TEST_MESSAGE_ID);
-    //ASSERT_ARE_EQUAL(IOTHUB_MESSAGE_RESULT, IOTHUB_MESSAGE_OK, msgResult);
+    //ASSERT_ARE_EQUAL(int, IOTHUB_MESSAGE_OK, msgResult);
```
danewalton: nit: remove

iothub_service_client/src/iothub_devicetwin.c
```diff
@@ -28,8 +28,10 @@
     IOTHUB_TWIN_REQUEST_UPDATE_DESIRED
 
 MU_DEFINE_ENUM(IOTHUB_TWIN_REQUEST_MODE, IOTHUB_TWIN_REQUEST_MODE_VALUES);
+
```
danewalton: So only public facing enums are getting the treatment?

danewalton:
added a few comments but otherwise looks good.

iothub_client/inc/iothub_client_core_common.h
amih90: what are those ENUMString()?

amih90:
Hi John,
what will be the best approach if I would like to introduce a new Enum and I need to use STRING_TO_ENUM, if using MU_STRING_TO_ENUM is going to be deprecated?

I did 
```
#define NETID_RESULT_VALUES    \
    NETID_TCP,              \
    NETID_UDP,              \
    NETID_RAW,              \
    NETID_U_DGR,            \
    NETID_U_STR,            \
    NETID_P_RAW,            \
    NETID_P_DGR,            \
    NETID_NL

    MU_DEFINE_ENUM(NETID_RESULT, NETID_RESULT_VALUES)
    MU_DEFINE_ENUM_STRINGS(NETID_RESULT, NETID_RESULT_VALUES);
   ... 
    NETID_RESULT netid_result;
    MU_STRING_TO_ENUM(netid_str, NETID_RESULT, &netid_result);
```
and it works.

but if I change to standard C enum I need a method for each enum I create in order to map string to enum, or there is an helper inside the sdk to do so?

this is my enum after this change
```
typedef enum IOTSECURITY_COLLECTOR_NETWORK_NETID_TAG {
    NETID_TCP = 0,
    NETID_UDP = 1,
    NETID_RAW = 2,
    NETID_U_DGR = 3,
    NETID_U_STR = 4,
    NETID_P_RAW = 5,
    NETID_P_DGR = 6,
    NETID_NL = 7,
} IOTSECURITY_COLLECTOR_NETWORK_NETID;

```

thanks,
Ami Hollander

jspaith:
@amih90 - we definitely recommend not using MU_DEFINE_ENUM.  It doesn't play well with doxygen and it's an extra concept count for users who should just see the "typedef enum".

In general you should think about whether you need the friendly string output in 1st place.  It'll take extra ROM space.  If you need this though you'd need to create your own helper.  This PR has some xxxStrings() API's around it, but we're doing this because we don't want to break compat with anyone using the MU_ENUM_TO_STRING() as-is.  If we were starting from scratch we wouldn't have that as an option in the first place.

iothub_client/inc/iothub_client_core_common.h
jspaith: Agree.  Let's track - [https://github.com/Azure/azure-iot-sdk-c/issues/1354](https://github.com/Azure/azure-iot-sdk-c/issues/1354).

---
In reply to: [350323583](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r350323583) [](ancestors = 350323583)

iothub_client/inc/iothub_client_core_common.h
jspaith: Bug.  I basically wanted to get PR out for early feedback knowing there were gaps.  I shall address before round 2.

---
In reply to: [350329737](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r350329737) [](ancestors = 350329737)

iothub_client/inc/iothub_client_core_common.h
jspaith: This is what MU_ENUM_TO_STRINGS() ends up calling once the macro unexpanded.  By putting these in, it means that existing code will just work.

---
In reply to: [350328454](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r350328454) [](ancestors = 350328454)

iothub_client/inc/iothub_client_core_common.h
```diff
@@ -17,137 +17,138 @@ extern "C"
...
+{
+    IOTHUB_CLIENT_OK = 0,
+    IOTHUB_CLIENT_INVALID_ARG = 1,
+    IOTHUB_CLIENT_ERROR = 2,
+    IOTHUB_CLIENT_INVALID_SIZE = 3,
+    IOTHUB_CLIENT_INDEFINITE_TIME = 4
+} IOTHUB_CLIENT_RESULT;
 
-    MU_DEFINE_ENUM_WITHOUT_INVALID(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_RESULT_VALUES);
+const char* IOTHUB_CLIENT_RESULTStrings(IOTHUB_CLIENT_RESULT value);
```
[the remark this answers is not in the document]
jspaith: This is in a header and callers of MU_ENUM_TO_STRING would want to invoke this, so it shouldn't be static.

---
In reply to: [352721362](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r352721362) [](ancestors = 352721362)

iothub_client/inc/iothub_client_core_common.h
jspaith: Yes.  MU_ENUM_TO_STRING will just work - 0 changes - with these.

---
In reply to: [352722212](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r352722212) [](ancestors = 352722212)

iothub_client/tests/common_device_method_e2e/iothubclient_common_device_method_e2e.c
jspaith: I kind of like them because there's been times I've seen the message coming out of the log and I can go right to problem in code without needing to stare at test.

Now that said I can see how putting the e2e on the device itself is already getting tight.  For future projects I'm fine with us having guidance NOT to do this.

---
In reply to: [350351849](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r350351849) [](ancestors = 350351849)

iothub_service_client/src/iothub_devicetwin.c
jspaith: Yes.  I don't think the ROI is as good on internal logic.  At minimum I'd want to get the public API changes in now.

---
In reply to: [350385167](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r350385167) [](ancestors = 350385167)

iothub_client/inc/iothub_client_core_common.h
```diff
@@ -17,137 +17,138 @@ extern "C"
...
+    IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF_WITH_JITTER = 5,
+    IOTHUB_CLIENT_RETRY_RANDOM = 6
+} IOTHUB_CLIENT_RETRY_POLICY;
 
-    /** @brief Enumeration passed in by the IoT Hub when the event confirmation
-    *           callback is invoked to indicate status of the event processing in
-    *           the hub.
-    */
-    MU_DEFINE_ENUM_WITHOUT_INVALID(IOTHUB_CLIENT_RETRY_POLICY, IOTHUB_CLIENT_RETRY_POLICY_VALUES);
+const char* IOTHUB_CLIENT_RETRY_POLICYStrings(IOTHUB_CLIENT_RETRY_POLICY value);
```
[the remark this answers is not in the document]
jspaith: Public facing "API".

---
In reply to: [352721449](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r352721449) [](ancestors = 352721449)

iothub_client/inc/iothub_client_core_common.h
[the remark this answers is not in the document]
jspaith: public facing "API".

---
In reply to: [352721229](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r352721229) [](ancestors = 352721229)

jspaith: Kicking off offline discussion about best way to handle this in terms of comments, should it be at bottom of file, etc etc.?

---
In reply to: [353488125](https://github.com/Azure/azure-iot-sdk-c/pull/1342#discussion_r353488125) [](ancestors = 353488125,352722212)

jspaith:
If this ever goes in, see #1714 - we should consider an `AZ_IOT_DEPRECATED` or some of the functions like iothub_client.h which we'd like warning on not being around for.

Issue #1263, opened 2019-10-16
PnP telemetry e2e tests should include validation of the received message at the EventHub  


jasmineymlo:
Current PnP E2E telemetry tests only validate on acknowledgements.  Telemetry should be checked at the receiving end (EventHub endpoint) for completeness.

az-iot-builder-01:
@jasmineymlo, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1361, opened 2019-12-05
Add eventHub validation for telemetry e2e tests. #1263

jasmineymlo:


digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
```diff
@@ -311,6 +312,58 @@ function veryAllInterfacesAreRegistered(done) {
...
+    };
+
+    EventHubClient.createFromIotHubConnectionString(testHubConnectionString)
+        .then((client) => ehClient = client)
+        .then(() => ehClient.getPartitionIds())
+        .then((partitionIds) => {
+          partitionIds.forEach((partitionId) => {
+            ehClient.receive(partitionId, onEventHubMessage, onEventHubError, { eventPosition: EventPosition.fromEnqueuedTime(startAfterTime) });
+          });
+          return new Promise((resolve) => setTimeout(() => resolve(), 3000));
```
jspaith: >3000 [](start = 70, length = 4)

Magic number - please move to a defined constant.

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
```diff
@@ -311,6 +312,58 @@ function veryAllInterfacesAreRegistered(done) {
...
+    let receiveOK = false;
+    let ehClient: EventHubClient;
+
+    const startAfterTime = new Date(Date.now());
+    const onEventHubError = (err) => {
+      throw err;
+    };
+    const onEventHubMessage = (receivedMsg) => {
+        if((receivedMsg.annotations['iothub-connection-device-id'] === testDeviceInfo.deviceId) &&
+        (receivedMsg.annotations['iothub-interface-name'] === testData.interfaceName)){
```
jspaith: >receivedMsg [](start = 9, length = 11)

Could we also please check for a PnP specific annotation here?  I think in current SDK version it's like iothub-interface-instancename (that'll change at some point).  See DT_InterfaceClient_CreateTelemetryMessage() setting DT_INTERFACE_NAME_PROPERTY for what device code specifically I'd like to test is correctly applying this annotation in e2e fashion.  Device Explorer or else Miguel/Sam can help with the exact name of this thing as it comes to EventHub.

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
```diff
@@ -311,6 +312,58 @@ function veryAllInterfacesAreRegistered(done) {
...
+        .then((client) => ehClient = client)
+        .then(() => ehClient.getPartitionIds())
+        .then((partitionIds) => {
+          partitionIds.forEach((partitionId) => {
+            ehClient.receive(partitionId, onEventHubMessage, onEventHubError, { eventPosition: EventPosition.fromEnqueuedTime(startAfterTime) });
+          });
+          return new Promise((resolve) => setTimeout(() => resolve(), 3000));
+        }).then(() => {
+            runTestCommand(testData, 0, (err, responsepPayload, responseStatus) => {
+                err = new Error('Faking error case')
```
jspaith: >err = new Error('Faking error case') [](start = 16, length = 36)

First could we have a better string than "Faking error case"?

I don't want to re-document TypeScript/Mocha in the comments of this test (since we don't anywhere else) but could you please let me know at least in comments or offline what's up with this line?  As a non-Node guy the intent here isn't clear.  We create an error object which should presumably always succeed (almost) and then done(err) when it does?

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
jspaith: Ugh, never mind.  You're doing this already and I missed.  Resolving.

---
In reply to: [354605504](https://github.com/Azure/azure-iot-sdk-c/pull/1361#discussion_r354605504) [](ancestors = 354605504)

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
jasmineymlo: I need to delete the line...  I was just force-testing error cases to see what is the behavior, and accidentally checked that in.  Thanks for catching that!!

PR #1384, opened 2019-12-23
Fix sample dockerfile in cross-compilation documentation

ewertons:
# Checklist
- [] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

# Description of the problem

https://github.com/Azure/azure-iot-sdk-c/issues/1382

# Description of the solution
 
Use the correct way to change directories in Dockerfile.

### DIFF
--- a/doc/Docker_SDK_Cross_Compile.md
+++ b/doc/Docker_SDK_Cross_Compile.md
@@ -64,8 +64,9 @@ RUN tar -xvf OpenWrt-Toolchain-ramips-for-mipsel_24kec+dsp-gcc-4.8-linaro_uClibc
 
 # Download the Azure IoT SDK for C
 RUN git clone https://github.com/azure/azure-iot-sdk-c.git
-RUN cd azure-iot-sdk-c
+WORKDIR azure-iot-sdk-c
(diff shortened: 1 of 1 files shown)


doc/Docker_SDK_Cross_Compile.md
```diff
@@ -64,8 +64,9 @@ RUN tar -xvf OpenWrt-Toolchain-ramips-for-mipsel_24kec+dsp-gcc-4.8-linaro_uClibc
 
 # Download the Azure IoT SDK for C
 RUN git clone https://github.com/azure/azure-iot-sdk-c.git
-RUN cd azure-iot-sdk-c
+WORKDIR azure-iot-sdk-c
```
CIPop: The only observation from Docker would be to use absolute paths if at all possible, for clarity. 
I don't know if that's possible or needed in this case. (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

ewertons: Yes, for this sample we are not using absolute paths. We can update all paths to absolute in a separate fix if we get to prioritize that.

Issue #1362, opened 2019-12-05
IoTHubClientCore_LL_* APIs are publicly documented and header file does not discourage use


drwill-ms:
There are lots of these APIs listed on docs.microsoft.com:

https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-client-core-ll-h
•	https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-client-core-ll-h/iothubclientcore-ll-setdevicemethodcallback
•	https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-client-core-ll-h/iothubclientcore-ll-setdevicemethodcallback-ex

I don't see anything in the [header file](https://github.com/Azure/azure-iot-sdk-c/blob/cf5aaf33e1202b0e87ce054c911e9b979a4cfc59/iothub_client/inc/iothub_client_core_ll.h) to discourage use.

From: John Spaith
The IoTHubClientCore* API’s should never be used directly.  You should be using IoTHubDeviceClient_(LL_)_* or else the module equivalent.

I could’ve sworn that we documented this – especially in the header file itself - but clearly it’s not.   Could you please open up a GitHub issue on this?  Ugh :

You’ll notice is the proper API set there’s only a IoTHubDeviceClient_LL_SetDeviceMethodCallback().  The _Ex_ function is a (poorly named) internal only thingy used between the convenience layer and the _LL_ if I’m remembering correctly, which is why it wasn’t into the device* side.

From: David R. Williamson
Can you point me to any good examples for IoTHubClientCore_LL_SetDeviceMethodCallback_Ex?

ewertons:
@drwill-ms ,
you are absolutely correct. I am not sure what happened with the documentation we had, but we will follow up on this right away.
Thanks.

ewertons:
Hi @drwill-ms ,
after reviewing the documentation we concluded that that particular header file documentation was not supposed to be published, since it is an internal header. 
It has since been removed from the public documentation.
Thanks for bringing this issue to our attention.
-Azure IoT SDK Team.

az-iot-builder-01:
[...]

PR #1390, opened 2020-01-02
Doc explaining LL vs convenience layers

### LINKED ISSUE #1186: Document LL versus convenience layer
We need clearer documentation about when and how to use LL versus convenience layer.

There's some explanation in specs and some references to samples, but this feels like it deserves a separate topic in docs.

jspaith:
Document difference between LL and convenience layer in a single place we can refer customers to.  Fixes #1186.


doc/threading_notes.md
```diff
@@ -0,0 +1,49 @@
+# Threading model and optimizing for resource constrained devices
+
+## IoTHub Device Client and the \_LL\_ (lower layer) API
+
+### The lower layer
+
+There are two distinct but closely related sets of APIs for the IoTHub client.
+Applications that want to run on a single thread should use the \_LL\_ (aka lower level) API.  Sometimes the hardware/OS can only support one thread at a time or else spinning additional threads is expensive.  Applications on more capable hardware may also use the \_LL\_ layer for greater control over scheduling.
+
+Calls to the IoThub API's ending in Async do not cause the operation to execute immediately.  Instead, the work - both for sending data across the network and receiving data - is queued.  The application must manually schedule the work by calling `IoTHubDeviceClient_LL_DoWork` (when using a device client).
```
ewertons: By "when using a device client" you mean instead of using module client? The average reader might not have that distinction in mind.

doc/threading_notes.md
```diff
@@ -0,0 +1,49 @@
...
+
+This is typically done in some sort of simple loop, e.g.
+
+```c
+    IOTHUB_DEVICE_CLIENT_LL_HANDLE iotHubLLHandle;
+
+    while (appIsRunning) {
+        SendTelemetryIfThereIsAnyToSend(iotHubLLHandle);
+        IoTHubDeviceClient_LL_DoWork(iotHubLLHandle);
+        Sleep(1000);
```
ewertons: Our recommended DoWork frequency (at this moment) is at least every one hundred milliseconds. Maybe that should be reflected here.

doc/threading_notes.md
```diff
@@ -0,0 +1,49 @@
...
+
+**CRITICAL(!) notes for \_LL\_ application development**  
+*NOT FOLLOWING THESE STEPS WILL POTENTIALLY RESULT IN VERY HARD TO DIAGNOSE BUGS*
+
+* You **MUST NOT** use an \_LL\_ handle on more than one thread (unless you provide *all* your own locking).  Part of what makes the \_LL\_ layer use fewer resources is that it does not use locks.  It assumes a single threaded environment where locks are not needed.
+* Even if you are not actively sending data while using \_LL\_ layer, you still need to periodically invoke the appropriate `DoWork` to make the client accept data from the network.  This periodic `DoWork` call is also required as otherwise the underlying network transport and/or service may timeout the service due to inactivity.
+
+### The convenience layer
+
+Applications may also use the non \_LL\_ API set, which is also known as the "convenience layer."
```
ewertons: period before closing quotes

doc/threading_notes.md
```diff
@@ -0,0 +1,49 @@
...
+* Even if you are not actively sending data while using \_LL\_ layer, you still need to periodically invoke the appropriate `DoWork` to make the client accept data from the network.  This periodic `DoWork` call is also required as otherwise the underlying network transport and/or service may timeout the service due to inactivity.
+
+### The convenience layer
+
+Applications may also use the non \_LL\_ API set, which is also known as the "convenience layer."  
+
+Just like the \_LL\_ layer, IoTHub API's ending in Async queue work to be performed later and do not block waiting for the service accepting or rejecting the request.  The difference is that the convenience layer itself automatically takes care of sending the data.  In other words, there is no `DoWork`.  The convenience layer does this for you automatically by spinning a worker thread to implicitly `DoWork` for your application.  The conveneince layer also performs locking, allowing a given `IOTHUB_DEVICE_CLIENT_HANDLE` to be safely used by  different threads.
```
ewertons: This kind of comment is a pain in the @#$@#$, sorry... there might be some misspells (conveneince)

doc/threading_notes.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Avoid long-running callbacks
+
+Regardless of whether you are using the \_LL\_ or convenience layer, application callbacks that you implement - such as processing a twin update - must NOT block for extended periods of time.  
+
+The IoTHub SDK uses a single dispatcher thread to handle all callbacks to user code.  This same thread handles all network I/O.  This is true whether the \_LL\_ or convenience layer is used.  The only difference is that in the \_LL\_ layer, your thread is the dispatcher when it calls into the appropriate DoWork() call.
+
+An application callback that takes a long time to run is problematic.  The SDK will not be able to call other pending callbacks as it is blocked on the long-running one.  If the call back code takes long enough (minutes) there is the risk that the SDK will not be able to fire its periodic network keep-alive and that the entire connection will be dropped.
```
ewertons: Not a must, but just asking anyway... should we add a footer note advising that new work should not use iothub_client.h ? I wouldn't say this is the focus of this document, but just in case you agree that could help customers.

ewertons: Indeed there are situations where running a blocking callback for minutes might not even be needed. A single minute blocking could cause the MQTT transport to assume a publish failed because the puback came too late, and start triggering the reconnection logic.

doc/threading_notes.md
jspaith: Fixed, and re-ran through spell-checker.

doc/threading_notes.md
jspaith: I always taught - and some quick research seems to back this up - that the period goes in the quote.  E.G. https://style.mla.org/the-placement-of-a-comma-or-period-after-a-quotation/ for US at least.

doc/threading_notes.md
jspaith: Changed per suggestion.

doc/threading_notes.md
jspaith: Linked to section where this is spelled out in more detail after this bit.

doc/threading_notes.md
jspaith: This is tracked already (#1362) and I think it's a somewhat different thing, since it'd require changing the public headers and docs themselves.  I don't want to try and add too much on this one, so will use #1362 to track.

Issue #1264, opened 2019-10-16
PnP e2e tests suite should take dependency on node service client official released npm package


jasmineymlo:
The service application code in the PnP e2e test suite is written in TypeScript using the Digital Twin service SDK.  We should move to taking dependency on the official released npm package of the Node digital twin service client SDK, instead of a copy of the node sdk in our repo.

az-iot-builder-01:
@jasmineymlo, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1415, opened 2020-02-03
remove local copy of node digitaltwin service SDK and points to the released package for e2e #1264

jasmineymlo:
- add package.json for package dependencies
- remove local copy of node digitaltwin service SDK and point to the released package
- update mocha_exec scripts to do "npm install" which would use package.json to install dependencies.


jspaith:
Great change!  Thanks @jasmineymlo and also @anthonyvercolano for consulting on this.

PR #1433, opened 2020-02-14
Add parameter to 'New-CACertsDevice'

creationsoft:
By default, the 'New-CACertsDevice' function will create a device certificate signed by the Root CA cert. My understanding is that we want to create a device cert signed by the lowest Intermediate CA in the chain. It's possible I've misunderstood this so welcome feedback. Thank you.

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/tools/CACertificates/CACertificateOverview.md
+++ b/tools/CACertificates/CACertificateOverview.md
@@ -85,7 +85,7 @@ On Azure IoT Hub, navigate to the IoT Devices section, or launch Azure IoT Explo
 
 #### IoT Leaf Device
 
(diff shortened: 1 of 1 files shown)


danewalton:
Hi @creationsoft thanks for the feedback. I think in this case though, the device should be signed with the root cert. Couple sentences above in the `.md` it says:

>  Devices can authenticate to IoT Hub using a certificate that is signed by the Root CA from Step 2.

So the default behavior is to have it signed by the Root (for simplicity sake). If you would like, however, you may choose the signing cert as you have and use a chained scenario.

PR #1436, opened 2020-02-18
iothubtransport_mqtt_common: fix compiler issue

fgiancane8:
The source file `src/iothubtransport_mqtt_common.c` may fail to compile
in some stricter compiler environment (flag -Wmaybe-unitialized
enabled).

Fix the issue by explicitly initialize all the variables in all the
possile combinations of code path of execution.

Signed-off-by: Francesco Giancane <francesco.giancane@accenture.com>

 

# Checklist
- [X] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
- [ ] ~~I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).~~
- ~~If this is a modification that impacts the behavior of a public API~~
  - [ ] ~~I edited the corresponding document in the `devdoc` folder and added or modified requirements.~~
  - [X] This pull-request is submitted against the `master` branch. 
  - [X] I have merged the latest `master` branch prior to submission and re-merged as needed after I took any feedback.
  - [X] I have squashed my changes into one with a clear description of the change.

# Description of the problem
`src/iothubtransport_mqtt_common.c` failed to compile because not every code path lead to correct variable assignment.

# Description of the solution
Fixing the issue by explicitly set variables in appropriate manner in every code path.

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -524,12 +524,18 @@ static int parse_device_twin_topic_info(const char* resp_topic, bool* patch_msg,
                 else if (token_count == 3)
                 {
                     *status_code = (int)atol(STRING_c_str(token_value));
+                    *patch_msg = false;
                     if (STRING_TOKENIZER_get_next_token(token_handle, token_value, "/?$rid=") == 0)
                     {
                         *request_id = (size_t)atol(STRING_c_str(token_value));
+                        result = 0;
                     }
-                    *patch_msg = false;
-                    result = 0;
+                    else
+                    {
+                        LogError("Could not parse request_id.");
+                        *request_id = 0;
+                        result = MU_FAILURE;
+                    }
                     break;
                 }
                 token_count++;
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -533,7 +533,6 @@ static int parse_device_twin_topic_info(const char* resp_topic, bool* patch_msg,
                     else
                     {
                         LogError("Could not parse request_id.");
-                        *request_id = 0;
                         result = MU_FAILURE;
                     }
                     break;
@@ -1520,7 +1519,7 @@ static void mqtt_notification_callback(MQTT_MESSAGE_HANDLE msgHandle, void* call
             IOTHUB_IDENTITY_TYPE type = retrieve_topic_type(topic_resp, STRING_c_str(transportData->topic_InputQueue));
             if (type == IOTHUB_TYPE_DEVICE_TWIN)
             {
-                size_t request_id;
+                size_t request_id = 0;
                 int status_code;
                 bool notification_msg;
                 if (parse_device_twin_topic_info(topic_resp, &notification_msg, &request_id, &status_code) != 0)


iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -524,12 +524,18 @@ static int parse_device_twin_topic_info(const char* resp_topic, bool* patch_msg,
...
+                    *patch_msg = false;
                     if (STRING_TOKENIZER_get_next_token(token_handle, token_value, "/?$rid=") == 0)
                     {
                         *request_id = (size_t)atol(STRING_c_str(token_value));
+                        result = 0;
+                    }
+                    else
+                    {
+                        LogError("Could not parse request_id.");
+                        *request_id = 0;
```
ewertons: Hi @fgiancane8 , just one comment, would you mind changing this line of code and verifying again against your compiler?
As a guideline we do not set out variables if the function fails. 
LSS, line 536 should be removed (unless your compiler complains about that).

fgiancane8: Hi @ewertons ,

Thanks for the feedback. I added the `*request_id = 0` assignment because the esp32 compiler (which is a gcc with `-Wall` `-Wextra` flags) complained about usage of a not initialised variable (`-Wmaybe-uninitialized` flag).

There are no real cases of not having the request id, unless serious bugs in place.

But if, as guideline, you do not set the return values when function fails, I think I can rework the PR to explicitly initialise `request_id` in the caller fixing the issue.

fgiancane8: I also take the opportunity to point out that #1394 (which is another PR of mine) addresses same issues (errors found by enabling `-Wall -Wextra`. Could you please have a look at them?

fgiancane8: Hi @ewertons ,

I updated the PR in order to be compliant with guidelines.

However I am not completely happy of the result as `request_id` is initialized to a valid request ID from Azure point of view.

What do you think if we change the `request_id` type from `size_t` to `ssize_t` (or any other signed integer alias) so that we can explicitly initialize request_id to -1 which is invalid from the Cloud point of view.

This is quite a philosophical reasoning, as if the function `parse_device_twin_topic_info` fails the branch taking into account request_id is not executed.
Changing the variable type is not also an interface breaking as it is used in just the function `mqtt_notification_callback` which is static.

Thanks,
Francesco

fgiancane8: LSS, 0 for request_id for the failing branch is OK, -1 would be cleaner.

jspaith:
@fgiancane8  - thanks for this PR and our apologies for having such a lag in getting back to you.

First, could you let me know your compiler version of gcc | OS environment | any extra flags?  I've been trying to reproduce `-Wmaybe-unitialized` using samples from gcc itself on both gcc5, 7, and 9 on Ubuntu18 and Ubuntu20 but without success.  And I have indeed turned on compiler optimizations since I don't think it hits without those.

I worry when I can't reproduce a bug we're hitting though I get how compiler may be unhappy on something like this.

Second, the change looks reasonable including erroring out if there's no RESP_Id from my re-reading of spec.  The MQTT topic parse code has been cleaned up a bunch, though the core of `parseDeviceTwinTopicInfo` there's been a lot of changes.  **My plan is to take core of this commit into a separate PR, make you co-author of course, and then close out this PR.**

Third, I'd rather not set `size_t request_id = 0` in the caller per Ewerton's comments RE not init'ng variables until we know what they should be (or zero'd out here).  If you look at how status_code works in `parseDeviceTwinTopicInfo`, it's set to something in `parseDeviceTwinTopicInfo` even on error cases.  I'll move for `request_id` to similar pattern.

Thanks again

fgiancane8:
Hello @jspaith ,

and thanks for further reviewing my PR! Please find attached my comments inline.

> @fgiancane8 - thanks for this PR and our apologies for having such a lag in getting back to you.
> 
> First, could you let me know your compiler version of gcc | OS environment | any extra flags? I've been trying to reproduce `-Wmaybe-unitialized` using samples from gcc itself on both gcc5, 7, and 9 on Ubuntu18 and Ubuntu20 but without success. And I have indeed turned on compiler optimizations since I don't think it hits without those.
> 
I agree with you. To be honest, in standard environments, if I recall correctly, this issue was not encountered. I found it when porting the SDK to the ESP32 platform. I do not have the precise commit but I was working with `master` branch from `esp-idf` repository and of course latest version of the compiler. the `maybe-uninitialized` flag was suggested by GCC itself.

> I worry when I can't reproduce a bug we're hitting though I get how compiler may be unhappy on something like this.
> 
Sure, let me find a slot of time to review the issue on current machines and double check if I can reproduce myself. (Apart from ESP32)
> Second, the change looks reasonable including erroring out if there's no RESP_Id from my re-reading of spec. The MQTT topic parse code has been cleaned up a bunch, though the core of `parseDeviceTwinTopicInfo` there's been a lot of changes. **My plan is to take core of this commit into a separate PR, make you co-author of course, and then close out this PR.**
> 
Sounds good to me :D
> Third, I'd rather not set `size_t request_id = 0` in the caller per Ewerton's comments RE not init'ng variables until we know what they should be (or zero'd out here). If you look at how status_code works in `parseDeviceTwinTopicInfo`, it's set to something in `parseDeviceTwinTopicInfo` even on error cases. I'll move for `request_id` to similar pattern.
> 
ACK for reworking the step in which out values are not initialized.
> Thanks again

jspaith:
Makes sense!  I'm closing this PR out now because #2143 will supersede it.  I've tagged you so you can follow along.  As I go into more detail on #2143, there's bigger problems than uninitialized variables here that I'm fixing & adding tests for going forward.

Happy to continue conversation if you have more feedback, but would prefer to do it on #2143 which is the open one.

PR #1439, opened 2020-02-19
Add SetOption for appending Model Id in Mqtt Connect.

jasmineymlo:
*** CHANGE PREVIEW ONLY, NOT READY FOR MERGE ***


iothub_client/inc/internal/iothub_internal_consts.h
```diff
@@ -9,7 +9,8 @@ extern "C"
 {
 #endif
 
-    static const char* IOTHUB_API_VERSION = "2019-07-01-preview";
+    //static const char* IOTHUB_API_VERSION = "2019-07-01-preview";
```
jspaith: > //static const char* IOTHUB_API_VERSION = "2019-07-01-preview"; [](start = 3, length = 64)

Please remove commented out line.

iothub_client/inc/iothub_client_options.h
```diff
@@ -41,6 +41,7 @@ extern "C"
     static STATIC_VAR_UNUSED const char* OPTION_MESSAGE_TIMEOUT = "messageTimeout";
     static STATIC_VAR_UNUSED const char* OPTION_BLOB_UPLOAD_TIMEOUT_SECS = "blob_upload_timeout_secs";
     static STATIC_VAR_UNUSED const char* OPTION_PRODUCT_INFO = "product_info";
+    static STATIC_VAR_UNUSED const char* OPTION_DT_MODEL_ID = "dt_model_id";
```
jspaith: >dt_model_id [](start = 63, length = 11)

Let's sync to talk about best name for this.  I'm concerned with underlying IoTHub layer "knowing" anything about DT concepts specifically (even though it's coded correctly RE no hard-dependency).

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -2319,6 +2340,24 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_LL_SetOption(IOTHUB_CLIENT_CORE_LL_HANDLE
...
         }
+        else if (strcmp(optionName, OPTION_DT_MODEL_ID) == 0)
+        {
+            if (handleData->dt_model_id != NULL)
+            {
+                STRING_delete(handleData->dt_model_id);
+                handleData->dt_model_id = NULL;
+            }
+
+            if ((handleData->dt_model_id = STRING_construct_sprintf("&%s=%s", DT_MODEL_ID_TOKEN, STRING_c_str(URL_EncodeString((const char*)value)))) == NULL)
```
jspaith: >if ((handleData->dt_model_id = STRING_construct_sprintf("&%s=%s", DT_MODEL_ID_TOKEN, STRING_c_str(URL_EncodeString((const char*)value)))) == NULL) [](start = 12, length = 146)

I get it's pedantic, but we shouldn't URL encode at this stage because if (!!) we ever wanted AMQP to callback into this, the URL encode wouldn't make sense or it.

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -615,6 +619,22 @@ static const char* IoTHubClientCore_LL_GetProductInfo(void* ctx)
     return result;
 }
 
+static const char* IoTHubClientCore_LL_GetDTModelId(void* ctx)
```
jspaith: >IoTHubClientCore_LL_GetDTModelId [](start = 19, length = 32)

The design here is generally right, but I wonder instead could you extend out IoTHubClientCore_LL_GetProductInfo() so that maybe it's like

int IoTHubClientCore_LL_GetProductInfo(char** product_ino, char** dt_model_id); // once we get model_id name agreed on.

This means there's just one callback to worry with and if hypothetically we added more fields to this later, it'd just be extended out this fcn instead of new ones.

In the MQTT layer, it would check the value(s) returned here and then do right thing.
?

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -2319,6 +2340,24 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_LL_SetOption(IOTHUB_CLIENT_CORE_LL_HANDLE
...
+        else if (strcmp(optionName, OPTION_DT_MODEL_ID) == 0)
+        {
+            if (handleData->dt_model_id != NULL)
+            {
+                STRING_delete(handleData->dt_model_id);
+                handleData->dt_model_id = NULL;
+            }
+
+            if ((handleData->dt_model_id = STRING_construct_sprintf("&%s=%s", DT_MODEL_ID_TOKEN, STRING_c_str(URL_EncodeString((const char*)value)))) == NULL)
+            {
```
jspaith: This is hard-coding an MQTT construct into the client layer, which is theoretically transport agnostic.  It seems here we should save off the dt_model_id as-is and then let the underlying layer do the needed work.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2254,6 +2254,15 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
             }
         }
 
+        const char* dt_model_id = transport_data->transport_callbacks.dt_model_id_cb(transport_data->transport_ctx);
```
jspaith: >dt_model_id_cb [](start = 70, length = 14)

Please see my comment about potentially using the get_prodcut_info() as one-stop shop for this instead.

If we leave this explicit callback here, I believe that we would need to have this block of code in the isProductInfoSet block (or potentially in its own isDtSet, which could also argue for using the product_info() as only call).

jspaith:
>     if ((result = DigitalTwin_DeviceClient_RegisterInterfacesAsync(dtDeviceClientHandle, DIGITALTWIN_SAMPLE_DEVICE_CAPABILITY_MODEL_ID, interfaceClientHandles, numInterfaceClientHandles, DigitalTwinSampleDevice_InterfacesRegistered, &appDigitalTwinRegistrationStatus)) != DIGITALTWIN_CLIENT_OK)

Just want to make sure you're tracking need to remove DCM from this call?  It's OK if not this checkin as it won't break anything to send this extra telemetry, but ideally would be a fast-follow up.

---
Refers to: digitaltwin_client/samples/digitaltwin_sample_device/digitaltwin_sample_device.c:118 in 2f9a6d6. [](commit_id = 2f9a6d61728f97fdb40aa92146912527b8b12bfd, deletion_comment = False)

jasmineymlo:
pending ut updates

digitaltwin_client/inc/digitaltwin_device_client.h
```diff
@@ -51,10 +51,11 @@ typedef void* DIGITALTWIN_DEVICE_CLIENT_HANDLE;
   @param    deviceHandle[in]            An IOTHUB_DEVICE_CLIENT_HANDLE that has been already created and bound to a specific connection string (or transport, or DPS handle, or whatever
                                         mechanism is preferred).  See remarks about its lifetime management.
   @param    dtDeviceClientHandle[out]   A <c>DIGITALTWIN_DEVICE_CLIENT_HANDLE</c> to be used by the application.
- 
+  @param    deviceCapabilityModel[in]   The device's Device Capability Model (DCM) identifier.
```
jspaith: >device's [](start = 44, length = 8)

Nit: I think "device's Device Capability Model" could just be "Device Capability Model".

digitaltwin_client/inc/digitaltwin_device_client.h
```diff
@@ -51,10 +51,11 @@ typedef void* DIGITALTWIN_DEVICE_CLIENT_HANDLE;
   @param    deviceHandle[in]            An IOTHUB_DEVICE_CLIENT_HANDLE that has been already created and bound to a specific connection string (or transport, or DPS handle, or whatever
                                         mechanism is preferred).  See remarks about its lifetime management.
   @param    dtDeviceClientHandle[out]   A <c>DIGITALTWIN_DEVICE_CLIENT_HANDLE</c> to be used by the application.
- 
+  @param    deviceCapabilityModel[in]   The device's Device Capability Model (DCM) identifier. 
+
   @returns  A DIGITALTWIN_CLIENT_RESULT.
 */
-MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DigitalTwin_DeviceClient_CreateFromDeviceHandle, IOTHUB_DEVICE_CLIENT_HANDLE, deviceHandle, DIGITALTWIN_DEVICE_CLIENT_HANDLE*, dtDeviceClientHandle);
+MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DigitalTwin_DeviceClient_CreateFromDeviceHandle, IOTHUB_DEVICE_CLIENT_HANDLE, deviceHandle, DIGITALTWIN_DEVICE_CLIENT_HANDLE*, dtDeviceClientHandle, const char*, deviceCapabilityModel);
```
jspaith: >const char*, deviceCapabilityModel [](start = 196, length = 34)

The deviceCapabilityModel as an IN parameter should (by convention rest of SDK uses) come before the dtDeviceClientHandle, which is the OUT param.  Similar feedback for the _LL_ version of this function.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -668,19 +668,16 @@ static DT_SEND_TELEMETRY_CALLBACK_CONTEXT* CreateDTSendTelemetryCallbackContext(
 
 
 // RegisterDTInterfaces_Callback is called when we receive acknowledgement from the service that the message we sent registering the interfaces was received.
-static void RegisterDTInterfaces_Callback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, void* userContextCallback)
+static void RegisterDTInterfaces_Callback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, DT_CLIENT_CORE* dtClientCore)
 {
     LogInfo("DigitalTwin Client Core: Processing register DigitalTwin interfaces callback.  confirmationResult=%s", MU_ENUM_TO_STRING(IOTHUB_CLIENT_CONFIRMATION_RESULT,result));
 
-    if (userContextCallback == NULL)
+    if (dtClientCore == NULL)
```
jspaith: >if (dtClientCore == NULL) [](start = 4, length = 25)

Because this is a static function, we shouldn't explicitly check the dtClientCore as non-NULL.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -849,21 +801,19 @@ DIGITALTWIN_CLIENT_RESULT DT_ClientCoreRegisterInterfacesAsync(DT_CLIENT_CORE_HA
...
         dtClientCore->registerInterfacesCallbackContext.userContextCallback = userContextCallback;
-    
+
         // Once we're processing an interface update, no other caller initiated operations on this device client can occur.
         dtClientCore->registrationStatus = DT_CLIENT_REGISTRATION_STATUS_REGISTERING;
         result = DIGITALTWIN_CLIENT_OK;
     }
 
     (void)InvokeBindingUnlock(dtClientCore, &lockHeld);
+    RegisterDTInterfaces_Callback(IOTHUB_CLIENT_CONFIRMATION_OK, dtClientCore);
```
jspaith: >RegisterDTInterfaces_Callback [](start = 4, length = 29)

We should only call this function if (result==DIGITALTWIN_CLIENT_OK).

jspaith:
>     if ((deviceHandle == NULL) || (dtDeviceClientHandle == NULL))

Also per existing conventions, please check deviceCapabilityModel==NULL here (and also in the _LL_ equivalent code here).

---
Refers to: digitaltwin_client/src/dt_device.c:107 in 6abcea2. [](commit_id = 6abcea23b779009993dae332f583fd9df77f65fa, deletion_comment = False)

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2217,7 +2219,7 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
 
     if (result == 0)
     {
-        if (!transport_data->isProductInfoSet)
+        if (!transport_data->isOptionalParameterSet)
         {
```
jspaith: >{ [](start = 8, length = 1)

The block between the { and } here is up to +50 lines.  Please create a static helper function to do this work here.

jspaith: Also could we see about not having the if/nesting, e.g.

if (doSomething1 == error) 
{
   error
}
else 
{
   if (doSomething2 == error) 
   {
     error
   }
   else 
    {
         ... continuing to go deeper

and move to the more standard

if (doSomething1 == error) 
{
   error
 }
else if (doSomething2 == error) 
{
   error
}

---
In reply to: [385295048](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385295048) [](ancestors = 385295048)

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2247,7 +2249,19 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
                 }
                 else
                 {
-                    transport_data->isProductInfoSet = true;
+                    STRING_HANDLE param;
+                    const char* dt_model_id = transport_data->transport_callbacks.dt_model_id_cb(transport_data->transport_ctx);
+                    if (dt_model_id != NULL)
+                    {
+                        param = STRING_construct_sprintf("&%s=%s", DT_MODEL_ID_TOKEN, STRING_c_str(URL_EncodeString(dt_model_id)));
```
jspaith: >param [](start = 24, length = 5)

No error checking on STRING_contruct_sprintf() failing.

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -2319,6 +2336,24 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_LL_SetOption(IOTHUB_CLIENT_CORE_LL_HANDLE
                 result = IOTHUB_CLIENT_OK;
             }
         }
+        else if (strcmp(optionName, OPTION_DT_MODEL_ID) == 0)
+        {
+            if (handleData->dt_model_id != NULL)
```
jspaith: >if (handleData->dt_model_id != NULL) [](start = 12, length = 36)

We shouldn't let callers pass this multiple times, since the hub itself requires tearing down connection on a new DCM.  I think we should fail here instead if already set.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -668,19 +668,16 @@ static DT_SEND_TELEMETRY_CALLBACK_CONTEXT* CreateDTSendTelemetryCallbackContext(
 
 
 // RegisterDTInterfaces_Callback is called when we receive acknowledgement from the service that the message we sent registering the interfaces was received.
-static void RegisterDTInterfaces_Callback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, void* userContextCallback)
+static void RegisterDTInterfaces_Callback(IOTHUB_CLIENT_CONFIRMATION_RESULT result, DT_CLIENT_CORE* dtClientCore)
```
jspaith: >RegisterDTInterfaces_Callback [](start = 12, length = 29)

We should discuss more and I need to stare at code a bit more, but I *think* this entire function should run under the lock grabbed at beggining of the registerInterfaces.

The BeginClientCoreCallbackProcessing() that other callbacks grab will make sure that even if a callback arrives in middle of this function, it won't be processed until we're complete here.

The IoTHub layer *never* calls into a callback with a lock held, so there's not a deadlock happening here (where the DT.Lock() and calls into say invokeSetDevicetwin callback to do a IoTHub.Lock(), and then a callback would do a IoTHub.Lock() and a DT.Lock(); again per the IoTHub not locking on callbacks, this can't happen).

BeginClientCoreCallbackProcessing() is specifically designed to run in callbacks, not in the user thread.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2247,7 +2249,19 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
...
+
+                        if (STRING_concat_with_STRING(transport_data->configPassedThroughUsername, param) != 0)
+                        {
+                            LogError("Failed to set DT Model Id");
+                        }
+                    }
+
+                    transport_data->isOptionalParameterSet = true;
```
danewalton: should this be set even after a LogError failure?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -236,7 +238,7 @@ typedef struct MQTTTRANSPORT_HANDLE_DATA_TAG
     int http_proxy_port;
     char* http_proxy_username;
     char* http_proxy_password;
-    bool isProductInfoSet;
+    bool isOptionalParameterSet;
```
jspaith: >isOptionalParameterSet [](start = 9, length = 22)

I think "optionalParameter" should be more specific.  Something like isOptionalConnect or some such?

jasmineymlo: OptionalConnectParameter?

---
In reply to: [385380333](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385380333) [](ancestors = 385380333)

jspaith: Sure, but should proabbly actually be like isOptionalConnectParameterSet?

jasmineymlo: yes

---
In reply to: [385381471](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385381471) [](ancestors = 385381471)

iothub_client/src/iothubtransport_mqtt_common.c
jasmineymlo: the optional parameters are only supposed to be set once 

---
In reply to: [385312375](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385312375) [](ancestors = 385312375)

danewalton: sure but even if it fails? I'm ignorant on this so not sure. Just seems odd to have it fail but then say `true` that it was set.

digitaltwin_client/samples/digitaltwin_sample_device/digitaltwin_sample_device.c
```diff
@@ -72,7 +72,7 @@ static const int digitalTwinSampleDevice_sendTelemetryFrequency = 20;
 //
 
 // TODO: Fill in DIGITALTWIN_SAMPLE_DEVICE_CAPABILITY_MODEL_ID. E.g. 
-#define DIGITALTWIN_SAMPLE_DEVICE_CAPABILITY_MODEL_ID "urn:YOUR_COMPANY_NAME_HERE:sample_device:1"
+#define DIGITALTWIN_SAMPLE_DEVICE_CAPABILITY_MODEL_ID "urn:MyCompanyName:sample_device:1"
```
jspaith: >MyCompanyName [](start = 59, length = 13)

Is there a reason we changed this or should it be reverted?

digitaltwin_client/inc/digitaltwin_device_client.h
```diff
@@ -75,7 +76,7 @@ MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DigitalTwin_DeviceClient_CreateFr
  
   @returns  A DIGITALTWIN_CLIENT_RESULT.
 */
-MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DigitalTwin_DeviceClient_RegisterInterfacesAsync, DIGITALTWIN_DEVICE_CLIENT_HANDLE, dtDeviceClientHandle, const char*, deviceCapabilityModel, DIGITALTWIN_INTERFACE_CLIENT_HANDLE*, dtInterfaces, unsigned int, numDTInterfaces, DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK, dtInterfaceRegisteredCallback, void*, userContextCallback);
+MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DigitalTwin_DeviceClient_RegisterInterfacesAsync, DIGITALTWIN_DEVICE_CLIENT_HANDLE, dtDeviceClientHandle, DIGITALTWIN_INTERFACE_CLIENT_HANDLE*, dtInterfaces, unsigned int, numDTInterfaces, DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK, dtInterfaceRegisteredCallback, void*, userContextCallback);
```
jspaith: >DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK [](start = 236, length = 41)

Just occurred to me but I don't think we need DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK anymore, do we?  There's nothing truly async as far as caller is concerned.  So we could remove the dt_client_core.c's DT_REGISTER_INTERFACES_CALLBACK_CONTEXT stuff also.  We still would want to leave the callbacks into interfaces DT_InterfaceList_RegistrationCompleteCallback() around though.  If we remove this callback, it also simplifies the samples so there's not the polling.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -667,92 +667,27 @@ static DT_SEND_TELEMETRY_CALLBACK_CONTEXT* CreateDTSendTelemetryCallbackContext(
...
-    {
-        LogError("userContextCallback=%p.  Cannot process callback", userContextCallback);
-    }
-    else
-    {
-        DT_SEND_TELEMETRY_CALLBACK_CONTEXT* dtSendTelemetryCallbackContext = (DT_SEND_TELEMETRY_CALLBACK_CONTEXT*)userContextCallback;
-        DT_CLIENT_CORE* dtClientCore = dtSendTelemetryCallbackContext->dtClientCore;
+    // We start listening for incoming commands and properties after binding is invoked.
+    // If either of these fail, we will report an error to the application.
+    LogInfo("DigitalTwin Client Core: Interfaces successfully registered.  Register for device method and twin callbacks if needed");
```
jspaith: >LogInfo("DigitalTwin Client Core: Interfaces successfully registered.  Register for device method and twin callbacks if needed"); [](start = 4, length = 129)

I don't think we should have this LogInfo() anymore.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -849,21 +779,23 @@ DIGITALTWIN_CLIENT_RESULT DT_ClientCoreRegisterInterfacesAsync(DT_CLIENT_CORE_HA
...
         // Once we're processing an interface update, no other caller initiated operations on this device client can occur.
         dtClientCore->registrationStatus = DT_CLIENT_REGISTRATION_STATUS_REGISTERING;
         result = DIGITALTWIN_CLIENT_OK;
     }
 
     (void)InvokeBindingUnlock(dtClientCore, &lockHeld);
+
+    if (result == DIGITALTWIN_CLIENT_OK)
+    {
+        InvokeBindingOnDemand(dtClientCore);
```
jspaith: >InvokeBindingOnDemand [](start = 8, length = 21)

#1
This needs to be called while the BindingLock is still held.  So it'd be something like

else if (result=DT_Interface_BindInterfaces()...) 
{
  // error
}
else if (result = InvokeBindingOnDemand() ...)
{

}

#2
Instead of InvokeBindingOnDemand(), maybe something like SubscribeForIncomingRequests() or something.

#3
SetRegistrationCompleteInvokeCallbacks() however needs to not be called with the lock.  Here I recommenda  new helper function, *only* called on success of everything here of course, and that does do the callback stuff.

Something like

else if (...) // blah, core things
{
...
}
(void)InvokeBindingUnlock(dtClientCore, &lockHeld);

    if (result == DIGITALTWIN_CLIENT_OK)
    {
        SetRegistrationCompleteInvokeCallbacks(dtClientCore);
    }

And in then new SetRegistrationCompleteInvokeCallbacks(), you'll need to do the BeginClientCoreCallbackProcessing() and EndClientCoreCallbackProcessing().  These make sure that handle can't be killed under us while we invoke a callback routiune, but at same time the lock for dt_client_core is NOT held so we do'nt run into deadlocks.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,49 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
     }
 }
 
+static void append_optional_connect_parameters(PMQTTTRANSPORT_HANDLE_DATA transport_data)
```
jspaith: >append_optional_connect_parameters [](start = 12, length = 34)

This function should return an error, since failirue to set DCM specifically should cause fatal errors.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,49 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+            LogError("Cannot build device capability model id string");
+        }
+        else if (STRING_concat_with_STRING(transport_data->configPassedThroughUsername, param) != 0)
+        {
+            LogError("Failed to set device capability model id parameter in connect");
+        }
+        else
+        {
+            transport_data->isOptionalConnectParameterSet = true;
+            STRING_delete(clone);
```
jspaith: >STRING_delete(clone); [](start = 12, length = 21)

We need to free these outside of the block, because if there's a failure above we could allocate this stuff but not free.

If you set clone=NULL & param=NULL during initialization, you can just do an unconditional STRING_Delete() always and it'll correctly handle NULL cases.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,49 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+
+        if ((clone = (product_info == NULL) ? STRING_construct_sprintf("%s%%2F%s", CLIENT_DEVICE_TYPE_PREFIX, IOTHUB_SDK_VERSION) : URL_EncodeString(product_info)) == NULL)
+        {
+            LogError("Failed obtaining the product info");
+        }
+        else if (STRING_concat_with_STRING(transport_data->configPassedThroughUsername, clone) != 0)
+        {
+            LogError("Failed concatenating the product info");
+        }           
+        else if ((dt_model_id = transport_data->transport_callbacks.dt_model_id_cb(transport_data->transport_ctx)) == NULL)
```
jspaith: >((dt_model_id = transport_data->transport_callbacks.dt_model_id_cb(transport_data->transport_ctx)) == NULL) [](start = 16, length = 107)

This is not a fatal error.  Need to handle cases where this is and isn't set.

jspaith: In particular there's a building of param token below no matter what happens here.

---
In reply to: [385933472](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385933472) [](ancestors = 385933472)

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,49 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+        {
+            LogInfo("Not setting device capability model id");
+        }
+        else if ((param = STRING_construct_sprintf("&%s=%s", DT_MODEL_ID_TOKEN, STRING_c_str(URL_EncodeString(dt_model_id)))) == NULL)
```
jspaith: >URL_EncodeString [](start = 93, length = 16)

Mem leak on URL_EncodeString() here.  This needs to be put into its own STRING_HANDLE and then indpendently freed.

jspaith:
>     else if (dtClientCore->registrationStatus == DT_CLIENT_REGISTRATION_STATUS_REGISTERING)

Can get rid of DT_CLIENT_REGISTRATION_STATUS_REGISTERING state :).

---
Refers to: digitaltwin_client/src/dt_client_core.c:768 in 9b9d547. [](commit_id = 9b9d54743a067df38cc8fd31b1e06151497cb6f2, deletion_comment = False)

iothub_client/src/iothubtransport_mqtt_common.c
jasmineymlo: param token would not be build if dt_model_id is NULL

---
In reply to: [385933779](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r385933779) [](ancestors = 385933779,385933472)

digitaltwin_client/inc/digitaltwin_device_client.h
jasmineymlo: I have preliminarily removed DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK and related context stuff for your review.  A caveat is that the application do not get any insight/notification of the state of connection, thus interfaces would start operation despite mqtt connect failure.  If we determined that this is by-design for refresh, I will make additional change in sample LL and e2e to remove APP_DIGITALTWIN_REGISTRATION_STATUS_TAG related code similar to what I have already done in digitialtwin_sample_device.c.

jspaith: Per offline email, I think this is right.  Next time I'm in we can discuss more f2f.

---
In reply to: [386687567](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r386687567) [](ancestors = 386687567)

digitaltwin_client/samples/digitaltwin_sample_device/digitaltwin_sample_device.c
```diff
@@ -78,73 +78,17 @@ static const int digitalTwinSampleDevice_sendTelemetryFrequency = 20;
...
-        LogError("Interface registration callback invoked with an error=<%s>", MU_ENUM_TO_STRING(DIGITALTWIN_CLIENT_RESULT,dtInterfaceStatus));
-        *appDigitalTwinRegistrationStatus = APP_DIGITALTWIN_REGISTRATION_FAILED;
-    }
-}
-
-APP_DIGITALTWIN_REGISTRATION_STATUS appDigitalTwinRegistrationStatus = APP_DIGITALTWIN_REGISTRATION_PENDING;
-
 // Invokes DigitalTwin_DeviceClient_RegisterInterfacesAsync, which indicates to Azure IoT which DigitalTwin interfaces this device supports.
 // The DigitalTwin Handle *is not valid* until this operation has completed (as indicated by the callback DigitalTwinSampleDevice_RegisterDigitalTwinInterfacesAndWait being invoked).
 static DIGITALTWIN_CLIENT_RESULT DigitalTwinSampleDevice_RegisterDigitalTwinInterfacesAndWait(DIGITALTWIN_DEVICE_CLIENT_HANDLE dtDeviceClientHandle, DIGITALTWIN_INTERFACE_CLIENT_HANDLE* interfaceClientHandles, int numInterfaceClientHandles)
```
jspaith: >DigitalTwinSampleDevice_RegisterDigitalTwinInterfacesAndWait [](start = 33, length = 60)

Let's remove this helper function since AndWait() is N/A now.  Can just move the DigitalTwin_DeviceClient_RegisterInterfacesAsync back into the main().

jspaith: Same feedback with LL version...

---
In reply to: [387337016](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r387337016) [](ancestors = 387337016)

digitaltwin_client/src/dt_client_core.c
```diff
@@ -506,23 +496,7 @@ DT_CLIENT_CORE_HANDLE DT_ClientCoreCreate(DT_IOTHUB_BINDING* iotHubBinding, DT_L
 // themselves of the status of the registration.
 static void SetRegistrationCompleteInvokeCallbacks(DT_CLIENT_CORE* dtClientCore, DIGITALTWIN_CLIENT_RESULT dtInterfaceStatus)
```
jspaith: >SetRegistrationCompleteInvokeCallbacks [](start = 12, length = 38)

Please remove this function and just invoke DT_InterfaceList_RegistrationCompleteCallback() directly.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -849,21 +724,30 @@ DIGITALTWIN_CLIENT_RESULT DT_ClientCoreRegisterInterfacesAsync(DT_CLIENT_CORE_HA
     {
         LogError("DT_InterfaceList_BindInterfaces failed, result = %d", result);
     }
-    else if ((result = SendInterfacesToRegisterMessage(dtClientCore, deviceCapabilityModel)) != DIGITALTWIN_CLIENT_OK)
+    else if (InvokeBindingSetDeviceMethodCallbackIfNeeded(dtClientCore, DTDeviceMethod_Callback) != 0)
+    {
+        // We start listening for incoming commands and properties after binding is invoked.
+        // If either of these fail, we will report an error to the application.
+        LogError("Cannot bind to device method callbacks");
```
jspaith: >callbacks [](start = 47, length = 9)

Nit: Please include the result in the LogError() message, and for the SetDeviceTwinCallbackIfNeeded() call too.

This will help make it easier to diagnose errors from logs only.

jspaith:
>         for (int i = 0; (i < digitalTwin_E2E_registerInterfaceMaxPolls) && (appDigitalTwinRegistrationStatus == APP_DIGITALTWIN_REGISTRATION_PENDING); i++)

Per removing callback being right thin, this code & and related from tests can be removed.

---
Refers to: digitaltwin_client/tests/dt_e2e/device/main.c:113 in 9cca65c. [](commit_id = 9cca65cc3a5cd1939f8dde14fc2ee4f594546829, deletion_comment = False)

jspaith:
>     // the this executable's CRT and the Node.JS integration into stdout/stderr.  We want the

Not your fault but could you please fix it anyway while you're in here :) -- the words "the this" should just be "this".

---
Refers to: digitaltwin_client/tests/dt_e2e/device/main.c:244 in 9cca65c. [](commit_id = 9cca65cc3a5cd1939f8dde14fc2ee4f594546829, deletion_comment = False)

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
```diff
@@ -463,7 +463,10 @@ describe("DigitalTwin Properties E2E tests", function() {
         verifyPropertySet(done, testPropertyInterfaceName, "ProcessUpdatedProperty1", `ValueOfPropertyBeingSet_1`, 0)
     })
 
-    it("Verify all interfaces registered", function(done) {
-        veryAllInterfacesAreRegistered(done)
-    })
+    // Disable test requires updates from node service client sdk and the test code to support changes for Refresh.
```
jspaith: >// Disable test requires updates from node service client sdk and the test code to support changes for Refresh [](start = 4, length = 110)

Is their a work item somewhere tracking this work?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,62 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
     }
 }
 
+static int append_optional_connect_parameters(PMQTTTRANSPORT_HANDLE_DATA transport_data)
+{
+    int result = 0;
```
jspaith: >int result = 0 [](start = 4, length = 14)

Please leave result as uninitialized.  This will make it so compiler will catch any places that we don't set this.

This would mean we'd need to set result=MU_FAILURE explicitly in the various cases like out-of-memory.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2217,42 +2275,7 @@ static int SendMqttConnectMsg(PMQTTTRANSPORT_HANDLE_DATA transport_data)
...
-                }
-                else
-                {
-                    transport_data->isProductInfoSet = true;
-                }
-
-                STRING_delete(clone);
-            }
-        }
+        append_optional_connect_parameters(transport_data);
```
jspaith: >append_optional_connect_parameters [](start = 8, length = 34)

If this fails, we should fail out the function.  Only reason it would fail would be like on out of memory or bad URL encoding, all which points to huge problems in app and/or device state.

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -2319,6 +2336,23 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_LL_SetOption(IOTHUB_CLIENT_CORE_LL_HANDLE
...
+            {
+                LogError("Failed setting the Device Capability Model ID");
```
jspaith: >Failed setting the Device Capability Model ID [](start = 26, length = 45)

Please put a more meaningful error, like DT Model already specified.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2174,6 +2176,62 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+        {
+            LogError("Failed obtaining the product info");
+        }
+        else if (STRING_concat_with_STRING(transport_data->configPassedThroughUsername, clone) != 0)
+        {
+            LogError("Failed concatenating the product info");
+        }           
+        else if ((dt_model_id = transport_data->transport_callbacks.dt_model_id_cb(transport_data->transport_ctx)) == NULL)
+        {
+            LogInfo("Not setting device capability model id");
```
jspaith: >LogInfo("Not setting device capability model id"); [](start = 12, length = 50)

Please remove this LogInfo() - it'll be too verbose in the non-PnP scenarios.

Also having an early exit here with the dt_model_id==NULL is not how we typcially code rest of SDK (where falling into an else() of a long block like this means error) so I worry about it confusion a casual reader.

I think a cleaner way will be 

else if ((dt_model_id = blah()  != NULL)
{
     put all the code for handling dt_model_id into this block.  You'll need to have a final "result=0".
}
else {
  result = 0;
}

digitaltwin_client/src/dt_device.c
```diff
@@ -148,9 +153,9 @@ DIGITALTWIN_CLIENT_RESULT DigitalTwin_DeviceClient_CreateFromDeviceHandle(IOTHUB
     return result;
 }
 
-DIGITALTWIN_CLIENT_RESULT DigitalTwin_DeviceClient_RegisterInterfacesAsync(DIGITALTWIN_DEVICE_CLIENT_HANDLE dtDeviceClientHandle, const char* deviceCapabilityModel, DIGITALTWIN_INTERFACE_CLIENT_HANDLE* dtInterfaces, unsigned int numDTInterfaces, DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK dtInterfaceRegisteredCallback, void* userContextCallback)
+DIGITALTWIN_CLIENT_RESULT DigitalTwin_DeviceClient_RegisterInterfacesAsync(DIGITALTWIN_DEVICE_CLIENT_HANDLE dtDeviceClientHandle, DIGITALTWIN_INTERFACE_CLIENT_HANDLE* dtInterfaces, unsigned int numDTInterfaces)
 {
-    return DT_ClientCoreRegisterInterfacesAsync((DT_CLIENT_CORE_HANDLE)dtDeviceClientHandle, deviceCapabilityModel, dtInterfaces, numDTInterfaces, dtInterfaceRegisteredCallback, userContextCallback);
```
jasmineymlo: >DT_ClientCoreRegisterInterfacesAsync [](start = 11, length = 36)

Do we remove async here??

iothub_client/src/iothubtransport_mqtt_common.c
jasmineymlo: I have refactor with your recommendation but we are breaking the pattern:
if (error)
{ 
    error handling;
}
else if (error)
{
    error handling;
} ...

also, the result is initialized as 0 anyways.  But I did the assignment as suggested.


---
In reply to: [387346346](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r387346346) [](ancestors = 387346346)

digitaltwin_client/tests/dt_e2e/service/dt_e2e_test.ts
jasmineymlo: yes, its tracked in devops TASK 6366438

digitaltwin_client/inc/internal/dt_client_core.h
```diff
@@ -61,7 +61,7 @@ MOCKABLE_FUNCTION(, DT_CLIENT_CORE_HANDLE, DT_ClientCoreCreate, DT_IOTHUB_BINDIN
 MOCKABLE_FUNCTION(, void, DT_ClientCoreDestroy, DT_CLIENT_CORE_HANDLE, dtClientCoreHandle);
 MOCKABLE_FUNCTION(, void, DT_ClientCoreDoWork, DT_CLIENT_CORE_HANDLE, dtClientCoreHandle);
 
-MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DT_ClientCoreRegisterInterfacesAsync, DT_CLIENT_CORE_HANDLE, dtClientCoreHandle, const char*, deviceCapabilityModel, DIGITALTWIN_INTERFACE_CLIENT_HANDLE*, dtInterfaces , unsigned int, numDTInterfaces, DIGITALTWIN_INTERFACE_REGISTERED_CALLBACK, dtInterfaceRegisteredCallback, void*, userContextCallback);
+MOCKABLE_FUNCTION(, DIGITALTWIN_CLIENT_RESULT, DT_ClientCoreRegisterInterfacesAsync, DT_CLIENT_CORE_HANDLE, dtClientCoreHandle, DIGITALTWIN_INTERFACE_CLIENT_HANDLE*, dtInterfaces , unsigned int, numDTInterfaces);
```
jspaith: >DT_ClientCoreRegisterInterfacesAsync [](start = 47, length = 36)

Should remove the "Async" because caller isn't Async anmore.

digitaltwin_client/src/dt_client_core.c
```diff
@@ -849,21 +724,33 @@ DIGITALTWIN_CLIENT_RESULT DT_ClientCoreRegisterInterfacesAsync(DT_CLIENT_CORE_HA
...
-        dtClientCore->registrationStatus = DT_CLIENT_REGISTRATION_STATUS_REGISTERING;
+        dtClientCore->registrationStatus = DT_CLIENT_REGISTRATION_STATUS_REGISTERED;
         result = DIGITALTWIN_CLIENT_OK;
     }
 
     (void)InvokeBindingUnlock(dtClientCore, &lockHeld);
+
+    if (result == DIGITALTWIN_CLIENT_OK)
+    {
+        // Mark internal state as complete and signal to application the status of the DigitalTwin registration request.
```
jspaith: >application [](start = 57, length = 11)

Nit: This is more specifically the "interface's" instead of "application".

digitaltwin_client/src/dt_device.c
jspaith: Yes.

---
In reply to: [387346573](https://github.com/Azure/azure-iot-sdk-c/pull/1439#discussion_r387346573) [](ancestors = 387346573)

PR #1457, opened 2020-03-10
opt out of codesign

danewalton:


build/.vsts-ci.yml
```diff
@@ -1,4 +1,6 @@
 name: $(BuildID)_$(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+variables:
+  runCodesignValidationInjection: true
```
jbobotek: shouldn't it be false? Also, I'm trying something similar from the DevOps angle rather than through the yaml.

PR #1460, opened 2020-03-11
submodule umock update

ttins:
Update submodule references:
Fix unit tests to adhere to new umock aliasing structure


iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
     return handle;
 }
 
+char* umock_stringify_AMQP_MESSENGER_MESSAGE_DISPOSITION_INFO(const AMQP_MESSENGER_MESSAGE_DISPOSITION_INFO* value)
+{
```
ewertons: Tye, do you know when is stringify used? is it just for logging?
Generating an empty string doesn't seem to be the correct way. What do you think?

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
...
+    (void)value;
+    result[0] = '\0';
+    return result;
+}
+
+int umock_are_equal_AMQP_MESSENGER_MESSAGE_DISPOSITION_INFO(const AMQP_MESSENGER_MESSAGE_DISPOSITION_INFO* left, const AMQP_MESSENGER_MESSAGE_DISPOSITION_INFO* right)
+{
+    int result;
+
+    if (((left == NULL) && (right != NULL)) || ((right == NULL) && (left != NULL)))
```
ewertons: You need one more check in case both are NULL, otherwise you will get a SEGFAULT on the next check when derefencing a NULL pointer.

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
...
+    {
+        result = 0;
+    }
+    else
+    {
+        if ((left->source != right->source) || (left->message_id != right->message_id))
```
ewertons: source is a char* . I'm not sure if this is the correct way to compare these fields.
Could you just add a comment above: // source field might need to be compared by value in the future if additional tests require.

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
...
+        result = -1;
+    }
+    else
+    {
+        destination->source = (char*)TEST_malloc(sizeof(source->source));
+        if (destination->source == NULL)
+        {
+            // failed to malloc
+            result = -1;
+            TEST_free(destination->source);
```
ewertons: If source is NULL, no need to free it.

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
...
+{
+    int result;
+
+    if ((source == NULL) || (destination == NULL))
+    {
+        result = -1;
+    }
+    else
+    {
+        destination->source = (char*)TEST_malloc(sizeof(source->source));
```
ewertons: Nit: C99 spec recommends not casting the value from malloc. Something I learned myself recently from another CR.

ewertons: using sizeof() on source to get its size might be dangerous... it's not guaranteed that tests will pass a const string in there. I recommend using strlen()

ewertons: You should also not touch source->source if it's NULL. Must check if source->source is NULL and if it is, set destination to NULL and return success.

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
```diff
@@ -959,6 +959,79 @@ static AMQP_MESSENGER_HANDLE create_and_start_messenger2(AMQP_MESSENGER_CONFIG*
...
+        if (destination->source == NULL)
+        {
+            // failed to malloc
+            result = -1;
+            TEST_free(destination->source);
+        }
+        else
+        {
+            if (source->source == NULL)
```
ewertons: This whole block makes no sense, and it will fail valgrind with memory leaks because you firt allocated memory above and stored in destination->source, and down here just override destination->source with the pointer value of source->source.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
     return handle;
 }
 
+char* umock_stringify_BINARY_DATA()
```
ewertons: Shouldn't these functions take arguments?

ewertons: The same applies here about what was mentioned for AMQP disposition above. These are not true copies. Have you run this code on linux under valgrind as well to check for memory leaks?

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+{
+    char* result = (char*)TEST_malloc(1);
+    result[0] = '\0';
+    return result;
+}
+
+int umock_are_equal_BINARY_DATA()
+{
```
ewertons: Shouldn't it take argumetns and compare?

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+    //force fall through to success bypassing access violation
+    int result = 1;
+    return result;
+}
+
+int umock_copy_BINARY_DATA()
+{
```
ewertons: Args? compare?

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+    return result;
+}
+
+void umock_free_BINARY_DATA()
+{
+    //do nothing
+}
+
+char* umock_stringify_TELEMETRY_MESSENGER_MESSAGE_DISPOSITION_INFO(const TELEMETRY_MESSENGER_MESSAGE_DISPOSITION_INFO* value)
+{
```
ewertons: Same comments here are above for AMQP disposition.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+    int result;
+
+    if (((left == NULL) && (right != NULL)) || ((right == NULL) && (left != NULL)))
+    {
+        result = 0;
+    }
+    else
+    {
+        if ((left->source != right->source) || (left->message_id != right->message_id))
```
ewertons: Same here. Possible SEGFAULT when both left and right are NULL

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+        result = -1;
+    }
+    else
+    {
+        destination->source = (char*)TEST_malloc(sizeof(source->source));
+        if (destination->source == NULL)
+        {
+            // failed to malloc
+            result = -1;
+            TEST_free(destination->source);
```
ewertons: Not needed.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+        }
+        else
+        {
+            if (source->source == NULL)
```
ewertons: Again this seems to be needed before the previous block.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+            {
+                result = -1;
+            }
+            else
+            {
+                destination->message_id = source->message_id;
+                destination->source = source->source;
```
ewertons: Same as before, this makes no sense. Valgring will mark a memory leak.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+        }
+    }
+
+    return result;
+}
+
+void umock_free_TELEMETRY_MESSENGER_MESSAGE_DISPOSITION_INFO(TELEMETRY_MESSENGER_MESSAGE_DISPOSITION_INFO* value)
+{
+    TEST_free(value);
+    value->source = NULL;
```
ewertons: This is basic memory segfault. Cannot access memory location after calling free() on it.

ewertons: And should not set source to NULL because it could have a dynamically allocated string.

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
```diff
@@ -1665,11 +1665,109 @@ static TELEMETRY_MESSENGER_HANDLE create_and_start_messenger2(TELEMETRY_MESSENGE
...
+    value->message_id = 0;
```
ewertons: No need to set numeric field member after free on structure.

iothub_client/tests/iothubtr_amqp_twin_msgr_ut/iothubtr_amqp_twin_msgr_ut.c
```diff
@@ -713,6 +714,31 @@ static TWIN_MESSENGER_HANDLE create_and_start_twin_messenger(TWIN_MESSENGER_CONF
     return handle;
 }
 
+char* umock_stringify_BINARY_DATA()
```
ewertons: Same comments as above for BINARY_DATA

iothub_client/tests/uamqp_messaging_ut/uamqp_messaging_ut.c
```diff
@@ -516,6 +516,58 @@ static void reset_test_data()
     memset(saved_malloc_returns, 0, sizeof(saved_malloc_returns));
 }
 
+char* umock_stringify_BINARY_DATA()
```
ewertons: Same comments as above for BINARY_DATA

iothub_service_client/tests/iothub_msging_ll_ut/iothub_messaging_ll_ut.c
```diff
@@ -409,12 +411,36 @@ static AMQP_VALUE TEST_AMQP_MAP = ((AMQP_VALUE)0x6258);
 static MAP_HANDLE TEST_MAP_HANDLE = (MAP_HANDLE)0x103;
 static IOTHUB_MESSAGE_HANDLE TEST_IOTHUB_MESSAGE_HANDLE = (IOTHUB_MESSAGE_HANDLE)0x4242;
 
+char* umock_stringify_BINARY_DATA()
```
ewertons: Same comments as above for BINARY_DATA

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
```diff
@@ -127,6 +127,72 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
 
 static TEST_MUTEX_HANDLE g_testByTest;
 
+char* umock_stringify_TPMA_SESSION(const TPMA_SESSION* value)
```
ewertons: Same comments as in for AMQP disposition.

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
```diff
@@ -127,6 +127,72 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+{
+    int result;
+
+    if (((left == NULL) && (right != NULL)) || ((right == NULL) && (left != NULL)))
+    {
+        result = 0;
+    }
+    else
+    {
+        if (left == right)
```
ewertons: So our comparison is by pointer address only (shallow compare)? Is that enough?

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
```diff
@@ -127,6 +127,72 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+{
+    int result;
+
+    if ((source == NULL) || (destination == NULL))
+    {
+        result = -1;
+    }
+    else
+    {
+        destination->continueSession = source->continueSession;
```
ewertons: Shallow copy. This could be dangerous, specially if we free instances separately.

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
```diff
@@ -127,6 +127,72 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+        result = 0;
+    }
+
+    return result;
+}
+
+void umock_free_TPMA_SESSION(TPMA_SESSION* value)
+{
+    // reset values
+    value->continueSession = 0;
```
ewertons: This should not be needed at all. Only if memory is allocated.

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
```diff
@@ -456,6 +456,63 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
 
 static TEST_MUTEX_HANDLE g_testByTest;
 
+char* umock_stringify_BINARY_DATA(const BINARY_DATA* value)
```
ewertons: Same comments as before for BINARY_DATA.

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
```diff
@@ -456,6 +456,63 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+        else
+        {
+            result = 0;
+        }
+    }
+
+    return result;
+}
+
+int umock_copy_BINARY_DATA(BINARY_DATA* destination, const BINARY_DATA* source)
```
ewertons: This one is good! That's what we should have everywhere above.

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
```diff
@@ -456,6 +456,63 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+        result = 0;
+    }
+
+    return result;
+}
+
+void umock_free_BINARY_DATA(BINARY_DATA* value)
+{
+    my_gballoc_free((void*)value->bytes);
+    value->bytes = NULL;
```
ewertons: This is a SEGFAULT waiting to happen. Cannot believe tests passed without a SEGFAULT. If they didn't, this function is not being called.

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
```diff
@@ -456,6 +456,63 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
...
+    value->length = 0;
```
ewertons: This is a SEGFAULT waiting to happen. Cannot believe tests passed without a SEGFAULT. If they didn't, this function is not being called.

NEver touch memory after calling free()

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: This is how another passing UT was set up for BINARY_DATA structure being umock registered, my understanding is when you register a VALUE_TYPE you need a stringify, are_equal, copy, and free, for umock to register custom types like structs as per (https://github.com/Azure/umock-c/blob/master/doc/umock_c.md) so stringify is used to return a string representation of a custom value type, I also concur that an empty string doesn't seem to be correct, however the function is just needed to register the value type for umock, it doesn't change anything else in the UT.

ttins: https://github.com/Azure/azure-iot-sdk-c/blob/f5728ee3ddb39a76890fe1220fe286f557ed3eec/iothub_client/tests/iothubtransportamqp_methods_ut/iothubtransportamqp_methods_ut.c

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: True, I missed that check, if both are NULL we are ok since they are equal to each other, but since I'm checking members below that would SEGFUALT

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: strlen(), of course, will switch that

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: Will investigate with Dan.

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: strcmp like we talked about

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: fixed via our chat

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: removed via our chat

ttins: fixed via our chat

iothub_client/tests/iothubtr_amqp_msgr_ut/iothubtr_amqp_msgr_ut.c
ttins: refactored via our chat

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: Yes, but via our chat, we decided to use the functions as shells for now, I did add the args back in then cast to void, same functionality but at least it looks better

ttins: same as above

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: same as above

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: same as above

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: fixed : mirrored the AMQP Disposition

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: fixed

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: removed

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: fixed

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: fixed

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: fixed, has nothing to free, just voids the arg

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: removed

iothub_client/tests/iothubtr_amqp_twin_msgr_ut/iothubtr_amqp_twin_msgr_ut.c
ttins: see above

iothub_client/tests/uamqp_messaging_ut/uamqp_messaging_ut.c
ttins: see above

iothub_service_client/tests/iothub_msging_ll_ut/iothub_messaging_ll_ut.c
ttins: see above

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
ttins: TPMA_SESSION is struct of unsigned ints, so this should be valid? I tested for shallow copy after the calls and values stay as expected on both source and destination.

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
ttins: No its not enough, fixed

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
ttins: fixed mirrored AMQP disposition

provisioning_client/tests/hsm_client_tpm_ut/hsm_client_tpm_ut.c
ttins: removed all resets

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
ttins: see above

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
ttins: The only reason this one works, is when source has a length, its bytes are not null. This might require some more investigation into how the tests are setup might be the key to the other tests using registered BINARY_DATA

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
ttins: removed, even though not caught in the valgrind test

provisioning_client/tests/prov_transport_amqp_common_ut/prov_transport_amqp_common_ut.c
ttins: removed

iothub_client/tests/iothubtr_amqp_tel_msgr_ut/iothubtr_amqp_tel_msgr_ut.c
ttins: Filled in all shell functions in hash 7f188b9

PR #1469, opened 2020-03-19
Cap the maximum delay between attemps of reconnection (retry policy)

ewertons:
This change caps the maximum delay between attempts of reconnection to thirty seconds.
Such constraint applies to the following
- Retry policies: IOTHUB_CLIENT_RETRY_LINEAR_BACKOFF, IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF, IOTHUB_CLIENT_RETRY_EXPONENTIAL_BACKOFF_WITH_JITTER.
- Transports: AMQP, AMQP/WS, MQTT, MQTT/WS.

This has been a long-time request, and now it aligns with other Azure IoT SDKs (e.g., https://github.com/Azure/azure-iot-sdk-csharp/blob/f65988b6122cd43df32c123d707395d1bb4367b3/iothub/device/src/TransientFaultHandling/RetryStrategy.cs#L43)

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/src/iothub_client_retry_control.c
+++ b/iothub_client/src/iothub_client_retry_control.c
@@ -10,8 +10,9 @@
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/xlogging.h"
 
-#define RESULT_OK           0
-#define INDEFINITE_TIME     ((time_t)-1)
+#define RESULT_OK                 0
+#define INDEFINITE_TIME           ((time_t)-1)
+#define DEFAULT_MAX_DELAY_IN_SECS 30
 
 typedef struct RETRY_CONTROL_INSTANCE_TAG
 {
@@ -20,6 +21,7 @@ typedef struct RETRY_CONTROL_INSTANCE_TAG
 
     unsigned int initial_wait_time_in_secs;
     unsigned int max_jitter_percent;
+    unsigned int max_delay_in_secs;
 
     unsigned int retry_count;
     time_t first_retry_time;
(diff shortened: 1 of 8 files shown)


iothub_client/src/iothub_client_retry_control.c
```diff
@@ -10,8 +10,9 @@
 #include "azure_c_shared_utility/optimize_size.h"
 #include "azure_c_shared_utility/xlogging.h"
 
-#define RESULT_OK           0
-#define INDEFINITE_TIME     ((time_t)-1)
+#define RESULT_OK                 0
+#define INDEFINITE_TIME           ((time_t)-1)
+#define DEFAULT_MAX_DELAY_IN_SECS 30
```
CIPop: C# has this set at 10s by default. See docs here for one example: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md
Just pointing this out, I think 30 or even higher is a good idea if the server or network is busy to save power, etc.

iothub_client/tests/iothub_client_retry_control_ut/iothub_client_retry_control_ut.c
```diff
@@ -925,9 +925,9 @@ TEST_FUNCTION(Should_Retry_EXPONENTIAL_BACKOFF_success)
     unsigned int option_value = 2;
     int set_option_result = retry_control_set_option(handle, RETRY_CONTROL_OPTION_INITIAL_WAIT_TIME_IN_SECS, &option_value);
 
-    int expected_retry_times[] = { 0, 2, 4, 8, 16 };
+    int expected_retry_times[] = { 0, 2, 4, 8, 16, 30, 30 };
```
CIPop: See above: first retry should not be after 0 seconds (that spins the CPU, likely useless if the network is at fault).

iothub_client/tests/iothub_client_retry_control_ut/iothub_client_retry_control_ut.c
```diff
@@ -943,11 +943,11 @@ TEST_FUNCTION(Should_Retry_INTERVAL_success)
     unsigned int max_retry_time_in_secs = 19;
     RETRY_CONTROL_HANDLE handle = create_retry_control(IOTHUB_CLIENT_RETRY_INTERVAL, max_retry_time_in_secs);
 
-    int expected_retry_times[] = { 0, 5, 10, 15, 20 };
+    int expected_retry_times[] = { 0, 5, 10, 15, 20, 25, 30, 30 };
```
CIPop: Same as above: first retry should be after at least 1 second.

iothub_client/src/iothub_client_retry_control.c
ewertons: So I think the csharp documentation is outdated then. The cshap code shows it's using 30 seconds for max delay. Shouldn't that be the case?
I also do agree 30 seconds seems more appropriate, although that is a empirical opinion.

iothub_client/tests/iothub_client_retry_control_ut/iothub_client_retry_control_ut.c
ewertons: I would disagree. 0 seconds was added for a reason. In lab tests we got that by the time the SDK took to detect a disconnection (either by a delay on OS notification or no notification at all), the underlying network available again, but the iothub connection was severed. Retrying at 0 seconds showed to be the most effective on most of wired and wifi connection gaps.

iothub_client/tests/iothub_client_retry_control_ut/iothub_client_retry_control_ut.c
ewertons: Same as above.

iothub_client/src/iothub_client_retry_control.c
ewertons: Follow up item for the minimum delay: https://github.com/Azure/azure-iot-sdk-c/issues/1470

PR #1472, opened 2020-03-19
Buffer and arithmetic overflow fix

ttins:


ttins:
Keep it simple, add check in blob.c for blockID within expected range

ttins:
Possible don't merge until https://github.com/Azure/azure-iot-sdk-c/pull/1471 is merged

PR #1477, opened 2020-03-26
Add NULL dereference checks

ttins:


iothub_client/src/message_queue.c
```diff
@@ -54,7 +54,15 @@ static bool find_item_by_message_ptr(LIST_ITEM_HANDLE list_item, const void* mat
     MESSAGE_QUEUE_ITEM* current_item = (MESSAGE_QUEUE_ITEM*)singlylinkedlist_item_get_value(list_item);
     MQ_MESSAGE_HANDLE* target_item = (MQ_MESSAGE_HANDLE*)match_context;
 
-    return (current_item->message == target_item);
+    if (current_item == NULL)
```
danewalton: the style in this doc (and I think most of this SDK) is to have one return point. So please update to have something like 
```c
bool result;

if()
{
    result = false;
}
else
{
//...
}

return result;
```

serializer/src/codefirst.c
```diff
@@ -560,6 +560,12 @@ EXECUTE_COMMAND_RESULT CodeFirst_InvokeAction(DEVICE_HANDLE deviceHandle, void*
         result = EXECUTE_COMMAND_ERROR;
         LogError("parameterValues error %s ", MU_ENUM_TO_STRING(EXECUTE_COMMAND_RESULT, result));
     }
+    /*Codes_SRS_CODEFIRST_99_200:[ If devicHeader (callbackUserContext) is NULL then EXECUTE_COMMAND_ERROR shall be returned.]*/
```
danewalton: typo: deviceHeader

ttins: whoops good catch

iothub_client/src/message_queue.c
ttins: else:
result = (current_item->message == target_item);

PR #1500, opened 2020-04-09
Move from urn's to dtmi

jspaith:
Interfaces in DTDLv2 are no longer prefixed with `urn:` but instead follow a `dtmi:` pattern.  Furthermore, the separator between the name and the version is now a `;'.

This change:
* Renames `urn` => `dtmi` in samples, documentation, UT, etc.
* Changes the parser to validate `dtmi` and componentNames are valid, following DTDLv2 rules.

Typically we don't validate on client itself and let server fail policy issues.  But a developer getting this wrong creates very unpleasant debugging experience (I've seen this in practice).  

The application will send the invalid data to the server and the server will fail by killing the TCP connection.  The underlying IoTHub SDK will assume this TCP Connection dropping is because of a network issue, not caller's fault, and kick in its retry logic.  The application developer will understandably also think something is wrong with their network or IoTHub and not think to look at how they formatted a string.  

So... we do a local check first to provide meaningful error information in the developer inner loop.


danewalton:
Should this be going into master?

jspaith:
Note I initially had this merging against `master`.  Uhm, not yet...

danewalton:
> 
> 
> Note I initially had this merging against `master`. Uhm, not yet...

Okay yea I just checked again and now it's not into master. I thought I was seeing things. To clarify I saw the original PR going against master and thought it should not be doing that. Looks like you updated.

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -278,65 +287,184 @@ static bool DT_IsAlpha(char c)
...
-// TCP connection being closed.  Because this can be so hard for developer to diagnose interface being
-// invalid versus network issues, we provide this check.
-int DT_InterfaceClient_CheckNameValid(const char* valueToCheck, bool isInterfaceId)
+// IsEndOfSegmentCharacter returns whether the current character is the end of a segment.
+// For component names, this will simply be a 0.  For interfaces, it will be a ';' or ':'.
+static bool IsEndOfSegmentCharacter(char c, DT_SEGMENT_TYPE segmentType)
+{
+    if (segmentType == DT_SEGMENT_TYPE_COMPONENT_NAME)
+    {
+        return (c==0);
```
danewalton, line 296: nit: spacing here
```suggestion
        return (c == 0);
```

ericwol-msft: '\0'

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -278,65 +287,184 @@ static bool DT_IsAlpha(char c)
...
+    }
+    else
+    {
+        return (c == DT_Segment_Separator || c == DT_Version_Separator);
+    }
+}
+
+// IsNameSegmentValid checks whether either a component name or a segment of
+// an interface is legal DTMI.  Legal requires the segment to start with an alphabetical character,
```
ericwol-msft, line 305: Legal requires the segment  => Syntax requires the segment

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -278,65 +287,184 @@ static bool DT_IsAlpha(char c)
...
+// be followed by alpha/numbers/_, but not end with a _.
+static bool IsNameSegmentValid(const char* segmentName, DT_SEGMENT_TYPE segmentType, const char** endPosition)
```
ericwol-msft, line 307: This can't be done in a regex expression?

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -278,65 +287,184 @@ static bool DT_IsAlpha(char c)
...
-    int result = 0;
-    const char* current = valueToCheck;
+    int result;
+    const char* endPosition;
 
-    if (valueToCheck == NULL)
+    if (componentName == NULL)
     {
-        LogError("Invalid parameter.  valueToCheck=%p", valueToCheck);
+        LogError("Component is NULL");
```
ericwol-msft, line 360: componentName is NULL

digitaltwin_client/src/dt_interface_client.c
```diff
@@ -278,65 +287,184 @@ static bool DT_IsAlpha(char c)
...
         result = MU_FAILURE;
     }
-    else if (strlen(valueToCheck) > DT_InterfaceMaxLength)
+    else if (strlen(componentName) > DT_ComponentNameMaxLength)
     {
-        LogError("Name %s is too long, maximum length is %lu", valueToCheck, (unsigned long)DT_InterfaceMaxLength);
+        LogError("Component %s is too long, must be under %lu characters", componentName, (unsigned long)DT_ComponentNameMaxLength);
```
ericwol-msft, line 365: componentName

digitaltwin_client/src/dt_interface_client.c
jspaith: Fixed both spacing and '\0'.

digitaltwin_client/src/dt_interface_client.c
jspaith: It can, but the formal definition of this is quite long.  This is one I thought the English (and of course code) was easier to follow.

For languages with regex parsers this would presumably how they'd chech string and logic below would be in code and self-documenting.

```
dtmi =       "dtmi" ":" dtpath ";" dtver
dtpath =     dtseg *(":" dtseg)
dtseg =      dtuserseg / dssysseg
dtuserseg =  ALPHA [ *(ALPHA / DIGIT / "_") (ALPHA / DIGIT) ]
dtsysseg =   "_" *(ALPHA / DIGIT / "_") (ALPHA / DIGIT)
dtver =      digitnz 0*8DIGIT
digitnz =    %x31-39
             ; 1-9
```

digitaltwin_client/src/dt_interface_client.c
jspaith: Fixing.

digitaltwin_client/src/dt_interface_client.c
jspaith: Yup, fixing.

digitaltwin_client/src/dt_interface_client.c
jspaith: Wow, it's like Yoda wrote the original.  Fixing.

PR #1526, opened 2020-05-06
Reconcile C SDK naming and samples with latest

jspaith:
Getting C SDK into alignment with others and broader PnP around naming and samples.
* Most importantly is renaming the `root_interface_id` variations to the agreed on standard `modelID` (with casing appropriate per context in language, of course).
* A missing command should result in a 404, not a 5xx.
* Fix the sample device's model so that it contains SDKInformation and DeviceInformation.
* Use official DeviceInformation interface name and the convention of component names.
* Remove asynchronous commands samples (only) from DTDL and the EnvironmentalSensor.
* `my_company:com` as standard instead of `YOUR_COMPANY` all over the place.


jspaith:
Big shoutout to @olivakar who did A TON of investigation to help find many of these problems.

PR #1527, opened 2020-05-11
Incorporate latest DTDL changes

jspaith:
There has been additional late breaking requests to change the sample DTDL files, mostly around naming convention (prefer `example:com` to `examples:com`) and various Pascal and Camel casing conventions.

The PR we're inspired by is [this](https://github.com/Azure/IoTPlugandPlay/pull/54), which has been signed off on and merged.

### DIFF
--- a/digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
+++ b/digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
@@ -30,7 +30,7 @@ DIGITALTWIN_SAMPLE_DEVICEINFO_STATE digitaltwinSample_DeviceInfoState;
 
 // DigitalTwin interface name from service perspective.
 static const char digitaltwinSampleDeviceInfo_InterfaceId[] = "dtmi:azure:DeviceManagement:DeviceInformation;1";
-static const char digitaltwinSampleDeviceInfo_InterfaceName[] = "Deviceinformation";
+static const char digitaltwinSampleDeviceInfo_InterfaceName[] = "deviceInformation";
 
 // DigitalTwinSampleDeviceInfo_PropertyCallback is invoked when a property is updated (or failed) going to server.
 // In this sample, we route ALL property callbacks to this function and just have the userContextCallback set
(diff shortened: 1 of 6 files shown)


digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
```diff
@@ -30,7 +30,7 @@ DIGITALTWIN_SAMPLE_DEVICEINFO_STATE digitaltwinSample_DeviceInfoState;
 
 // DigitalTwin interface name from service perspective.
 static const char digitaltwinSampleDeviceInfo_InterfaceId[] = "dtmi:azure:DeviceManagement:DeviceInformation;1";
```
rido-min, line 32: Not from this PR, but.. why we need the interface id? Looks like it's only used for logging.

jspaith: #1 - RE interfaceID, very good eyes.  I've been monitoring this too though I should've created a GitHub issue (please @rido-min go ahead and create one now if you want or lmk and I'll create it).  Agree it should go away; only reason to keep would... not interesting anymore.  I do need to talk to other SDKs to get them inline though.  Post bug bash, pre-GA activity, and we'll be reconciling individual API at a time per SDK now that we generally have right shape.

#2 - RE the copy of files.  I really don't like it either and it's another pre-GA thing I will fix.  (Another good GitHub issue against C SDK either of us can create.)  Why do we have it like this today??

Because - these files are SO much in flux in the sample DTDL (as we've seen) that a customer who cloned to an older version of the repo, say August 2019, if they needed to rely on a link could end up following link to DTDL that no longer matched their older .c files.

Fix here is that once our overall PnP Repo strategy is finalized (it's changing I believe?, though I'm not tracking) the fix is that whenever we update version of enviromental sensor from (n) to (n+1), we just create a new file but leave the (n) interface around for legacy.

rido-min: @jspaith,  for 1) I just created #1528 

For 2), I understand copying the files with the samples, however I believe there is room to better organize the models  eg: in a new folder `models` with a README pointing to the DTDL repo.

jspaith: On (2) I've opened https://github.com/Azure/azure-iot-sdk-c/issues/1529 to track end-fix.  I'd like to just focus on the end state of pointing to proper URls when time comes.  In meantime I'm going to get this code checked in for bash.

PR #1535, opened 2020-05-13
Update componentName of sdkInfo

rido-min:
@jspaith 
To be compatible with https://github.com/Azure/IoTPlugandPlay/blob/master/samples/SampleDevice.model.json#L19 

Also, I believe the component should receive the instanceName from the device, instead of setting the name in the implementation

 

# Checklist

# Description of the problem

# Description of the solution


jspaith:
Thanks @rido-min for catching, about to merge.

RE "Also, I believe the component should receive the instanceName from the device, instead of setting the name in the implementation".

In general I agree.  sdkInformation and deviceInformation are special cases though.  We WANT those to always have the same component name, as much as possible, to simplify common case.  I get we can't enforce but enough people will copy our code that it'll be a more or less standard.

PR #1539, opened 2020-05-18
Update API version & new test for twin array support

jspaith:
IoTHub has added support for JSON arrays inside device twins.  To enable this, we need to up our API version string.  

Also extending e2e twin testing to check that arrays are handled correctly.  Even though it's all opaque to C SDK not that much extra testing.


jspaith:
Note that I'm NOT checking this in as the feature still has a few places to be enabled.  Should be shortly.

Issue #1547, opened 2020-05-27
Remove preview API version from C SDK


jspaith:
We have added logic into the C SDK master branch so that it has hard-coded knowledge about a preview API version.  This is NOT the default API, which will continue to be an officially supported API.  But some code will trigger this.

Once PnP goes to GA, we need to remove this logic.  See `IOTHUB_API_PREVIEW_VERSION` in the file `iothub_internal_consts.h`.  This should be removed and **all** logic should just the official `IOTHUB_API_VERSION` (which will itself point to a newer API version that has been deployed on the hub).

jspaith:
Even though code is going into our master branch, I'm tagging with `public-preview` because we're blocked from removing this into the `public-preview-pnp` work is ready to light up.

PR #1552, opened 2020-06-04
Add ModelID parameter to IoTHub Client (MQTT only)

jspaith:
In order to support applications building PnP / DigitalTwin clients on the `master` IoTHub API, we need to let them specify their modelId for the MQTT CONNECT packet, which lives in the userName field.  

This is done via a `DeviceClient_SetOption(handle, OPTION_MODEL_ID, "dtmi:...");`.  It must be specified prior to the first CONNECT, since trying to add this or allow it to be changed later will considerably complicate the state management.  The only time a modelId will change on a given device/module identity in practice is during firmware updates in any case, which implies a new handle in any case.

The modelID parsing feature requires an API version which is still in preview and not deployed worldwide.  To account for this, we will use a preview API version if modelId has been specified and otherwise use the existing model ID.  This API version split is not long-term desirable; #1547 tracks fixing.

Note that this change is MQTT(_WS) only, per decision to only support C device SDK with PnP over MQTT(_WS).

### DIFF
--- a/iothub_client/inc/internal/iothub_internal_consts.h
+++ b/iothub_client/inc/internal/iothub_internal_consts.h
@@ -10,6 +10,9 @@ extern "C"
 #endif
 
     static const char* IOTHUB_API_VERSION = "2017-11-08-preview";
+    // TODO: https://github.com/Azure/azure-iot-sdk-c/issues/1547 tracks removing this preview 
+    // variable once the underlying logic is enabled on all IoTHubs.
+    static const char* IOTHUB_API_PREVIEW_VERSION = "2020-05-31-preview";
 
     static const char* SECURITY_INTERFACE_INTERNAL_ID = "iothub-interface-internal-id";
     static const char* SECURITY_INTERFACE_INTERNAL_ID_VALUE = "security*azureiot*com^SecurityAgent^1*0*0";
--- a/iothub_client/inc/internal/iothub_transport_ll_private.h
+++ b/iothub_client/inc/internal/iothub_transport_ll_private.h
@@ -38,7 +38,7 @@ extern "C"
     typedef void (*pfTransport_Twin_ReportedStateComplete_Callback)(uint32_t item_id, int status_code, void* ctx);
     typedef void (*pfTransport_Twin_RetrievePropertyComplete_Callback)(DEVICE_TWIN_UPDATE_STATE update_state, const unsigned char* payLoad, size_t size, void* ctx);
     typedef int (*pfTransport_DeviceMethod_Complete_Callback)(const char* method_name, const unsigned char* payLoad, size_t size, METHOD_HANDLE response_id, void* ctx);
-    typedef const char* (*pfTransport_GetOption_DT_MODEL_ID_Callback)(void* ctx);
+    typedef const char* (*pfTransport_GetOption_MODEL_ID_Callback)(void* ctx);
 
     /** @brief    This struct captures device configuration. */
     typedef struct IOTHUB_DEVICE_CONFIG_TAG
@@ -68,7 +68,7 @@ extern "C"
         pfTransport_Twin_ReportedStateComplete_Callback twin_rpt_state_complete_cb;
         pfTransport_Twin_RetrievePropertyComplete_Callback twin_retrieve_prop_complete_cb;
         pfTransport_DeviceMethod_Complete_Callback method_complete_cb;
-        pfTransport_GetOption_DT_MODEL_ID_Callback dt_get_model_id_cb;
+        pfTransport_GetOption_MODEL_ID_Callback get_model_id_cb;
     } TRANSPORT_CALLBACKS_INFO;
 
     typedef STRING_HANDLE (*pfIoTHubTransport_GetHostname)(TRANSPORT_LL_HANDLE handle);
--- a/iothub_client/inc/internal/iothub_transport_ll_private.h
+++ b/iothub_client/inc/internal/iothub_transport_ll_private.h
@@ -38,7 +38,7 @@ extern "C"
     typedef void (*pfTransport_Twin_ReportedStateComplete_Callback)(uint32_t item_id, int status_code, void* ctx);
     typedef void (*pfTransport_Twin_RetrievePropertyComplete_Callback)(DEVICE_TWIN_UPDATE_STATE update_state, const unsigned char* payLoad, size_t size, void* ctx);
     typedef int (*pfTransport_DeviceMethod_Complete_Callback)(const char* method_name, const unsigned char* payLoad, size_t size, METHOD_HANDLE response_id, void* ctx);
-    typedef const char* (*pfTransport_GetOption_MODEL_ID_Callback)(void* ctx);
+    typedef const char* (*pfTransport_GetOption_Model_Id_Callback)(void* ctx);
 
     /** @brief    This struct captures device configuration. */
     typedef struct IOTHUB_DEVICE_CONFIG_TAG
@@ -68,7 +68,7 @@ extern "C"
         pfTransport_Twin_ReportedStateComplete_Callback twin_rpt_state_complete_cb;
         pfTransport_Twin_RetrievePropertyComplete_Callback twin_retrieve_prop_complete_cb;
         pfTransport_DeviceMethod_Complete_Callback method_complete_cb;
-        pfTransport_GetOption_MODEL_ID_Callback get_model_id_cb;
+        pfTransport_GetOption_Model_Id_Callback get_model_id_cb;
     } TRANSPORT_CALLBACKS_INFO;
 
     typedef STRING_HANDLE (*pfIoTHubTransport_GetHostname)(TRANSPORT_LL_HANDLE handle);
--- a/iothub_client/inc/internal/iothub_transport_ll_private.h
+++ b/iothub_client/inc/internal/iothub_transport_ll_private.h
@@ -38,6 +38,7 @@ extern "C"
     typedef void (*pfTransport_Twin_ReportedStateComplete_Callback)(uint32_t item_id, int status_code, void* ctx);
     typedef void (*pfTransport_Twin_RetrievePropertyComplete_Callback)(DEVICE_TWIN_UPDATE_STATE update_state, const unsigned char* payLoad, size_t size, void* ctx);
     typedef int (*pfTransport_DeviceMethod_Complete_Callback)(const char* method_name, const unsigned char* payLoad, size_t size, METHOD_HANDLE response_id, void* ctx);
+    typedef const char* (*pfTransport_GetOption_DT_MODEL_ID_Callback)(void* ctx);
 
     /** @brief    This struct captures device configuration. */
     typedef struct IOTHUB_DEVICE_CONFIG_TAG
@@ -67,6 +68,7 @@ extern "C"
         pfTransport_Twin_ReportedStateComplete_Callback twin_rpt_state_complete_cb;
         pfTransport_Twin_RetrievePropertyComplete_Callback twin_retrieve_prop_complete_cb;
         pfTransport_DeviceMethod_Complete_Callback method_complete_cb;
+        pfTransport_GetOption_DT_MODEL_ID_Callback dt_get_model_id_cb;
     } TRANSPORT_CALLBACKS_INFO;
 
     typedef STRING_HANDLE (*pfIoTHubTransport_GetHostname)(TRANSPORT_LL_HANDLE handle);
--- a/iothub_client/src/iothub_client_core_ll.c
+++ b/iothub_client/src/iothub_client_core_ll.c
@@ -128,7 +128,7 @@ typedef struct IOTHUB_CLIENT_CORE_LL_HANDLE_DATA_TAG
     STRING_HANDLE product_info;
     IOTHUB_DIAGNOSTIC_SETTING_DATA diagnostic_setting;
     SINGLYLINKEDLIST_HANDLE event_callbacks;  // List of IOTHUB_EVENT_CALLBACK's
-    STRING_HANDLE dt_model_id;
+    STRING_HANDLE model_id;
 }IOTHUB_CLIENT_CORE_LL_HANDLE_DATA;
 
 static const char HOSTNAME_TOKEN[] = "HostName";
@@ -626,7 +626,7 @@ static const char* IoTHubClientCore_LL_GetModelID(void* ctx)
     else
     {
         IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* iothub_data = (IOTHUB_CLIENT_CORE_LL_HANDLE_DATA*)ctx;
-        result = STRING_c_str(iothub_data->dt_model_id);
+        result = STRING_c_str(iothub_data->model_id);
     }
     return result;
 }
@@ -844,7 +844,7 @@ static IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* initialize_iothub_client(const IOTHUB_
             transport_cb.msg_input_cb = IoTHubClientCore_LL_MessageCallbackFromInput;
             transport_cb.msg_cb = IoTHubClientCore_LL_MessageCallback;
             transport_cb.method_complete_cb = IoTHubClientCore_LL_DeviceMethodComplete;
-            transport_cb.dt_get_model_id_cb = IoTHubClientCore_LL_GetModelID;
+            transport_cb.get_model_id_cb = IoTHubClientCore_LL_GetModelID;
 
             if (client_config != NULL)
             {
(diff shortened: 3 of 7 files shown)


iothub_client/inc/internal/iothub_transport_ll_private.h
```diff
@@ -38,6 +38,7 @@ extern "C"
     typedef void (*pfTransport_Twin_ReportedStateComplete_Callback)(uint32_t item_id, int status_code, void* ctx);
     typedef void (*pfTransport_Twin_RetrievePropertyComplete_Callback)(DEVICE_TWIN_UPDATE_STATE update_state, const unsigned char* payLoad, size_t size, void* ctx);
     typedef int (*pfTransport_DeviceMethod_Complete_Callback)(const char* method_name, const unsigned char* payLoad, size_t size, METHOD_HANDLE response_id, void* ctx);
+    typedef const char* (*pfTransport_GetOption_MODEL_ID_Callback)(void* ctx);
```
danewalton, line 41: nit: any reason why `_MODEL_ID_` is all caps?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2172,6 +2178,80 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+                result = 0;
+            }
+        }
+        else
+        {
+            result = 0;
+        }
+
+        // setting connect string is only allowed once in the lifetime of the device client.
+        transport_data->isConnectUsernameSet = true;
```
danewalton, line 2240: This is set to true even in the case that an above call failed. Expected behavior?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2432,15 +2477,17 @@ static int InitializeConnection(PMQTTTRANSPORT_HANDLE_DATA transport_data)
     return result;
 }
 
-static STRING_HANDLE buildConfigForUsername(const IOTHUB_CLIENT_CONFIG* upperConfig, const char* moduleId)
+// At handle creation time, we don't have all the fields required for building up the user name (e.g. productID)
+// Build what we can (instead of making separate copies the passed fields) for now.
```
danewalton, line 2481:
```suggestion
// Build what we can (instead of making separate copies of the passed fields) for now.
```

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2528,7 +2575,7 @@ static PMQTTTRANSPORT_HANDLE_DATA InitializeTransportHandleData(const IOTHUB_CLI
                 }
                 else
                 {
-                    /* Codes_SRS_IOTHUB_MQTT_TRANSPORT_07_008: [If the upperConfig contains a valid protocolGatewayHostName value the this shall be used for the hostname, otherwise the hostname shall be constructed using the iothubname and iothubSuffix.] */
+                    /* Codes_SRS_IOTHUB_MQTT_TRANSPORT_07_008: [If the upperConfig contains a valid protocolGatewayHostName value this shall be used for the hostname, otherwise the hostname shall be constructed using the iothubname and iothubSuffix.] */
```
danewalton, line 2578: nit
```suggestion
                    /* Codes_SRS_IOTHUB_MQTT_TRANSPORT_07_008: [If the upperConfig contains a valid protocolGatewayHostName value this shall be used for the hostname, otherwise the hostname shall be constructed using the iothubName and iothubSuffix.] */
```

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -150,6 +151,12 @@ static const char* my_Transport_GetOption_Product_Info_Callback(void* ctx)
     return "product_info";
 }
 
+static const char* my_Transport_GetOption_MODEL_ID_Callback(void* ctx)
+{
+    (void)ctx;
+    return "dtmi:testDeviceCapabilityModel;1";
```
danewalton, line 157: this has a `;` and the other examples in the above file use a `:`

Intended?
Also ignorance here. Why is this prefixed with `dtmi` and the others `urn`

iothub_client/inc/internal/iothub_transport_ll_private.h
jspaith: Fixing.

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: Fixed.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
jspaith: The `urn` is a legacy thing that slipped.  DTDLv1 used urn and DTDLv2 uses dtmi.   *Long* story!!

I'm updating though so we're consistent and on dtmi.

iothub_client/src/iothubtransport_mqtt_common.c
danewalton: haha trust me if my dad saw half the stuff I write he'd be shaking his head for sure

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: Yup, I knew that one going in.  The issue why I initially had it like this is we're modifying `transport_data->configPassedThroughUsername` in middle of this function and where later stuff can fail.  This means we can't possibly guess its state if we come through the `step2` function again.

The fix is to do everything in temporary storage, and then only set `configPassedThroughUsername` if everything is successful (and only set `isConnectUsernameSet=true` after).  Even though it's massively unlikely we'd recover if we hit an out of memory at this stage, more I think about it the less I like this.  I'm going to fixup esp. as amount of extra memory isn't that large compared to TLS setup & we delete inside this function itself.

iothub_client/inc/internal/iothub_internal_consts.h
```diff
@@ -10,6 +10,9 @@ extern "C"
 #endif
 
     static const char* IOTHUB_API_VERSION = "2017-11-08-preview";
```
ericwol-msft, line 12: Ugh...why is this still preview?

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -615,6 +615,22 @@ static const char* IoTHubClientCore_LL_GetProductInfo(void* ctx)
     return result;
 }
 
+static const char* IoTHubClientCore_LL_GetModelID(void* ctx)
```
ericwol-msft, line 618: hummm IoTHubClientCore_LL_GetModelID or IoTHubClientCore_LL_GetModelId ?

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -2319,6 +2337,23 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_LL_SetOption(IOTHUB_CLIENT_CORE_LL_HANDLE
...
+        else if (strcmp(optionName, OPTION_MODEL_ID) == 0)
+        {
+            if (handleData->model_id != NULL)
+            {
+                LogError("DT ModelId already specified.");
+                result = IOTHUB_CLIENT_ERROR;
+            } 
+            else if ((handleData->model_id = STRING_construct((const char*)value)) == NULL)
+            {
+                LogError("STRING_c_str failed");
```
ericwol-msft, line 2349: STRING_construct failed

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -236,7 +240,7 @@ typedef struct MQTTTRANSPORT_HANDLE_DATA_TAG
     int http_proxy_port;
     char* http_proxy_username;
     char* http_proxy_password;
-    bool isProductInfoSet;
+    bool isConnectUsernameSet;
```
ericwol-msft, line 243: Is azure sphere going to complain about this?

iothub_client/inc/internal/iothub_internal_consts.h
jspaith: Ugh indeed.  We're having side discussions RE getting latest updates from hub better.  I'll get going again.

This particular one I was planning on updating via #1539, the json array in twin one.  Since it's in progress already and tight deadline on this change, I'd wanted to not touch it here.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2172,6 +2178,80 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
     }
 }
 
+static int buildConfigForUsernameStep2IfNeeded(PMQTTTRANSPORT_HANDLE_DATA transport_data)
```
ericwol-msft, line 2181: buildConfigForUsernameStep2IfNeeded  -> buildConfigForUsernameForModelId

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: ;tl-dr - NO

Since you asked...

We still will send productInfo if set by application.  See `buildConfigForUsernameStep2IfNeeded`.  I changed the name of the variable because it's now not just productInfoSet that matters, but generally adding to userName (product & modelID).

There is a *slight* behavior change here.  Before if the app did below:

```c
IoThub_CreateHandleFromConnectionString();
IoTHub_SendTelemetry(h);
// wait for a bunch, at least long enough for CONNECT to go through
// And after time passed
IoTHub_SetOption("product_info", p);
```

Then if there were ever a disconnection and a successful re-connection, then the product_info would be appending to the subsequent CONNECT packet.

In the new world, this product_info would not be appending to subsequent CONNECT packet because we only build up userName once.  Since an app doesn't control reconnections and since setting product info so late would be odd, I don't see this change as being problematic.  Trying to maintain old behavior is going to really complicate things also would demand specific tests for permutations of the app I called out above to make sure productId was set (or not).

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: It's more than just ModelId.  We also put the API version on here, the productInfo (if app set, or alternatly our default "C-SDK-version-blah/OS-version", ModelId.

I'm fine with you not liking `buildConfigForUsernameStep2IfNeeded ` :).  Let me know if you have a better suggestion that captures the various stuff set here.

iothub_client/src/iothub_client_core_ll.c
jspaith: Fixed.

iothub_client/tests/iothubclientcore_ll_ut/iothub_client_core_ll_ut.c
```diff
@@ -5536,15 +5536,15 @@ TEST_FUNCTION(IoTHubClientCore_LL_SetOption_product_info_fails_case1)
...
+TEST_FUNCTION(IoTHubClientCore_LL_SetOption_model_id_succeeds)
 {
     //arrange
     IOTHUB_CLIENT_CORE_LL_HANDLE h = IoTHubClientCore_LL_Create(&TEST_CONFIG);
     umock_c_reset_all_calls();
     STRICT_EXPECTED_CALL(STRING_construct(IGNORED_PTR_ARG));
 
     //act
-    IOTHUB_CLIENT_RESULT result = IoTHubClientCore_LL_SetOption(h, OPTION_MODEL_ID, "urn:YOUR_COMPANY_NAME_HERE:sample_device:1");
+    IOTHUB_CLIENT_RESULT result = IoTHubClientCore_LL_SetOption(h, OPTION_MODEL_ID, "dtmi:YOUR_COMPANY_NAME_HERE:sample_device:1");
```
danewalton, line 5547: is this supposed to be 
```suggestion
    IOTHUB_CLIENT_RESULT result = IoTHubClientCore_LL_SetOption(h, OPTION_MODEL_ID, "dtmi:YOUR_COMPANY_NAME_HERE:sample_device;1");
```

jspaith: Yup.  Thanks for carefully reviewing UTs.  Fixing.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2172,6 +2178,81 @@ static STRING_HANDLE buildClientId(const char* device_id, const char* module_id)
...
+        else
+        {
+            result = 0;
+        }
+
+        if (result == 0)
+        {
+            STRING_delete(transport_data->configPassedThroughUsername);
+            transport_data->configPassedThroughUsername = userName;
+            userName = NULL;
```
danewalton, line 2238: doesn't this need to be `STRING_delete`'d first?

jspaith: Could you exlpain what problem is more specifically?  I need the `configPassedThroughUsername` when building up `userName` but then I'm done with it, so it gets deleted.

danewalton: I guess it looks a little odd that it is set to NULL here and then on line 2243 it gets deleted, although it was explicitly set to NULL.

danewalton: If the thing that it points to should only be deleted in the failure case I guess that makes sense.

jspaith: What this is doing, in C++ terms, is a [release](http://www.cplusplus.com/reference/memory/unique_ptr/release/) and then the userName destructor gets run no matter what.   Except it takes more lines of code to express that in C99, like everything.  And you're right in failure case it would clean up.

It's cleaner code in the final cleanup to delete anything possibly used in a final end block, and this special case release() thingy here.

This was ultimately your doing for complaining about the one-and-done nature of setting this in the error happened in 1st place :).

danewalton: ahh gotcha okay makes sense

ewertons:
This is scary that the merge button is enabled if Windows64 build fails. @ericwol-msft shouldn't this build be mandatory?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2432,15 +2477,17 @@ static int InitializeConnection(PMQTTTRANSPORT_HANDLE_DATA transport_data)
     return result;
 }
 
-static STRING_HANDLE buildConfigForUsername(const IOTHUB_CLIENT_CONFIG* upperConfig, const char* moduleId)
+// At handle creation time, we don't have all the fields required for building up the user name (e.g. productID)
+// Build what we can (instead of making separate copies the passed fields) for now.
+static STRING_HANDLE buildConfigForUsernameStep1(const IOTHUB_CLIENT_CONFIG* upperConfig, const char* moduleId)
```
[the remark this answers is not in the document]
jspaith: Done.

iothub_client/src/iothubtransport_mqtt_common.c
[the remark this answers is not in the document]
jspaith: Yeah, you had to bring that up didn't you:)?

Anyway I had same idea and it turns out that this requirement is not referenced anywhere else.  There is one with a matching code, but it's a different value.  (That one is at least consistent throughout.)

```
IoTHubTransportMqtt_GetSendStatus shall get the send status by calling into the IoTHubMqttAbstract_GetSendStatus function
```

Since we're not doing these reqs for anything than legacy, I didn't feel inclined to add a new req for this one off.  I'd almost rather delete this one than do that.

ericwol-msft:
> This is scary that the merge button is enabled if Windows64 build fails. @ericwol-msft shouldn't this build be mandatory?

fixed

eliavgnessin:
Thanks. Any plans adding PnP HTTP support? Or we should not expect it?

jspaith:
@eliavgnessin - no, we currently don't have plans for PnP over HTTP.  The primary reason is PnP relies needs DeviceMethods and DeviceTwins from IoTHub layer and these are not supported on HTTP.  

If you have a scenario that you think HTTP is a hard requirement please let us know.  

If it's for a size constrained device, you should in general be checking out https://github.com/Azure/azure-sdk-for-c where we're doing an (in progress) embedded 1st SDK in C.  MQTT based.  We'll eventually have a PnP layer on it.  

We will however merge the `public-preview-pnp` branch on this repo into `master` so you can start any early dev on this branch if you want (actually `public-preview` is the older, more stable branch you should be on).  I will make breaking API changes between now and merging to `master`, but the API will be recognizable.

Issue #1561, opened 2020-06-15
Need consistency on modelId and componentId checking


jspaith:
Currently the SDK has logic that checks whether a componentName is legal.  See `DT_InterfaceClient_CheckComponentNameValid`.

However, the modelID - which is the only bit sent across the wire - is not currently checked.  It easily could be by wiring it up to `DT_InterfaceClient_CheckInterfaceIdValid`.

It doesn't make sense to validate componentName and not modelId.  ModelID being wrong might (open discussion) be required.

So we need to either:
* Do no checking at all.
* Check modelID for valid.
* Check componentName AND modelId.

Our exsiting check componentName BUT NOT modelID is worst answer on this matrix.

jspaith:
Note additionally if we don't parse modelId, we need to remove `DT_InterfaceClient_CheckInterfaceIdValid` and its helpers.  Not checking component name lets us remove `DT_InterfaceClient_CheckComponentNameValid` and helpers.

jspaith:
Closing for now as `public-preview-pnp` isn't currently planned release mechanism.  #pnp-archive

Issue #1528, opened 2020-05-12
PnP components should not require the interface id


rido-min:
.. from https://github.com/Azure/azure-iot-sdk-c/pull/1527#discussion_r423374428

With the latest PnP discovery protocol, the interface id of the PnP component should not be included in the component implementation.

https://github.com/Azure/azure-iot-sdk-c/blob/07dc45d067f482c838446e290e35ec99b7c59336/digitaltwin_client/src/dt_interface_client.c#L559

The same rule should be applied to other languages

/cc @jspaith @jebrando

jspaith:
Closing for now as `public-preview-pnp` isn't currently planned release mechanism.  #pnp-archive

az-iot-builder-01:
@rido-min, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #1529, opened 2020-05-12
Point PnP interface sample DTDL to agreed upon


jspaith:
Currently in PnP, we have sample DTDLs for the sample device & environmental sensor.

There will eventually be a shared location for these samples, along with proper versioning of the files (e.g. enviorment-sensor1.json, enviromental-sensor2.json).  Currently this is in https://github.com/Azure/IoTPlugandPlay/tree/master/samples but it's subject to change and needs the versioning rules mentioned.

jspaith:
Closing for now as `public-preview-pnp` isn't currently planned release mechanism.  #pnp-archive

PR #1562, opened 2020-06-15
Remove interfaceID from PnP API client API & helpers

jspaith:
Per #1528, we have interfaceId in samples/test/API shape but this is NOT needed by the protocol between the device and hub anymore and was effectively a giant no-op.  Shout out to @rido-min for capturing this.

This change:
* Removes `interfaceId` from the API shape and most of the logic in core code dealing with it.
* Addresses an issue where we used the words "interfaceName" quite a bit.  What we really meant was component or componentName.  Fixed up.

NOTE::
There will be some followup work after this change.  In particular see #1561; we're not consistent pre-checking modelId and componentName.  If we decide to pre-check nothing (I think in new protocol design this is much more palatable, following up) then #1561 would also mean we'd remove `DT_InterfaceClient_CheckInterfaceIdValid`.  I don't want to remove this one yet, because if we do want to pre-parse modelID we'd just use `DT_InterfaceClient_CheckInterfaceIdValid` and I'm not yanking till we're sure either way.  The GitHub issue will act as our IOU to do this if need-be.


digitaltwin_client/samples/digitaltwin_sample_environmental_sensor/digitaltwin_sample_environmental_sensor.c
```diff
@@ -1,11 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-// Implements a sample DigitalTwin interface that integrates with an environmental sensor (for reporting 
-// temperature and humidity over time).  It also has basic commands and properties it can process,
-// such as setting brightness of a light and blinking LEDs.  Because this sample is designed
-// to be run anywhere, all of the sameple data and command processing is expressed simply with 
-// random numbers and LogInfo() calls.
+// Implements a sample DigitalTwin interface as defined by "dtmi:my_company:com:EnvironmentalSensor;1"
```
danewalton, line 4: does the user know where to find this `dtmi:my_company:com:EnvironmentalSensor;1`?
at least the json document in this directory has a value named `dtmi:com:example:EnvironmentalSensor;1`

digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-// Implements a sample DeviceInfo DigitalTwin interface.  As the name implies, this interface returns
-// information about the device such as the OS version, amount of storage, etc.
-//
+// Implements a sample DeviceInfo DigitalTwin interface of dtmi:azure:DeviceManagement:DeviceInformation;1.
```
danewalton, line 4: Is the JSON representing `dtmi:azure:DeviceManagement:DeviceInformation;1` missing from this directory?

digitaltwin_client/inc/internal/dt_interface_private.h
```diff
@@ -33,11 +33,9 @@ typedef enum DT_COMMAND_PROCESSOR_RESULT_TAG
 #define DT_COMMAND_ERROR_STATUS_CODE  500
 
 MOCKABLE_FUNCTION(, int, DT_InterfaceClient_CheckComponentNameValid, const char*, componentName);
-MOCKABLE_FUNCTION(, int, DT_InterfaceClient_CheckInterfaceIdValid, const char*, interfaceName);
+MOCKABLE_FUNCTION(, int, DT_InterfaceClient_CheckInterfaceIdValid, const char*, interfaceId);
```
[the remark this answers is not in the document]
jspaith: `DT_InterfaceClient_CheckInterfaceIdValid` at some point *might* be used to check the `modelId` parameter passed in by `DigitalTwin_DeviceClient_CreateFromDeviceHandle`.  This is GitHub issue I reference in comments.

If we don't do this checking after all, I'll remove this function.  If we need it, then I'll wire function up.  I didn't want to remove this function yet only to have to add it back, though.

digitaltwin_client/samples/digitaltwin_sample_environmental_sensor/digitaltwin_sample_environmental_sensor.c
jspaith: Yeah - that mismatch is no good.  I'm updating .c to reflect the JSON in dir.  Ultimately this JSON will be in a central spot all SDKs can point to (#1529 tracks) so the long-term documenting of this only will be solved once we get a spot to point at but not until then.

digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
jspaith: Yes.  Per #1529 the DeviceInfo needs a permanent spot we can link to.  Note that we checked the sample JSON into our C SDK because its samples, whereas DeviceInfo is a "system" interface we didn't want to checkin.  I get its inconsistent which is why we have the IOU bug.

digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
```diff
@@ -1,9 +1,9 @@
...
+// As the name implies, this interface returns information about the device such as the OS version, amount of storage, etc.
```
ericwol-msft, line 5: too wordy

```suggestion
// This interface returns information about the device such as the OS version, amount of storage, etc.
```

digitaltwin_client/samples/digitaltwin_sample_environmental_sensor/digitaltwin_sample_environmental_sensor.c
```diff
@@ -1,11 +1,10 @@
...
+// This demonstrates the core DigitalTwin concepts of telemetry, commands, and properties.
```
ericwol-msft, line 5:
```suggestion
// This demonstrates the core Azure Digital Twin concepts of telemetry, commands, and properties.
```

jspaith: Fixing.

digitaltwin_client/samples/digitaltwin_sample_device_info/digitaltwin_sample_device_info.c
jspaith: As the code I shall soon checkin implies, I have fixed this.

PR #1581, opened 2020-07-07
Adding TPMv2 Software Stack documentation

### LINKED ISSUE #1488: Multiple concurrent TPM commands cause device client to fail to connect.
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

Ubuntu 18.04 using gcc 7.5.0-3ubuntu1~18.04

**SDK Version (Please Give Commit SHA if Manually Compiling)**
SHA 4867d928942d52d9a2cd7d0871a066c3148a7507
compiled with use_prov_client=ON and use_tpm_simulator=ON

**Protocol**

MQTT, HTTP

**Describe the Bug**

Client connects to IoTHub after obtaining correct iotHubUri and deviceId from provisioning service, both using TPM simulator as authorization. This works fine:

```
auto IotHubClientHandle = IoTHubDeviceClient_CreateFromDeviceAuth(iotHubUri, deviceId, MQTT_Protocol);
// send and receive events work
```

Now when I add this call a very short time after the CreateFromDeviceAuth:
`auto success = IoTHubDeviceClient_UploadToBlobAsync(IotHubClientHandle, "filename", &data[0], data.size(), sUploadCallback, this);`

the application will very often (>50%) not succeed to connect to the IotHub at all. Errors like:
`Error: Time:Wed Apr  1 12:57:52 2020 File:/home/rerk/azure/DeviceConnector/external/azure-iot-sdk-c/provisioning_client/deps/utpm/src/tpm_codec.c Func:TSS_GetTpmProperty Line:679 Get Capability failure`
or other errors from tpm_codec.c appear continuously.

The reason for this appears to be that TPM commands are sent simultaneously from multiple threads in this scenario (see attached stacktrace screenshot), one from the IotHub connection itself, one from the blob upload:
unsafe_call_of_TSS_GetTpmProperty_cropped

As a fix, a lock was added in utpm/inc/azure_utpm_c/tpm_codec.h which guards the TSS_DispatchCmd function in utpm/src/tpm_codec.c. With this fix, the connection happens reliably and both Iothub connection and file upload work as intended.

The fix is checked in to https://github.com/rerkcp/azure-utpm-c.

**Console Logs**

See above.

CIPop:
Adding TPMv2 Software Stack instructions and multi-threading recommendations when using TPMv2.
Updating the submodule to include a fix to the `utpm` dependency.

Fixes #1488

### DIFF
--- a/provisioning_client/devdoc/using_provisioning_client.md
+++ b/provisioning_client/devdoc/using_provisioning_client.md
@@ -28,6 +28,20 @@ cmake -Duse_prov_client:BOOL=ON -Duse_tpm_simulator:BOOL=ON ..
 ./azure-iot-sdk-c/provisioning_client/deps/utpm/tools/tpm_simulator/Simulator.exe
 ```
 
+### TPM Resource Manager
+
+When running on Linux with a hardware TPM (or Hyper-V with virtualized TPMv2) a TPM resource manager should be used to broker access to the hardware and allow multi-threaded operation.
+
+__Important__: Multi-threaded or multi-process TPMv2 usage is supported only when using a TPM resource manager.
+
+See the following resources as well as the tpm2-abrmd manpage for details on how to properly secure the resource manager and access to the hardware TPM: https://tpm2-software.github.io/, https://github.com/tpm2-software/
+
+To install the TPM2 software stack and service on Ubuntu systems:
+
+```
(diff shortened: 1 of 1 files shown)


CIPop:
@amarochk @DMattoon Please review this PR.

PR #1584, opened 2020-07-09
Add TemperatureController to PnP samples

jspaith:
* Add TemperatureController model for a more complex PnP sample.
* Expand the common library specifically with PnP parsing helpers.
* Fix up fit and finish issues in the already checked in simple thermostat, so it's a bit more inline stylistically with rest of files under `./samples/pnp`.


iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+{
+    PNP_THERMOSTAT_COMPONENT* pnpThermostatComponent = (PNP_THERMOSTAT_COMPONENT*)pnpThermostatComponentHandle;
+
+    if (strcmp(propertyName, g_targetTemperaturePropertyName) != 0)
+    {
+        LogError("Property=%s was requested to be changed but is not part of the thermostat interface definition", propertyName);
+    }
+    else
+    {
+        double targetTemperature = json_value_get_number(propertyValue);
```
rido-min, line 327: `json_value_get_number` does not warn of incorrect inputs and always returns the default value of zero. 

Device code should be prepared to validate twin data as untrusted input.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+    pnpThermostatComponent->numTemperatureUpdates++;
+    pnpThermostatComponent->allTemperatures += desiredTemp;
+    pnpThermostatComponent->currentTemperature = desiredTemp;
+}
+
+//
+// SendTargetTemperatureResponse sends a PnP property indicating the device has received the desired targeted temperature
+//
+static void SendTargetTemperatureResponse(PNP_THERMOSTAT_COMPONENT* pnpThermostatComponent, IOTHUB_DEVICE_CLIENT_HANDLE deviceClient, int version)
```
rido-min, line 251: The sample should show how to report ACK with errors, seems like the method always reports PNP_STATUS_SUCCESS

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+    char componentName[PNP_MAXIMUM_COMPONENT_LENGTH + 1];
+    // Current temperature of this thermostat component
+    double currentTemperature;
+    // Minimum temperature this thermostat has been at during current execution run of this thermostat component
+    double minTemperature;
+    // Maximum temperature thermostat has been at during current execution run of this thermostat component
+    double maxTemperature;
+    // Number of times temperature has been updated, counting the initial setting as 1.  Used to determine average temperature of this thermostat component
+    int numTemperatureUpdates;
+    // Total of all temperature updates during current exceution run.  Used to determine average temperature of this thermostat component
```
danewalton, line 68:
```suggestion
    // Total of all temperature updates during current execution run.  Used to determine average temperature of this thermostat component
```

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+    double allTemperatures;
+}
+PNP_THERMOSTAT_COMPONENT;
+
+//
+// BuildUtcTimeFromCurrentTime writes the current time, in ISO 8601 format, into the specified buffer
```
danewalton, line 74: nit: ISO8601

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+static bool BuildUtcTimeFromCurrentTime(char* utcTimeBuffer, size_t utcTimeBufferSize)
+{
+    bool result;
+    time_t currentTime;
+    struct tm * currentTimeTm;
+
+    time(&currentTime);
+    currentTimeTm = gmtime(&currentTime);
+
+    if (snprintf(utcTimeBuffer, utcTimeBufferSize, g_ISO8601Format, currentTimeTm->tm_year + 1900, currentTimeTm->tm_mon, currentTimeTm->tm_mday,
```
danewalton, line 85: any reason for not using `strftime`?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+//
+// BuildMaxMinCommandResponse builds the response to the command for getMaxMinReport
+//
+static bool BuildMaxMinCommandResponse(PNP_THERMOSTAT_COMPONENT* pnpThermostatComponent, unsigned char** response, size_t* responseSize)
+{
+    int responseBuilderSize = 0;
+    unsigned char* responseBuilder = NULL;
+    bool result;
+    char currentTime[128];
```
danewalton, line 148: I think there a few time buffers used throughout. Should the size be `#define`'d?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -0,0 +1,381 @@
...
+
+    if (json_value_get_type(rootValue) != JSONNumber)
+    {
+        LogError("Delay payload is not a number");
+        result = PNP_STATUS_BAD_FORMAT;
+    }
+    else
+    {
+        // See caveats section in ../readme.md; we don't actually respect the delay value to keep the sample simple.
+        int delayInSeconds = (int)json_value_get_number(rootValue);
```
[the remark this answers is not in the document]
jspaith: If someone put in a delay of 1 hour, it means the thread is going to hang there for an hour.  C has a single threaded dispatcher so it would block any future callbacks through that time as well, which is why we tell folks not to block on callbacks.  I'd rather just have a known & doc'd caveat than demonstrate something we tell folks not to do.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -0,0 +1,381 @@
...
+        LogError("Failure creating IotHub device client");
+    }
+    else
+    {
+        LogInfo("Successfully created device client.  Hit Control-C to exit program\n");
+
+        // During startup, send the non-"writeable" properties.
+        PnP_TempControlComponent_ReportSerialNumber_Property(deviceClient);
+        PnP_DeviceInfoComponent_Report_All_Properties(deviceClient, g_deviceInfoComponentName);
+        PnP_TempControlComponent_Report_MaxTempSinceLastReboot_Property(g_thermostatHandle1, deviceClient);
```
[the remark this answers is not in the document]
jspaith: Ugh it's definitely not consistent.  Cleaning up.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+// PnP helper routines
+#include "pnp_protocol_helpers.h"
+#include "pnp_thermostat_component.h"
+
+// Core IoT SDK utilities
+#include "azure_c_shared_utility/xlogging.h"
+
+// The PnP convention defines the maximum length of a component 
+#define PNP_MAXIMUM_COMPONENT_LENGTH 64
```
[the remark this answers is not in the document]
jspaith: Yeah that's potentially useful moving to helpers.

Other components don't need this because they don't need to make copies into a malloc'd data structure so don't need to know max size up front.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+        free(responseBuilder);
+    }
+
+    return result;
+}       
+
+int PnP_ThermostatComponent_ProcessCommand(PNP_THERMOSTAT_COMPONENT_HANDLE pnpThermostatComponentHandle, const char *pnpCommandName, JSON_Value* commandJsonValue, unsigned char** response, size_t* responseSize)
+{
+    PNP_THERMOSTAT_COMPONENT* pnpThermostatComponent = (PNP_THERMOSTAT_COMPONENT*)pnpThermostatComponentHandle;
+    char* jsonStr = NULL;
```
danewalton, line 194: is this not used?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -0,0 +1,381 @@
...
+
+//
+// PnP_TempControlComponent_ReportSerialNumber_Property sends the "serialNumber" property to IoTHub
+//
+static void PnP_TempControlComponent_ReportSerialNumber_Property(IOTHUB_DEVICE_CLIENT_HANDLE deviceClient)
+{
+    IOTHUB_CLIENT_RESULT iothubClientResult;
+    STRING_HANDLE jsonToSend = NULL;
+
+    if ((jsonToSend = PnPHelper_CreateReportedProperty(NULL, g_serialNumberPropertyName, g_serialNumberPropertyValue)) == NULL)
```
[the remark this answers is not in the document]
danewalton: or is this the root

iothub_client/samples/pnp/pnp_temperature_controller/readme.md
```diff
@@ -0,0 +1,19 @@
+# PnP Temperature Controller sample
+
+This directory contains a sample a temperature controller that implements the model [dtmi:com:example:TemperatureController;1](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json).
+
+The model consists of:
+
+* The `reboot` command, `serialNumber` property, and `workingSet` telemetry that the temperature controller implements on the root interface.
+
+* Two thermostat subcomponents, `thermostat1` and `thermostat2`.  (To get a sense of this sample's greater complexity compared to the simpler one, [pnp_simple_thermostat](../pnp_simple_thermostat) *only* implemented a single thermostat with no subcomponents.  The samplein this directory has two such thermostats in addition to everything else.)
```
danewalton, line 9:
```suggestion
* Two thermostat subcomponents, `thermostat1` and `thermostat2`.  (To get a sense of this sample's greater complexity compared to the simpler one, [pnp_simple_thermostat](../pnp_simple_thermostat) *only* implemented a single thermostat with no subcomponents.  The sample in this directory has two such thermostats in addition to everything else.)
```

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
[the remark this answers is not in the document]
jspaith: The handle should actually come 1st.  This follows convention of rest of SDK where handle comes 1st.  The issue is that the deviceInfo & temperatureController don't need handles because they don't actually allocate memory and I'm trying to keep this as small as possible and only malloc where needed.  (I'd considered having a dummy handle for deviceInfo/tempController - so like `#define DEVICE_INFO_HANDLE (0x12345)` to keep it consistent, so it'd just be some magic # that got ignored, but I worry that's more likely to confuse since the handle business in tempController already adds some extra logic here.)  

The other potential option if the handle business is too complex is to just say forget abstraction and have the `struct PNP_THERMOSTAT_COMPONENT` defined in `pnp_thermostat_component.h` so it's just a struct.  @danewalton you've stared at this recently, and specifically your comment RE `This function says TempControlComponent which leads me to believe it is a subcomponent. If that's true, should the NULL param be the component's name?` points to places to clarify.  Do you have thoughts about just saying no handles, all structs for components (that the main() would still NOT directly access, other than knowing how much space to allocate to store on the .bss).  So we'd have 

```c
// pnp_thermostat.h
struct PNP_THERMOSTAT_COMPONENT {...}
PnP_ThermostatComponent_Initialize(PNP_THERMOSTAT_COMPONENT *p,const char* componentName); // replaces PnP_ThermostatComponent_CreateHandle.  

// pnp_temperature_controller.c
PNP_THERMOSTAT_COMPONENT g_thermostat1; // static allocation at top of file

```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,326 @@
...
+
+        if ((propertyName == NULL) || (propertyValue == NULL))
+        {
+            // This should never happen because we are simply accessing parson tree.  Do not pass NULL to application in case it does occur.
+            LogError("Unexpected error retrieving the property name and/or value of component=%s at element at index=%lu", objectName, (unsigned long)i);
+            continue;
+        }
+
+        // The g_JSONComponentMetadata marker is the metadata indicating this is a component the device gets on
+        // when its gets the full device twin.  Don't call the application's callback for metadata.
```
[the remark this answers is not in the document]
jspaith: Rewrote the whole thing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Are you sure?  https://www.iso.org/iso-8601-date-and-time-format.html & https://en.wikipedia.org/wiki/ISO_8601

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -0,0 +1,381 @@
...
+
+// Name of subcomponents that TemmperatureController implements.
+static const char g_thermostatComponent1Name[] = "thermostat1";
+static const size_t g_thermostatComponent1Size = sizeof(g_thermostatComponent1Name) - 1;
+static const char g_thermostatComponent2Name[] = "thermostat2";
+static const size_t g_thermostatComponent2Size = sizeof(g_thermostatComponent2Name) - 1;
+static const char g_deviceInfoComponentName[] = "deviceInformation";
+
+static const char* g_modeledComponents[] = {g_thermostatComponent1Name, g_thermostatComponent2Name, g_deviceInfoComponentName};
+static size_t g_numModeledComponents = sizeof(g_modeledComponents) / sizeof(g_modeledComponents[0]);
```
hihigupt, line 58: >size_t [](start = 7, length = 6)

Can be const

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+#include "azure_c_shared_utility/xlogging.h"
+
+// The PnP convention defines the maximum length of a component 
+#define PNP_MAXIMUM_COMPONENT_LENGTH 64
+
+// The default temperature to use before any is set
+#define DEFAULT_TEMPERATURE_VALUE 30
+
+// Name of command this component supports to retrieve a report about the component.
+static const char g_GetMinMaxReport[] = "getMaxMinReport";
```
hihigupt, line 24: >g_GetMinMaxReport [](start = 18, length = 17)

nit: lowercase start letter.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+    return (PNP_THERMOSTAT_COMPONENT_HANDLE)thermostatComponent;
+}
+
+void PnP_ThermostatComponent_Destroy(PNP_THERMOSTAT_COMPONENT_HANDLE pnpThermostatComponentHandle)
+{
+    if (pnpThermostatComponentHandle != NULL)
+    {
+        free(pnpThermostatComponentHandle);
+    }
```
hihigupt, line 137: can we set it to null after free?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+        *response = responseBuilder;
+        *responseSize = (size_t)responseBuilderSize;
+    }
+    else
+    {
+        free(responseBuilder);
+    }
+
+    return result;
+}
```
hihigupt, line 189: >        [](start = 1, length = 7)

nit: remove spaces

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+
+PNP_THERMOSTAT_COMPONENT_HANDLE PnP_ThermostatComponent_CreateHandle(const char* componentName)
+{
+    PNP_THERMOSTAT_COMPONENT* thermostatComponent;
+
+    if (strlen(componentName) > PNP_MAXIMUM_COMPONENT_LENGTH)
```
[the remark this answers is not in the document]
jspaith: I suppose I don't have to, but if I don't make this check then I end up over-running the buffer if a length is too long.  I get it's sample code but having spent a career in OS's won't let me sleep well if I do that.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Good catch.  Other places in code do a `json_value_get_type` to make sure it's actually a number.  Adding that here too.  (`pnp_simple_thermostat.c` equivalent code of this has same bug which I'm fixing here too.)

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+        LogError("Unable to build response for component=%s", pnpThermostatComponent->componentName);
+        result = PNP_STATUS_INTERNAL_ERROR;
+    }
+    else
+    {
+        LogInfo("Returning success from command request for component=%s", pnpThermostatComponent->componentName);
+        result = PNP_STATUS_SUCCESS;
+    }
+
+    free(jsonStr);
```
hihigupt, line 224: this can be removed.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -0,0 +1,367 @@
...
+
+//
+// UpdateTemperatureAndStatistics updates the temperature and min/max/average values
+//
+static void UpdateTemperatureAndStatistics(PNP_THERMOSTAT_COMPONENT* pnpThermostatComponent, double desiredTemp, bool* maxTempUpdated)
+{
+    if (desiredTemp > pnpThermostatComponent->maxTemperature)
+    {
+        pnpThermostatComponent->maxTemperature = desiredTemp;
+        *maxTempUpdated = true;
```
hihigupt, line 236: >maxTempUpdated [](start = 9, length = 14)

lets make maxTempUpdated = false in all other cases

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.h
```diff
@@ -0,0 +1,20 @@
...
+#define PNP_DEVICEINFO_COMPONENT_H
+
+#include "iothub_device_client.h"
+
+//
+// PnP_DeviceInfoComponent_Report_All_Properties sends properties corresponding to the DeviceInfo interface to the cloud.
+//
+void PnP_DeviceInfoComponent_Report_All_Properties(IOTHUB_DEVICE_CLIENT_HANDLE deviceClient, const char* componentName);
+
+#endif
```
hihigupt, line 20: nit: add /* PNP_DEVICEINFO_COMPONENT_H */

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// The DTDL for this interface is defined at https://repo.azureiotrepository.com/Models/dtmi:azure:DeviceManagement:DeviceInformation;1?api-version=2020-05-01-preview
+
+
+// Standard C header files
+#include <stdio.h>
```
hihigupt, line 8: can this also be removed?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Fixed.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: yes i like this layer not putting assumptions on caller's state.  Fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.h
jspaith: fixed.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
```diff
@@ -0,0 +1,93 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
```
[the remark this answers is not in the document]
jspaith: Yes they can.  fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
```diff
@@ -0,0 +1,93 @@
...
+#include "pnp_deviceinfo_component.h"
+#include "pnp_protocol_helpers.h"
+
+// Core IoT SDK utilities
+#include "azure_c_shared_utility/xlogging.h"
+
+// Property names along with their simulated values.  
+// NOTE: the property values must be legal JSON values.  Strings specifically must be enclosed with an extra set of quotes to be legal json string values.
+// The property names in this sample do not hard-code the extra quotes because the underlying PnP helper adds this to names automatically.
+#define PNP_ENCODE_STRING_FOR_JSON(str) "\"" str "\""
```
hihigupt, line 23: how about using this 

#define stringfy(S) #S



Something like this
[https://onlinegdb.com/BkUc4pnkP](https://onlinegdb.com/BkUc4pnkP)

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
jspaith: Yes it's gone.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
jspaith: Definitely want to use the preprocesser stringify `#`.  I'm going to leave the macro named a more verbose `PNP_ENCODE_STRING_FOR_JSON` so that the intent here is clear.  Fact you have to explicitly add the extra "..." business for JSON strings via a C string build up has caused some confusion so I'd want to be really explicit on that point.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        // Free the memory allocated to track simulated thermostat.
+        PnP_ThermostatComponent_Destroy(g_thermostatHandle2);
+        PnP_ThermostatComponent_Destroy(g_thermostatHandle1);
+
+        // Clean up the iothub sdk handle
+        IoTHubDeviceClient_Destroy(deviceClient);
+        // Free all the sdk subsystem
+        IoTHub_Deinit();
```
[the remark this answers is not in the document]
jspaith: fixed.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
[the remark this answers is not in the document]
jspaith: OK thought more on this and while it's not technically needed, the risk of someone taking this code as-is and buffer overrunning really, really would not let me sleep at night.,  A buffer overrun wouldn't necessarily crash and even Valgrind wouldn't be able to detect it if it didn't over-run `sizeof(PNP_THERMOSTAT_COMPONENT)`.  I've also considered just relying on the Hub to barf because the componentLength is too big other stuff would break, but if the hub changed that policy then we couldn't count on that.

Bottom line here this is a memory bounds check, not a policy check I can rely on Cloud for.  Only way to safely remove this would be to `malloc` the `componentName` field but I don't want to do that either.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,326 @@
...
+// VisitDesiredObject visits each child JSON element of the desired device twin.  As we parse each property out, we invoke the application's passed in pnpPropertyCallback.
+//
+static bool VisitDesiredObject(JSON_Object* desiredObject, const char** componentsInModel, size_t numComponentsInModel, PnPHelperPropertyCallbackFunction pnpPropertyCallback, void* userContextCallback)
+{
+    JSON_Value* versionValue = NULL;
+    size_t numChildren;
+    int version;
+    bool result;
+
+    if ((versionValue = json_object_get_value(desiredObject, g_JSONPropertyVersion)) == NULL)
```
hihigupt, line 197: >versionValue [](start = 9, length = 12)

this need to be freed using json_free() after success

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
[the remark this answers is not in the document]
jspaith: OK - the issue here is that I want to keep sample simple, but I also want to model abstracting and specifically composition of multiple components where we don't have data structure details poking out in a header.

My inclination is to leave code as-is, except for some of the consistency cleanup that Rido had initially called out that is indeed legit.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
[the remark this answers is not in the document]
rido-min: got you, but then ... should we check for propertyNames/commandNames length? whatever we choose, we should try to be consistent with other samples. I've not seen those checks in the comp-less sample.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
danewalton: Ah that way works too. I think above in this file it's one word

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Fixing and also in simple thermostat which had same sprintf() pattern.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Well that's no good.  Thanks.  Cleaning up my comments.

iothub_client/samples/pnp/pnp_temperature_controller/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+#this is CMakeLists.txt for pnp_temperature_controller
+
+compileAsC99()
+
+set(pnp_temperature_controller_c_files
+    pnp_temperature_controller.c
+	pnp_thermostat_component.c
+	pnp_deviceinfo_component.c
+	../common/pnp_device_client_helpers.c
+	../common/pnp_protocol_helpers.c
```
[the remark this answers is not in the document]
jspaith: fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Per offline discussion this makes sense to me.  Let's create an IOU for this as ideally there'd be some consistency across SDKs on this and i don't want to hold checkin on this.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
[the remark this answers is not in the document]
jspaith: Yes this is indeed the root.  Had some other thoughts RE cleaning this up, namely splitting into separate files, we'll need to sync on offline.  I'm not going to hold up this PR on such a potentially large change but would have a separate checkin if makes sense.

iothub_client/samples/pnp/common/readme.md
```diff
@@ -0,0 +1,17 @@
+# Common helper utilities for PnP Sample
+
+This directory contains sample code that your application should be able to take with little or no modification to speed implementing a PnP device application.
+
+Probably the easiest way to understand how they work is to review the [pnp_temperature_controller](../pnp_temperature_controller) sample, which uses them extensively.
+
+For reference the files are:
+
+* `pnp_device_client_helpers` header and .c file implement a function to help create a `IOTHUB_DEVICE_CLIENT_HANDLE`.  The `IOTHUB_DEVICE_CLIENT_HANDLE` is an existing IoTHub Device SDK that can be used for device <=> IoTHub communication.  (There are many samples demonstrating its non-PnP uses in the [samples parent directory](../..).)
```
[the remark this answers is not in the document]
jspaith: The statement in the parenthesis is a complete sentence, not a random fragment, so I think there should be a period terminating it.  I also think the parenthesis makes sense because this isn't critical to understanding this sample.  So leaving as-is.

iothub_client/samples/pnp/common/pnp_device_client_helpers.c
```diff
@@ -23,77 +23,76 @@ IOTHUB_DEVICE_CLIENT_HANDLE PnPHelper_CreateDeviceClientHandle(const char* conne
...
-        result = MU_FAILURE;
+        LogError("Failure to initialize client, error=%d", iothubInitResult);
+        result = false;
     }
     // Create the deviceHandle itself.
     else if ((deviceHandle = IoTHubDeviceClient_CreateFromConnectionString(connectionString, MQTT_Protocol)) == NULL)
     {
-        LogError("Failure creating Iothub device.  Hint: Check you connection string");
-        result = MU_FAILURE;
+        LogError("Failure creating IotHub device.  Hint: Check you connection string");
```
ericwol-msft, line 37:
```suggestion
        LogError("Failure creating IotHub client.  Hint: Check you connection string");
```

iothub_client/samples/pnp/common/pnp_device_client_helpers.c
```diff
@@ -23,77 +23,76 @@ IOTHUB_DEVICE_CLIENT_HANDLE PnPHelper_CreateDeviceClientHandle(const char* conne
...
     // handle is created, but will implicitly connect to subscribe for device method and device twin callbacks below.
     else if ((iothubResult = IoTHubDeviceClient_SetOption(deviceHandle, OPTION_MODEL_ID, modelId)) != IOTHUB_CLIENT_OK)
     {
         LogError("Unable to set the ModelID, error=%d", iothubResult);
-        result = MU_FAILURE;
+        result = false;
     }
-    // Sets the callback function that processes incoming device methods, which is the channel PnP Commands are transferred over
-    else if ((iothubResult = IoTHubDeviceClient_SetDeviceMethodCallback(deviceHandle, deviceMethodCallback, NULL)) != IOTHUB_CLIENT_OK)
+    // Optionally sets the callback function that processes incoming device methods, which is the channel PnP Commands are transferred over
```
ericwol-msft, line 54:
```suggestion
    // Optionally, set the callback function that processes incoming device methods, which is the channel PnP Commands are transferred over
```

iothub_client/samples/pnp/common/pnp_device_client_helpers.c
```diff
@@ -23,77 +23,76 @@ IOTHUB_DEVICE_CLIENT_HANDLE PnPHelper_CreateDeviceClientHandle(const char* conne
...
+    else if ((deviceMethodCallback != NULL) && (iothubResult = IoTHubDeviceClient_SetDeviceMethodCallback(deviceHandle, deviceMethodCallback, NULL)) != IOTHUB_CLIENT_OK)
     {
         LogError("Unable to set device method callback, error=%d", iothubResult);
-        result = MU_FAILURE;
+        result = false;
     }
-    // Sets the callback function that processes device twin changes from the hub, which is the channel that PnP Properties are 
+    // Optionally sets the callback function that processes device twin changes from the IoTHub, which is the channel that PnP Properties are
```
ericwol-msft, line 60:
```suggestion
    // Optionall, set the callback function that processes device twin changes from the IoTHub, which is the channel that PnP Properties are 
```

iothub_client/samples/pnp/common/pnp_device_client_helpers.c
```diff
@@ -23,77 +23,76 @@ IOTHUB_DEVICE_CLIENT_HANDLE PnPHelper_CreateDeviceClientHandle(const char* conne
...
+    // Optionally sets the callback function that processes incoming device methods, which is the channel PnP Commands are transferred over
+    else if ((deviceMethodCallback != NULL) && (iothubResult = IoTHubDeviceClient_SetDeviceMethodCallback(deviceHandle, deviceMethodCallback, NULL)) != IOTHUB_CLIENT_OK)
     {
         LogError("Unable to set device method callback, error=%d", iothubResult);
-        result = MU_FAILURE;
+        result = false;
     }
-    // Sets the callback function that processes device twin changes from the hub, which is the channel that PnP Properties are 
+    // Optionally sets the callback function that processes device twin changes from the IoTHub, which is the channel that PnP Properties are 
     // transferred over.  This will also automatically retrieve the full twin for the application.
```
ericwol-msft, line 61:
```suggestion
    // transferred over.  This will also automatically retrieve the full twin for the application on startup. 
```

iothub_client/samples/pnp/common/pnp_device_client_helpers.c
```diff
@@ -23,77 +23,76 @@ IOTHUB_DEVICE_CLIENT_HANDLE PnPHelper_CreateDeviceClientHandle(const char* conne
...
     {
         LogError("Unable to set auto Url encode option, error=%d", iothubResult);
-        result = MU_FAILURE;
+        result = false;
     }
 #ifdef SET_TRUSTED_CERT_IN_SAMPLES
     // Setting the Trusted Certificate.  This is only necessary on systems without built in certificate stores.
     else if ((iothubResult = IoTHubDeviceClient_SetOption(deviceHandle, OPTION_TRUSTED_CERT, certificates)) != IOTHUB_CLIENT_OK)
     {
         LogError("Unable to set auto Url encode option, error=%d", iothubResult);
```
ericwol-msft, line 77:
```suggestion
        LogError("Unable to set the trusted cert, error=%d", iothubResult);
```

iothub_client/samples/pnp/common/pnp_device_client_helpers.h
```diff
@@ -8,10 +8,14 @@
 
 //
 // PnPHelper_CreateDeviceClientHandle creates an IOTHUB_DEVICE_CLIENT_HANDLE that will be ready to interact with PnP.
-// Beyond basic handle creation, it also sets the handle to the appropriate Model Id, sets up callback functions
+// Beyond basic handle creation, it also sets the handle to the appropriate ModelId, optionally sets up callback functions
 // for Device Method and Device Twin callbacks (to process PnP Commands and Properties, respectively)
 // as well as some other basic maintenence on the handle. 
 //
+// If the application's model does not implement PnP commands, the application should set deviceMethodCallback to NULL.
+// Anologously, if there are no desired PnP properties deviceTwinCallback should be NULL.  Not subscribing for these
```
ericwol-msft, line 16:
```suggestion
// Anologously, if there are no desired PnP properties, deviceTwinCallback should be NULL.  Not subscribing for these
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+// Format used when building a response for a root property that does contain metadata
+static const char g_propertyWithResponseSchemaWithoutComponent[] =  "{\"%s\":{\"value\":%s,\"ac\":%d,\"ad\":\"%s\",\"av\":%d}}";
+// Format used when building a response for a component's property that does contain metadata
+static const char g_propertyWithResponseSchemaWithComponent[] =  "{\"""%s\":{\"__t\":\"c\",\"%s\":{\"value\":%s,\"ac\":%d,\"ad\":\"%s\",\"av\":%d}}}";
+
+// Character that separates a PnP component from the specific command on the component.
+static const char g_commandSeparator = '*';
+
+// The version of the desired twin is represented by the $version metadata.
+static const char g_JSONPropertyVersion[] = "$version";
```
ericwol-msft, line 27: Humm...
```suggestion
static const char g_DtdlPropertyVersion[] = "$version";
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+
+// IoTHub adds a JSON field "__t":"c" into desired top-level JSON objects that represent components.  Without this marking, the object 
+// is treated as a property off the root component.
+static const char g_JSONComponentMetadata[] = "__t";
```
ericwol-msft, line 31: ??

```suggestion
static const char g_DtDlComponentMetadata[] = "__t";
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+
+// Name of desired JSON field when retrieving a full twin.
+static const char g_JSONDesiredName[] = "desired";
```
ericwol-msft, line 34:
```suggestion
static const char g_DtdlDesiredName[] = "desired";
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+// Telemetry message property used to indicate the message's component.
+static const char PnP_TelemetryComponentProperty[] = "$.sub";
+
+STRING_HANDLE PnPHelper_CreateReportedProperty(const char* componentName, const char* propertyName, const char* propertyValue)
+{
+    STRING_HANDLE jsonToSend;
+
+    if (componentName == NULL) 
+    {
+        jsonToSend = STRING_construct_sprintf(g_propertyWithoutResponseSchemaWithoutComponent, propertyName, propertyValue);
```
ericwol-msft, line 45:
```suggestion
        jsonToSend = STRING_construct_sprintf(componentName  ? g_propertyWithoutResponseSchemaWithComponent : g_propertyWithoutResponseSchemaWithoutComponent, propertyName, propertyValue);
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+    return jsonToSend;
+}
+
+STRING_HANDLE PnPHelper_CreateReportedPropertyWithStatus(const char* componentName, const char* propertyName, const char* propertyValue, int result, const char* description, int ackVersion)
+{
+    STRING_HANDLE jsonToSend;
+
+    if (componentName == NULL) 
+    {
+        jsonToSend = STRING_construct_sprintf(g_propertyWithResponseSchemaWithoutComponent, propertyName, propertyValue, result, description, ackVersion);
```
ericwol-msft, line 66:
```suggestion
        jsonToSend = STRING_construct_sprintf(componentName ? g_propertyWithResponseSchemaWithComponent : g_propertyWithResponseSchemaWithoutComponent, propertyName, propertyValue, result, description, ackVersion);
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+    return messageHandle;
+}
+
+//
+// VisitComponentProperties visits each sub element of the the given objectName in the desired JSON.  Each of these sub elements corresponds to
+// a property of this component, which we'll invoke the application's pnpPropertyCallback to inform.
+// 
+static void VisitComponentProperties(const char* objectName, JSON_Value* value, int version, PnPHelperPropertyCallbackFunction pnpPropertyCallback, void* userContextCallback)
+{
+    JSON_Object* object = json_value_get_object(value);
```
ericwol-msft, line 140: validate object?

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+    free(jsonStr);
+
+    return result;
+}
+
+char* PnPHelper_CopyPayloadToString(const unsigned char* payload, size_t size)
+{
+    char* jsonStr;
+
+    if ((jsonStr = (char*)malloc(size + 1)) == NULL)
```
ericwol-msft, line 316: "size + 1" to a var, so that it's not recomputed again for the error msg

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
```diff
@@ -0,0 +1,327 @@
...
+    {
+        LogError("Unable to allocate %lu size buffer", (unsigned long)(size + 1));
+    }
+    else
+    {
+        memcpy(jsonStr, payload, size);
+        jsonStr[size] = 0;
```
ericwol-msft, line 323:
```suggestion
        jsonStr[size] = '\0';
```

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -387,6 +389,10 @@ static void Thermostat_DeviceTwinCallback(DEVICE_TWIN_UPDATE_STATE updateState,
         // something is fundamentally wrong with how we've received the twin and we should not proceed.
         LogError("JSON field %s is not a number but must be", g_JSONPropertyVersion);
     }
+    else if (json_value_get_type(targetTemperatureValue) != JSONNumber)
+    {
+        LogError("JSON field %s is not a number but must be", g_JSONTargetTemperature);
```
ericwol-msft, line 394:
```suggestion
        LogError("JSON field %s is not a number", g_JSONTargetTemperature);
```

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: I'd considered but I don't technically have to here.  If `object` is NULL, then the subsequent call to `json_object_get_count` will end up returning a 0 and the following loop will be a giant no-op.  And note in practice this call shouldn't ever fail anyway, as we're already marked as a static so we can trust caller in our function and it only calls if it knows we're an object.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: Fixing.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: The issue with this is that `g_propertyWithoutResponseSchemaWithComponent` takes 3 variadic whereas `g_propertyWithoutResponseSchemaWithoutComponent` only takes 2.  I guess I could put some sort of dummy %s with a "" in if componentName==NULL, but I don't think that's as readable as having separate statements.  So I plan to leave as-is.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: Same feedback as other comment above, about variadic arguments not having same number in the two format identifiers.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: The `$version` is an IoTHub twin construct, not part of DTDL.  Similar feedback the comments for `__t` and `desired`.  DTDL's defining the model is not necessarily bound to an IoTHub twin though for now it is of course.

The naming as-is isn't great.  I'm thinking `g_JSONPropertyVersion` => `g_IoTHubDesiredVersion`, `g_JSONComponentMetadata` => `g_IoTHubTwinPnPComponentMarker`, `g_JSONDesiredName`=>`g_IoTHubDesiredObjectName`.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: See my comment to the `g_JSONPropertyVersion` feedback.

iothub_client/samples/pnp/common/pnp_protocol_helpers.c
jspaith: See comment for `g_JSONPropertyVersion` feedback.

iothub_client/samples/pnp/common/pnp_device_client_helpers.h
danewalton: Also change to `analogously`

Issue #1568, opened 2020-06-22
CreationTimeUtc for D2C message


sanjeev-magoo:
This is regarding setting of system property(CreationTimeUtc) to D2C message. Basically, we wanted to set CreationTimeUtc timestamp in every telemetry message originated from device. It is possible that device is generating D2C messages when working in offline mode, when these messages reached to cloud, we want to know the actual creation time of these messages. 
**Our findings-**
1.	We checked and found that “CreationTimeUtc” property on Message is only supported by C# and Java SDK. We are not able to find support in C SDK. Refer reference #1.
2.	Looks like CreationTimeUtc property on Message in the C# SDK results in SystemProperty “iothub-creation-time-utc” to be added on the message that is sent to IoT Hub, refer reference #2. 
3.	Refer readme.md file mentioned in reference #3, Refer the following section- 

Send   device-to-cloud message | ✔️* | ✔️* | ✔️* | ✔️* | ✔️* | Send   device-to-cloud messages (max 256KB) to IoT Hub with the option to add custom   properties. IoT Hub only supports batch send over AMQP and HTTPS only at the   moment. This SDK supports batch send over HTTP. * Batch send over AMQP and   AMQP-WS, and **add system properties on D2C messages are in progress.**
-- | -- | -- | -- | -- | -- | --
4.	Refer another reference #4 for same property.

**Questions-**

1. Could you please let us when this feature will be available with C SDK? 
2. Let us know, If there is way to set “iothub-creation-time-utc“ property programmatically in system properties of D2C message?

Note- We can temporarily set this property in application properties, but we are looking for solution using system properties as recommended by MS and to minimize the changes later.

**References-**

1. https://github.com/Azure/azure-iot-sdk-java/issues/397
2. https://github.com/Azure/azure-iot-sdk-csharp/blob/cd93e75fd8914b6247f23eb6a5cfd7aea459676f/iothub/device/src/MessageSystemPropertyNames.cs.
3. https://github.com/Azure/azure-iot-sdk-c/blob/master/readme.md
4. https://github.com/Azure/azure-iot-sdk-csharp/issues/376

ericwol-msft:
I can add this work to the backlog. Does this actually work for you?? When I do it, the creation time is not on the system properties, but on the message properties. Routing to blob storage does not respect the creation time. How are you using this??


{
  "body": {
    "foo": "bar"
  },
  "enqueuedTime": "2020-06-24T23:19:20.263Z",
[...]

PR #1596, opened 2020-07-22
CreationTimeUtc for D2C message #1568

ericwol-msft:
# Checklist

#1568

# Description of the problem
Need to add CreationTimeUtc to messages

# Description of the solution
Added IoTHubMessage_SetMessageCreationTimeUtcSystemProperty()


iothub_client/inc/iothub_message.h
```diff
@@ -356,6 +355,28 @@ MOCKABLE_FUNCTION(, IOTHUB_MESSAGE_RESULT, IoTHubMessage_SetConnectionModuleId,
 */
 MOCKABLE_FUNCTION(, const char*, IoTHubMessage_GetConnectionDeviceId, IOTHUB_MESSAGE_HANDLE, iotHubMessageHandle);
 
+/**
+* @brief   Sets the message creation time in UTC. CAUTION: SDK user should not call it directly, it is for internal use only.
+*
+* @param   iotHubMessageHandle Handle to the message.
+* @param   messageCreationTimeUtc Pointer to the message creation time
```
ewertons, line 362:
```Suggestion
* @param   messageCreationTimeUtc Pointer to the message creation time as null-terminated string
```

iothub_client/samples/iothub_ll_telemetry_sample/iothub_ll_telemetry_sample.c
```diff
@@ -151,10 +151,14 @@ int main(void)
...
 
                 // Set Message property
-                /*(void)IoTHubMessage_SetMessageId(message_handle, "MSG_ID");
+                /*
+                (void)IoTHubMessage_SetMessageId(message_handle, "MSG_ID");
                 (void)IoTHubMessage_SetCorrelationId(message_handle, "CORE_ID");
                 (void)IoTHubMessage_SetContentTypeSystemProperty(message_handle, "application%2fjson");
-                (void)IoTHubMessage_SetContentEncodingSystemProperty(message_handle, "utf-8");*/
+                (void)IoTHubMessage_SetContentEncodingSystemProperty(message_handle, "utf-8");
+                (void)IoTHubMessage_SetMessageCreationTimeUtcSystemProperty(message_handle, "2020-07-01T01:00:00.346Z");
```
danewalton, line 159: should this be abstracted into a static function which will get the current time?

iothub_client/src/iothub_client_dll.def
```diff
@@ -120,6 +120,7 @@ EXPORTS
     IoTHubMessage_GetContentType
     IoTHubMessage_GetContentTypeSystemProperty
     IoTHubMessage_GetContentEncodingSystemProperty 
+    IoTHubMessage_GetMessageCreationTimeUtcSystemProperty
```
ewertons, line 123: Good job not missing the dynamic library exports.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1265,6 +1277,19 @@ static int setMqttMessagePropertyIfPossible(IOTHUB_MESSAGE_HANDLE IoTHubMessage,
     // Not finding a system property to map to isn't an error.
     int result = 0;
 
+    if (nameLen > 5)
```
danewalton, line 1280: Is 5 b/c of the length of `ctime`?
In that case I would suggest 
```suggestion
    if (nameLen > strlen(MESSAGE_CREATION_TIME_UTC))
```
Since I don't know off the bat what `5` is here.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1265,6 +1277,19 @@ static int setMqttMessagePropertyIfPossible(IOTHUB_MESSAGE_HANDLE IoTHubMessage,
...
+    {
+        if (strcmp((const char*)&propName[nameLen - 5], MESSAGE_CREATION_TIME_UTC) == 0)
```
danewalton, line 1282: same here what does `5` represent

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1265,6 +1277,19 @@ static int setMqttMessagePropertyIfPossible(IOTHUB_MESSAGE_HANDLE IoTHubMessage,
     // Not finding a system property to map to isn't an error.
```
danewalton, line 1277: actually this function is confusing. should `propName` be a `const char**`?

iothub_client/src/iothub_message.c
```diff
@@ -348,6 +373,12 @@ IOTHUB_MESSAGE_HANDLE IoTHubMessage_Clone(IOTHUB_MESSAGE_HANDLE iotHubMessageHan
                 DestroyMessageData(result);
                 result = NULL;
             }
+            else if (source->creationTimeUtc != NULL && mallocAndStrcpy_s(&result->creationTimeUtc, source->creationTimeUtc) != 0)
+            {
+                LogError("unable to copy creationTimeUtc");
+                DestroyMessageData(result);
```
ewertons, line 379: DestroyMessageData should be changed to free ->creationTimeUtc, otherwise we will introduce a memory leak.
Also there should be a test failing somewhere without that change.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons, line 1280: It would be good to define what 5 is at the top (as a macro), so it doesn't become a magic number.

iothub_client/inc/iothub_message.h
```diff
@@ -356,6 +355,28 @@ MOCKABLE_FUNCTION(, IOTHUB_MESSAGE_RESULT, IoTHubMessage_SetConnectionModuleId,
...
+*
```
ewertons, line 363: So as I can see in the PR this is not implemented for AMQP transport.
That should be documented in these functions as a remark.

iothub_client/samples/iothub_ll_telemetry_sample/iothub_ll_telemetry_sample.c
ericwol-msft: good question, this is override the current time of the message. If you leave it as it is, it's obvious that it's an ISO date time string.

iothub_client/inc/iothub_message.h
ericwol-msft: should I?

iothub_client/src/iothubtransport_mqtt_common.c
ericwol-msft: This is how it's implemented in the rest of this function. Would require refactoring the function to make this not a one off.

iothub_client/src/iothubtransport_mqtt_common.c
ericwol-msft: Yes, it's an optimization of who ever wrote this function.
But there may be other strcmp in this if statement, so don't tie it to this one. See the rest of the implementation.

ericwol-msft: It's not just this 5, you would need to refactor the function to not make this a one off.

iothub_client/src/iothubtransport_mqtt_common.c
ericwol-msft: no, it's just a pointer to the property name string

iothub_client/inc/iothub_message.h
ewertons: That would be good

PR #1610, opened 2020-07-31
Add DPS to simple model & move to LL layer

jspaith:
* Move from convenience layer to LL based handles so this works on more restricted devices out of the box.
* Add DPS sample to the pnp_simple_thermometer.
* Code and comment cleanup

### DIFF
--- a/iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
+++ b/iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
@@ -27,11 +27,45 @@
 #include "certs.h"
 #endif // SET_TRUSTED_CERT_IN_SAMPLES
 
+#ifdef USE_PROV_MODULE_FULL
+#include "pnp_dps_ll.h"
+#endif // USE_PROV_MODULE_FULL
+
+
 // JSON parser library
 #include "parson.h"
 
+// Environment variable used to specify how app connects to hub and the two possible values
+static const char g_securityTypeEnvironmentVariable[] = "IOTHUB_DEVICE_SECURITY_TYPE";
+static const char g_securityTypeConnectionStringValue[] = "connectionString";
+static const char g_securityTypeDpsValue[] = "DPS";
+
 // Environment variable used to specify this application's connection string
 static const char g_connectionStringEnvironmentVariable[] = "IOTHUB_DEVICE_CONNECTION_STRING";
+
+#ifdef USE_PROV_MODULE_FULL
+// Environment variable used to specify this application's DPS id scope
+static const char g_dpsIdScopeEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_ID_SCOPE";
+
+// Environment variable used to specify this application's DPS device id
+static const char g_dpsDeviceIdEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_DEVICE_ID";
+
+// Environment variable used to specify this application's DPS device key
+static const char g_dpsDeviceKeyEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_DEVICE_KEY";
+
+// Environment variable used to optionally specify this application's DPS id scope
+static const char g_dpsEndpointEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_ENDPOINT";
+
+// Global provisioning endpoint for DPS if one is not specified via the environment
+static char g_dps_DefaultGlobalProvUri[] = "global.azure-devices-provisioning.net";
+
+// Values of connection / security settings read from environment variables and/or DPS runtime
+PNP_DEVICE_CONFIGURATION g_pnpDeviceConfiguration;
+#endif // USE_PROV_MODULE_FULL
+
+// Connection string used to authenticate device when connection strings are used
+const char* g_pnpDeviceConnectionString;
+
 
 // Amount of time to sleep between polling hub, in milliseconds.  Set to wake up every 100 milliseconds.
 static unsigned int g_sleepBetweenPolls = 100;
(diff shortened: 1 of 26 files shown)


iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -27,14 +27,50 @@
...
+static const char g_dpsDeviceIdEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_DEVICE_ID";
+
+// Environment variable used to specify this application's DPS device key
+static const char g_dpsDeviceKeyEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_DEVICE_KEY";
+
+// Environment variable used to optionally specify this application's DPS id scope
+static const char g_dpsEndpointEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_ENDPOINT";
+
+// Global provisioning endpoint for DPS if one is not specified via the environment
+static char g_dps_DefaultGlobalProvUri[] = "global.azure-devices-provisioning.net";
```
ericwol-msft, line 59: const?

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -27,14 +27,50 @@
...
+
+// Values of connection / security settings read from environment variables and/or DPS runtime
+PNP_DEVICE_CONFIGURATION g_pnpDeviceConfiguration;
+#endif // USE_PROV_MODULE_FULL
+
+// Connection string used to authenticate device when connection strings are used
+const char* g_pnpDeviceConnectionString;
+
+// Amount of time to sleep between polling hub, in milliseconds.  Set to wake up every 100 milliseconds.
+static unsigned int g_sleepBetweenPolls = 100;
```
ericwol-msft, line 69:
```suggestion
static unsigned int g_sleepBetweenPollsMs = 100;
```

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -27,14 +27,50 @@
...
+
+// Every time the main loop wakes up, on the g_sendTelemetryFrequency(th) pass will send a telemetry message.
+// So we will send telemetry every (g_sendTelemetryFrequency * g_sleepBetweenPolls) milliseconds; 60 seconds as currently configured.
+static const int g_sendTelemetryFrequency = 600;
```
ericwol-msft, line 73:
```suggestion
static const int g_sendTelemetryFrequencyMs = 600;
```

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -472,12 +507,137 @@ void Thermostat_SendCurrentTemperature(IOTHUB_DEVICE_CLIENT_HANDLE deviceClient)
...
+    }
+
+    return result;    
+}
+
+//
+// CreateDevicelientHandle performs actual handle creation (but nothing more), depending
+// on whether connection strings or DPS is used.
+//
+static IOTHUB_DEVICE_CLIENT_LL_HANDLE CreateDevicelientHandle(void)
```
ericwol-msft, line 618:
```suggestion
static IOTHUB_DEVICE_CLIENT_LL_HANDLE CreateDevicelientLLHandle(void)
```

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -555,41 +715,45 @@ static IOTHUB_DEVICE_CLIENT_HANDLE CreateDeviceClientHandleForPnP(const char* co
...
+        LogError("Cannot read required environment variable(s)");
     }
     else if (BuildUtcTimeFromCurrentTime(g_ProgramStartTime, sizeof(g_ProgramStartTime)) == false)
     {
         LogError("Unable to output the program start time");
     }
-    else if ((deviceClient = CreateDeviceClientHandleForPnP(connectionString)) == NULL)
+    else if ((deviceClientLL = CreateAndConfigureDeviceClientHandleForPnP()) == NULL)
     {
         LogError("Failed creating IotHub device");
```
ericwol-msft, line 732:
```suggestion
        LogError("Failed creating IotHub device client");
```

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
jspaith: Fixing.

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
jspaith: Fixing.

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
jspaith: This actually isn't in milliseconds.  It's basically every time through the main() loop, we'll DoWork() each time and every 600 do a send telemetry.  This lets us do a simple modulo on a counter, instead of having to do any math on time.

I agree "frequency" can mean too many things to different people.  I'm going with `g_sendTelemetryPollInterval` to try and capture that we're tied to polling and so people don't read into this as much.

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
jspaith: Fixing.

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
jspaith: Fixing.

jspaith:
@rido-min re your comment about model-id in CONNECT packet.  First thanks for reading debug spew.  We're fine in this case though; this CONNECT is for connectiong to the DPS service itself and we use the DPS payload and not an MQTT primitive to communicate modelId.  I've double checked both the DPS/PnP spec on this that nothing has changed as well as verifying the body is successfully put in request payload at runtime.

wickste:
> @rido-min re your comment about model-id in CONNECT packet. First thanks for reading debug spew. We're fine in this case though; this CONNECT is for connectiong to the DPS service itself and we use the DPS payload and not an MQTT primitive to communicate modelId. I've double checked both the DPS/PnP spec on this that nothing has changed as well as verifying the body is successfully put in request payload at runtime.

@jspaith I think what Rido is referring to is that we need to pass in the modelId when we ultimately create the DeviceClient, i.e. with IoTHubDeviceClient_CreateFromDeviceAuth. If we don't pass it in here, the solution will not be able to retrieve the modelId for its registered devices. The value we pass in to DPS is only used for IoTHub assignment, but it's not stored anywhere.

jspaith:
@wickste @rido-min  - understood and code should be doing right thing.  In simple thermomoter code, checkout the line with `IoTHubDeviceClient_LL_SetOption(deviceHandle, OPTION_MODEL_ID, g_ThermostatModelId)` this is invoked regardless of whether the handle was created with a connection string or DPS.  (For that matter, all the code after initial creation doesn't care about auth mechanism.)

I think confusion may be just how C versus C# handle modelId setting.  C doesn't have a mechanism to specify options at handle creation time so we have the set option is a well defined pattern for us, whereas C# can put this all into a single line.

It's more hidden because of helpers but in temperature sensor sample also sets it both for DPS & IoTHub.

wickste:
@jspaith ok, as long as it works we are good. I guess I am better at testing code than at reading code :-)

Issue #1618, opened 2020-08-06
Heap Buffer Overflow when sending message with Property


DonHomer1942:
Heap Buffer Overflow in iothubtransport_mqtt_common.c in function
```
static bool isSystemProperty(const char* tokenData)
{
    bool result = false;
    size_t propCount = sizeof(sysPropList) / sizeof(sysPropList[0]);
    size_t index = 0;
    for (index = 0; index < propCount; index++)
    {
        if (memcmp(tokenData, sysPropList[index].propName, sysPropList[index].propLength) == 0)
        {
            result = true;
            break;
        }
    }
    return result;
}
```
The overflow occurs at the memcmp, this happens when tokenData is shorter than a SystemProperty.
I noticed this because i always have -fsanitize=address in my build flags and the iothub_convenience_sample always crashed when I tried to send a message with a short Property attached to it.

ewertons:
Thank you for reporting this issue, @DonHomer1942 !
It is indeed a bug and we shall fix is briefly.

ewertons:
Fixed.
Thanks for reporting this issue and contributing to the Azure IoT SDKs!
-Azure IoT SDK Team.

az-iot-builder-01:
@DonHomer1942, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1664, opened 2020-09-22
Fix Heap Buffer Overflow when sending message with Property (#1618)

ewertons:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/1618

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -1203,15 +1203,17 @@ static int subscribeToNotifyStateIfNeeded(PMQTTTRANSPORT_HANDLE_DATA transport_d
     return result;
 }
 
-
 static bool isSystemProperty(const char* tokenData)
 {
     bool result = false;
     size_t propCount = sizeof(sysPropList) / sizeof(sysPropList[0]);
     size_t index = 0;
+    size_t tokenDataLength = strlen(tokenData);
+
     for (index = 0; index < propCount; index++)
     {
-        if (memcmp(tokenData, sysPropList[index].propName, sysPropList[index].propLength) == 0)
+        if (tokenDataLength >= sysPropList[index].propLength &&
+            memcmp(tokenData, sysPropList[index].propName, sysPropList[index].propLength) == 0)
         {
             result = true;
             break;


ericwol-msft (approved):

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1203,15 +1203,44 @@ static int subscribeToNotifyStateIfNeeded(PMQTTTRANSPORT_HANDLE_DATA transport_d
     return result;
 }
 
+// The tokenData variable carries a string in the format "<name>=<value".
```
momuno, line 1206: "\<name\>=\<value" to "\<name\>=\<value\>"

momuno (approved):

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1203,15 +1203,44 @@ static int subscribeToNotifyStateIfNeeded(PMQTTTRANSPORT_HANDLE_DATA transport_d
...
+// This function returns the length of the <name> (the portion of the string from position 0 to the last char before '=').
+// If the '=' character is not present (unlikely), the length returned is 0, which should be
+// properly handled by the calling function.
+static int getPropertyLength(const char* tokenData)
```
ericwol-msft, line 1210: {
    int result = 0;

    while (tokenData[result] == '\0')
    {
        if (tokenData[result] == '=')
        {
            return result;
        }
        result++;
    }

    return 0;
}

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1203,15 +1203,44 @@ static int subscribeToNotifyStateIfNeeded(PMQTTTRANSPORT_HANDLE_DATA transport_d
...
+}
 
 static bool isSystemProperty(const char* tokenData)
 {
     bool result = false;
     size_t propCount = sizeof(sysPropList) / sizeof(sysPropList[0]);
     size_t index = 0;
+    size_t tokenDataLength = getPropertyLength(tokenData);
+
     for (index = 0; index < propCount; index++)
```
ericwol-msft, line 1240: for (index = 0; tokenDataLength  > 0 && index < propCount; index++)

ericwol-msft (approved):

iothub_client/src/iothubtransport_mqtt_common.c
momuno: But what if tokenData does not have a terminating null character? Could there be a seg fault?

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: the tokenDataLength must be compared with each property in the list we maintain in code, so this check would not address the issue.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: This function has been removed. Unfortunately it didnt work for all the cases, according to how the MQTT transport is implemented. See comment on the issue discussion.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Good catch. However this function has been removed.

ewertons:
The MQTT transport in the Azure-IoT-SDK-C code optimizes the logic to search for system properties in C2D message topic names. 

Instead of looking for "%24.cid" (which is usually the first property in the topic name), the code checks if a token starts  

with "device/" (e.g., in the token "devices/csdk_e2eDevice_sas_j_please_delete_06a6b799-7867-45d8-85ba-4716cdbf15ae/messages/devicebound/%24.cid=MessageCorrelationIdForE2E") 

assuming that such token contains "cid" as first property, extracting the property value (the portion after '=') and setting the correlation id using IoTHubMessage_SetCorrelationId. 

Given that, this fix was simplified as much as possible.

Issue #1290, opened 2019-11-04
IoTHubMessage_SetProperty IOTHUB_MESSAGE_ERROR should discern between out of memory situations and input errors


Ozaq:
I have done my best to figure out if this has been requested earlier but I could not find any information. 

When working with `IoTHubMessage_SetProperty` I noticed that only printable US-ACSII characters are allowed for keys / values since `IoTHubMessage_CreateFromByteArray` and `IoTHubMessage_CreateFromString` set a ValidateAsciiCharactersFilter for the internal property map.

`IoTHubMessage_SetProperty` any error in this call will return a `IOTHUB_MESSAGE_ERROR` making it impossible to discern between out-of-memory situations and "ordinary" input validation. 

Please introduce another error state that make those cases differentiable.

Also please update the documentation to include the constraint about printable US-ASCII characters in https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-message-h/iothubmessage-setproperty.

az-iot-builder-01:
@Ozaq, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1665, opened 2020-09-23
Fix for Issue 1290

momuno:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/1290
# Description of the problem
Addresses Issue 1290

# Description of the solution
Uses the unused error `IOTHUB_MESSAGE_INVALID_TYPE` to differentiate when the ascii check fails in calling `IoTHubMessage_SetProperty.`
Adds 2 unit tests accordingly.

This PR comes from a branch of azure-iot-sdk-c instead of from a fork of that repo.  This is an attempt to pass the esp testing that was failing (possibly due to "References service endpoint. PRs from repository forks are not allowed to access secrets in the pipeline").


danewalton (approved):

iothub_client/inc/iothub_message.h
```diff
@@ -191,9 +191,9 @@ MOCKABLE_FUNCTION(, MAP_HANDLE, IoTHubMessage_Properties, IOTHUB_MESSAGE_HANDLE,
 *
 * @param   iotHubMessageHandle Handle to the message.
 *
-* @param   key name of the property to set.  Note that when sending messages via the HTTP transport, this value must not contain spaces.
+* @param   key name of the property to set using valid ASCII characters.  Note that when sending messages via the HTTP transport, this value must not contain spaces.
```
ericwol-msft, line 194: this is not true, it's defined as below

Property names and values can only contain ASCII **alphanumeric** characters, plus {'!', '#', '$', '%, '&', ''', '*', '+', '-', '.', '^', '_', '`', '|', '~'} 

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct

momuno: How do you think I best explain that here?  I was hoping "valid" would suffice.  I could just say between dec value [32, 126]

momuno: or [' ', '~']

momuno: Except this page says there should be no spaces.... https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-message-h/iothubmessage-setproperty

ericwol-msft: a space is not an ASCII alphanumeric character

momuno: The docs specify no space for the key.  But no mention for the value.
The code allows spaces.  Are you saying we should not allow spaces for the value either?

iothub_client/src/iothub_message.c
```diff
@@ -43,8 +43,8 @@ static bool ContainsOnlyUsAscii(const char* asciiValue)
     const char* iterator = asciiValue;
     while (iterator != NULL && *iterator != '\0')
     {
-        // Allow only printable ascii char
-        if (*iterator < ' ' || *iterator > '~')
+        // Allow only printable ascii char (excluding space ' ').
```
ericwol-msft, line 46: This is still not right. These chars are NOT allowed
" ( ) , / : ; < = > ? @ { }

momuno: I love how off the original code was from the specs lol.

momuno:
@wesmc7777 is updating the iothub_message.h file for documentation so I have removed that change from this commit.

iothub_client/src/iothub_message.c
momuno: Due to validation results, we should leave this as is.  Some protocols allows all characters [dec 32, dec 126], while others are limited.  It is also c2d vs d2c dependent.  I do not know of a way the code can determine which protocol is being used at this level, and I do not think it should be responsible to.  This puts the onus on the IoT Hub documentation to inform the user, and if the user uses invalid ascii characters, they will encounter an error elsewhere.

PR #1677, opened 2020-09-25
PnP API string update & e2e tests

### LINKED ISSUE #1547: Remove preview API version from C SDK

jspaith:
Enable PnP API string `2020-09-30` everywhere.  Removing the older API string and logic around optionally setting the PnP supported or GA API from device client.

We can also enable full E2E tests since IoT Hub instance in the CI gate also has this.  Tests simply verify that ModelId is successfully parsed by IoT Hub and put into the device's twin (as visible from service REST call).
* Only enabling for MQTT and MQTT_WS because this is only protocol PnP supports on C.
* There's not a natural existing test to put this into.  I didn't want to create a new ModelId-test-e2e-exe because that will create even more devices and load during a CI run.
  The closest test is the twin test, since it has machinery to retrieve the twin from service side which is needed for ModelId checking.
* We don't need to test other PnP functionality from core IoT Hub SDK - e.g. telemetry messages are stamped with componentName - as Hub is entirely responsible for this.  

Fixes #1547


testtools/iothub_test/src/iothub_account.c
```diff
@@ -976,6 +976,19 @@ IOTHUB_PROVISIONED_DEVICE* IoTHubAccount_GetX509Device(IOTHUB_ACCOUNT_INFO_HANDL
     }
 }
 
+IOTHUB_PROVISIONED_DEVICE* IoTHubAccount_GetDevice(IOTHUB_ACCOUNT_INFO_HANDLE acctHandle, IOTHUB_ACCOUNT_AUTH_METHOD accountAuthMethod)
+{
+    if (accountAuthMethod == IOTHUB_ACCOUNT_AUTH_X509)
+    {
+        return IoTHubAccount_GetX509Device(acctHandle);
+    }
+    else
```
ericwol-msft, line 985:
```suggestion
    else if (accountAuthMethod  == IOTHUB_ACCOUNT_AUTH_CONNSTRING)
```

testtools/iothub_test/src/iothub_account.c
```diff
@@ -976,6 +976,19 @@ IOTHUB_PROVISIONED_DEVICE* IoTHubAccount_GetX509Device(IOTHUB_ACCOUNT_INFO_HANDL
...
+    {
+        return IoTHubAccount_GetSASDevice(acctHandle);
+    }
+}
```
ericwol-msft, line 989: return NULL;

ericwol-msft (approved):

testtools/iothub_test/src/iothub_account.c
ericwol-msft: please note that IoTHubAccount_GetX509Device() and IoTHubAccount_GetSASDevice() can return NULL....but it looks like our code does not handle this.

testtools/iothub_test/src/iothub_account.c
ericwol-msft: Is this valid for SAS?

ericwol-msft (approved):

testtools/iothub_test/src/iothub_account.c
jspaith: Making change as suggested along with the change above with explict test for ==.  

The test basically use an "exception" based approach where if anything goes wrong, we kick an `ASSERT` right away.  We don't check against NULL because the test wouldn't have got to that state anyway.  The test init in this instance does a `dt_e2e_init` and if this doesn't setup devices we won't execute any further tests.

testtools/iothub_test/src/iothub_account.c
jspaith: Fixing the else if for explicit check.

RE question, the connection strings that we use gets generated thanks to the helper `createSASConnectionString`.  However I think this is a misnomer because this connection string includes the `;SharedAccessKey=` and not `SharedAccessSignature`.  Leaving as-is for now.

Issue #1625, opened 2020-08-13
Renew SAS token without disconnect using AMQP


MartinTJDK:
We have been using MQTT (WebSocket) and SAS token with an expiry time of 1 hour.
The SAS token is generated and fetched from a server, and is refreshed/requested each hour.
To avoid the need for disconnect and the reconnect, which MQTT requires in order to accept a new SAS token, we are trying to use AMQP.
But it is difficult to find out how we can force the revalidation of the connection using a new SAS token.
Can anyone point me in the right direction?
Thanks.

ewertons:
Hi @MartinTJDK ,
the Azure IoT C SDK does not have logic to request a new SAS token if the one provided expires.
Currently the only option is to destroy the device client and create a new one with the new SAS token.

ewertons:
The SAS token is held by the "Authorization Module":
https://github.com/Azure/azure-iot-sdk-c/blob/41ffe848773c03985df395a4ca29d986b142033b/iothub_client/src/iothub_client_core_ll.c#L829

The validation of tokens is done in this part of the code:
https://github.com/Azure/azure-iot-sdk-c/blob/41ffe848773c03985df395a4ca29d986b142033b/iothub_client/src/iothub_client_authorization.c#L573

That function (IoTHubClient_Auth_Is_SasToken_Valid) gets invoked in this module, which pertains to the AMQP transport:
https://github.com/Azure/azure-iot-sdk-c/blob/41ffe848773c03985df395a4ca29d986b142033b/iothub_client/src/iothubtransport_amqp_cbs_auth.c#L278 

Here you would need to make changes to `create_and_put_SAS_token_to_cbs` so it doesn't fail if the result is SAS_TOKEN_STATUS_INVALID. The block that handles sas tokens

https://github.com/Azure/azure-iot-sdk-c/blob/41ffe848773c03985df395a4ca29d986b142033b/iothub_client/src/iothubtransport_amqp_cbs_auth.c#L278

should be similar to the one that handles device key and auto-generation of sas tokens

https://github.com/Azure/azure-iot-sdk-c/blob/41ffe848773c03985df395a4ca29d986b142033b/iothub_client/src/iothubtransport_amqp_cbs_auth.c#L264

Next, there should be changes to `IoTHubClient_Auth_Get_SasToken` to invoke a callback (or in your case simpler, read from an ENV VAR?) to get a new SAS token from user application. The change in that function would probably happen in this block:
[...]

PR #1682, opened 2020-09-30
Added support for fetch of SAS token via callback

MartinTJDK:
#1625 

When using AMQP and SAS token authentication, it is not possible to inject updated SAS token in SDK without disconnecting.

Added possibility to assign a "request_sas_token_callback", which supplies Azure IoT SDK with an updated SAS token.
When a new SAS token is generated, or fetched from external source, the Azure IoT SDK is triggered to fetch the SAS token via the "IoTHubDeviceClient_LL_SetOption", "OPTION_SAS_TOKEN_LIFETIME". When the "request_sas_token_callback" callback is executed by the Azure IoT SDK, the "OPTION_SAS_TOKEN_LIFETIME" is set to the high value again. Next time a new SAS token is ready, the "OPTION_SAS_TOKEN_LIFETIME" is modified again.

### DIFF
--- a/iothub_client/src/iothubtransport_amqp_cbs_auth.c
+++ b/iothub_client/src/iothubtransport_amqp_cbs_auth.c
@@ -629,7 +629,9 @@ void authentication_do_work(AUTHENTICATION_HANDLE authentication_handle)
         {
             // Codes_SRS_IOTHUBTRANSPORT_AMQP_AUTH_09_040: [If `instance->state` is AUTHENTICATION_STATE_STARTED and user-provided SAS token was used, authentication_do_work() shall return]
             IOTHUB_CREDENTIAL_TYPE cred_type = IoTHubClient_Auth_Get_Credential_Type(instance->authorization_module);
-            if (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_KEY || cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_AUTH)
+            if ( (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_KEY) ||
+                 (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_AUTH) ||
+                 ((cred_type == IOTHUB_CREDENTIAL_TYPE_SAS_TOKEN) && IoTHubClient_Auth_Is_SasToken_Update_Supported(instance->authorization_module)) )
             {
                 // Codes_SRS_IOTHUBTRANSPORT_AMQP_AUTH_09_039: [If `instance->state` is AUTHENTICATION_STATE_STARTED and device keys were used, authentication_do_work() shall only verify the SAS token refresh time]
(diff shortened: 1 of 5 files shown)


iothub_client/src/iothubtransport_amqp_cbs_auth.c
```diff
@@ -629,7 +629,9 @@ void authentication_do_work(AUTHENTICATION_HANDLE authentication_handle)
         {
             // Codes_SRS_IOTHUBTRANSPORT_AMQP_AUTH_09_040: [If `instance->state` is AUTHENTICATION_STATE_STARTED and user-provided SAS token was used, authentication_do_work() shall return]
             IOTHUB_CREDENTIAL_TYPE cred_type = IoTHubClient_Auth_Get_Credential_Type(instance->authorization_module);
-            if (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_KEY || cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_AUTH)
+            if ( (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_KEY) ||
+                 (cred_type == IOTHUB_CREDENTIAL_TYPE_DEVICE_AUTH) ||
+                 ((cred_type == IOTHUB_CREDENTIAL_TYPE_SAS_TOKEN) && IoTHubClient_Auth_Is_SasToken_Update_Supported(instance->authorization_module)) )
```
ewertons, line 634: If I'm getting this right, this would prevent users from passing just one SAS token in the client config and no callback. That would be a breaking change.
I'll post an alternative briefly. Feel free to reply back with details if you believe my understanding is incorrect.

ewertons (commented):
So far this PR seems very clean and nice. I just need to review a few more points. Please take a look on the question I left.

iothub_client/src/iothubtransport_amqp_cbs_auth.c
MartinTJDK: I don't think so.
"IoTHubClient_Auth_Is_SasToken_Update_Supported(instance->authorization_module)" will check, if a callback is set.
If not, the "iothubtransport_amqp_cbs_auth.c" will never start any SAS token generation/update or request, but use the SAS token, which is supplied once.

ericwol-msft:
Are your still interested in this PR? Did you use a different SDK (C#)?

CIPop:
@MartinTJDK I would propose a slightly different approach for this feature. We already have a security provider within the custom_hsm construct. Unfortunately, we don't have one that allows the application to provide a SAS token. (We currently have one for each TPM, X509 Client Auth and Shared Access Key.)

* IoT Hub sample using HSM (no DPS, unfortunately confusingly within the provisioning_client folder...): 
https://github.com/Azure/azure-iot-sdk-c/tree/master/provisioning_client/samples/iothub_client_sample_hsm

* Custom HSM sample:
https://github.com/Azure/azure-iot-sdk-c/tree/master/provisioning_client/samples/custom_hsm_example

Since this is already the established way to expand authentication, we could add another interface that would allow the application to provide the SAS token, expiry information and registration_id (which must be the same with device_id in this scenario).

PR #1696, opened 2020-10-15
Fix mqtt disconnect

### LINKED ISSUE #1666: mqtt_client_dowork crash due to holding reference to destroyed xio
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

Mbed OS: 6.3.0
Toolchain: Arm Compiler 6.13

Arm Mbed ported Azure example:
https://github.com/ARMmbed/mbed-os-example-for-azure

**SDK Version (Please Give Commit SHA if Manually Compiling)**

**azure-iot-sdk-c**
LTS_07_2020_Ref01

**azure-umqtt-c**
2020-07-19

**Protocol**

MQTT

**Describe the Bug**

`mqtt_client.c` can hold `xio` object, which has been destroyed in `iothubtransportmqtt.c`. Please follow the sequence to replay it:

1. On mqtt disconnect, `DisconnectFromClient` gets called and `xio` object is destroyed:

    https://github.com/Azure/azure-iot-sdk-c/blob/2bff372a9b22b54d3d4bc9a3e74e349f68fcde46/iothub_client/src/iothubtransport_mqtt_common.c#L2407-L2412

    https://github.com/Azure/azure-iot-sdk-c/blob/2bff372a9b22b54d3d4bc9a3e74e349f68fcde46/iothub_client/src/iothubtransport_mqtt_common.c#L1860-L1864

    https://github.com/Azure/azure-iot-sdk-c/blob/2bff372a9b22b54d3d4bc9a3e74e349f68fcde46/iothub_client/src/iothubtransport_mqtt_common.c#L298-L302

1. `IoTHubTransport_MQTT_Common_DoWork` still rolls, and invokes `mqtt_client_dowork`:

    https://github.com/Azure/azure-iot-sdk-c/blob/2bff372a9b22b54d3d4bc9a3e74e349f68fcde46/iothub_client/src/iothubtransport_mqtt_common.c#L3131-L3134
    https://github.com/Azure/azure-iot-sdk-c/blob/2bff372a9b22b54d3d4bc9a3e74e349f68fcde46/iothub_client/src/iothubtransport_mqtt_common.c#L3228-L3229

1. In `mqtt_client_dowork`, program crash in `xio_dowork` because `mqtt_client->xioHandle` has been destroyed by above:

    https://github.com/Azure/azure-umqtt-c/blob/165f6f52f8b2aec54a28ea5ede41736589726cd0/src/mqtt_client.c#L1280-L1285

    https://github.com/Azure/azure-umqtt-c/blob/165f6f52f8b2aec54a28ea5ede41736589726cd0/src/mqtt_client.c#L1293-L1302

danewalton:
Fixes #1666

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -301,6 +301,7 @@ static void DestroyXioTransport(PMQTTTRANSPORT_HANDLE_DATA transport_data)
 {
     xio_destroy(transport_data->xioTransport);
     transport_data->xioTransport = NULL;
+    mqtt_client_close_xio(transport_data->mqttClient);
 }
 
 static void set_saved_tls_options(PMQTTTRANSPORT_HANDLE_DATA transport, OPTIONHANDLER_HANDLE new_options)
@@ -1893,12 +1894,6 @@ static void DisconnectFromClient(PMQTTTRANSPORT_HANDLE_DATA transport_data)
                 disconnect_ctr++;
                 ThreadAPI_Sleep(50);
             } while ((disconnect_ctr < MAX_DISCONNECT_VALUE) && (transport_data->disconnect_recv_flag == 0));
-
-            // Something went wrong sending the disconnect | Force the closing of the xio
-            if (disconnect_ctr >= MAX_DISCONNECT_VALUE)
-            {
-              mqtt_client_close_xio(transport_data->mqttClient);
-            }
         }
         DestroyXioTransport(transport_data);
 
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -300,8 +300,8 @@ static void free_proxy_data(MQTTTRANSPORT_HANDLE_DATA* mqtt_transport_instance)
 static void DestroyXioTransport(PMQTTTRANSPORT_HANDLE_DATA transport_data)
 {
     xio_destroy(transport_data->xioTransport);
+    mqtt_client_clear_xio(transport_data->mqttClient);
     transport_data->xioTransport = NULL;
-    mqtt_client_close_xio(transport_data->mqttClient);
 }
 
 static void set_saved_tls_options(PMQTTTRANSPORT_HANDLE_DATA transport, OPTIONHANDLER_HANDLE new_options)
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -318,6 +318,7 @@ static void free_transport_handle_data(MQTTTRANSPORT_HANDLE_DATA* transport_data
     if (transport_data->mqttClient != NULL)
     {
         mqtt_client_deinit(transport_data->mqttClient);
+        transport_data->mqttClient = NULL;
     }
 
     if (transport_data->retry_control_handle != NULL)
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -299,8 +299,8 @@ static void free_proxy_data(MQTTTRANSPORT_HANDLE_DATA* mqtt_transport_instance)
 // Destroys xio transport associated with MQTT handle and resets appropriate state
 static void DestroyXioTransport(PMQTTTRANSPORT_HANDLE_DATA transport_data)
 {
-    xio_destroy(transport_data->xioTransport);
     mqtt_client_clear_xio(transport_data->mqttClient);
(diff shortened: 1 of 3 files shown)


iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -317,6 +318,7 @@ static void free_transport_handle_data(MQTTTRANSPORT_HANDLE_DATA* transport_data
     if (transport_data->mqttClient != NULL)
     {
         mqtt_client_deinit(transport_data->mqttClient);
+        transport_data->mqttClient = NULL;
```
danewalton, line 321: @ewertons @jspaith can you guys verify this is correct? I believe we should be clearing it to `NULL` in the case that the `mqtt_client_deinit()` frees it.

jspaith: This isn't actually needed and we should remove this line from PR.  

At bottom of this function we have a `free(transport_data);` and nothing in this function will be mucking with the `mqttClient`.  We only really NULL stuff out if its part of a structure that has an ongoing life so later code knows not to mess with it.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -1171,6 +1171,7 @@ static void setup_initialize_reconnection_mocks()
         .CopyOutArgumentBuffer_retry_action(&retry_action, sizeof(retry_action));
     STRICT_EXPECTED_CALL(tickcounter_get_current_ms(IGNORED_PTR_ARG, IGNORED_PTR_ARG));
     STRICT_EXPECTED_CALL(xio_retrieveoptions(IGNORED_PTR_ARG));
+    STRICT_EXPECTED_CALL(mqtt_client_clear_xio(IGNORED_PTR_ARG));
```
jspaith, line 1174: General comment I'm putting just on this line...

This issue of was hit in field with #1666.  We should add a new test that repros that this as the customer ran into it so that we know if we fall into that case in future (and especially as code around us changes) we'll still be protected.

I get these are a fair amount of work.  I can help with advise on best way to model this if needed.

jspaith (approved):

iothub_client/src/iothubtransport_mqtt_common.c
danewalton: So the issue is, line 320 here will free the mqtt client but it will not set the pointer to `NULL`. The code then proceeds down to line 347 where it calls `DestroyXioTransport(transport_data);` which clears via `mqtt_client_clear_xio(transport_data->mqttClient);`. Since the pointer to the `mqtt_client` is not `NULL`, it derefs and segfaults.

Without it, here is a callstack which segfaults:
image

jspaith: Ah - I didn't see that.  Please don't segfault:).  The code as you have it makes sense then.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
danewalton: ahhhh crap. yea that would be ideal. I'll link base with you to figure out how we can go about doing that.

danewalton: added. as discussed, this new unit test UNIT tests the disconnect scenario. it does not e2e test the scenario from the gh issue.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -2198,6 +2203,54 @@ TEST_FUNCTION(IoTHubTransport_MQTT_Common_Destroy_One_Message_Ack_succeeds)
...
+
+    DList_InsertTailList(config.waitingToSend, &(message1.entry));
+
+    TRANSPORT_LL_HANDLE handle = setup_iothub_mqtt_connection(&config);
+
+    IoTHubTransport_MQTT_Common_DoWork(handle);
+    umock_c_reset_all_calls();
+
+    STRICT_EXPECTED_CALL(mqtt_client_disconnect(IGNORED_PTR_ARG, IGNORED_PTR_ARG, IGNORED_PTR_ARG));
+    for(int i = 0; i < 50; ++i)
```
jspaith, line 2248: Nit: 50 is a magic # hard-coded in tests.  I get this isn't part of public mqtt_common.c (nor should it be) but test should #define this at top of file in any case.

jspaith (approved):

PR #1699, opened 2020-10-16
Fix #1698 and refactor multiblock upload code

### LINKED ISSUE #1698: UploadToBlob callback internals cannot assume callback will fill in values
In `Blob_UploadMultipleBlocksFromSasUri` we have code below (I'm redacting occasional line with a // ...)

```c
unsigned char const * source; /* data set by getDataCallbackEx */
size_t size; /* source size set by getDataCallbackEx */
//..
	getDataReturnValue = getDataCallbackEx(FILE_UPLOAD_OK, &source, &size, context);
	if (getDataReturnValue == IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_ABORT)
	{
		// ...
		result = BLOB_ABORTED;
	}
	else if (source == NULL || size == 0)
```
This is a problem.  The `getDataCallbackEx` is a callback function that the application itself writes and we cannot assume that it will properly fill in the `size` and `source` values.  We should default these to 0 and NULL before invoking callback.  This will at least get us out of this loop so we won't be calling this poorly behaving function anymore.

I don't think there's back-compat concerns in fixing this even though it's around a user-code callout point.  If an app is not setting these today and not returning `IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_ABORT`, then the rest of this function is going to  use garbage size/source values and potentially crash.

jspaith:
Part1
First, fix #1698 which is explained in depth there.  But since there's extra changes in this PR the only part to this is initializing some variables.

```c
    unsigned char const * source = NULL; /* data set by getDataCallbackEx */
    size_t size = 0; /* source size set by getDataCallbackEx */
```

Part2 
The code is originally written though needed some cleanup.  `Blob_UploadMultipleBlocksFromSasUri` is almost 250 lines long and got to 15 levels of indentation.  

Fixed so that there's not as deep nesting of the else and created two helper functions, the most important being `InvokeUserCallbackAndSendBlobs` where we could put the loop.

Part3
Fix up the UT.  The _required_ changes were just fixing up the UT call list.  

The UT were also using the old-style hard-coded index of tests to skip in failure mode.  Replace with `umock_c_negative_tests_can_call_fail(i))`.

### DIFF
--- a/iothub_client/src/blob.c
+++ b/iothub_client/src/blob.c
@@ -293,7 +293,6 @@ static BLOB_RESULT SendBlockIdList(HTTPAPIEX_HANDLE httpApiExHandle, const char*
     }
 
     return result;
-
 }
 
 
--- a/iothub_client/src/blob.c
+++ b/iothub_client/src/blob.c
@@ -13,6 +13,10 @@
 #include "azure_c_shared_utility/azure_base64.h"
 #include "azure_c_shared_utility/shared_util_options.h"
 
+static const char blockListXmlBegin[]  = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\r\n<BlockList>";
+static const char blockListXmlEnd[] = "</BlockList>";
+static const char blockListUriMarker[] = "&comp=blocklist";
+
 BLOB_RESULT Blob_UploadBlock(
         HTTPAPIEX_HANDLE httpApiExHandle,
         const char* relativePath,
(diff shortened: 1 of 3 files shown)


iothub_client/src/blob.c
```diff
@@ -131,257 +135,263 @@ BLOB_RESULT Blob_UploadBlock(
...
+                            blockID,
+                            blockIDList,
+                            httpStatus,
+                            httpResponse);
+
+                    BUFFER_delete(requestContent);
+                }
+
+                /*Codes_SRS_BLOB_02_026: [ Otherwise, if HTTP response code is >=300 then Blob_UploadMultipleBlocksFromSasUri shall succeed and return BLOB_OK. ]*/
+                if (result != BLOB_OK || *httpStatus >= 300)
```
ewertons, line 211: Just to make sure, I think it would be safer for us to make sure we check if httpStatus is within the 2xx range. 
LSS, we should consider httpStatus < 200 also an error, since we do not have code to handle 1xx.
LSShorter, 

```suggestion
                if (result != BLOB_OK || *httpStatus < 200 || *httpStatus >= 300)
```

iothub_client/src/blob.c
```diff
@@ -131,257 +135,263 @@ BLOB_RESULT Blob_UploadBlock(
...
+BLOB_RESULT Blob_UploadMultipleBlocksFromSasUri(const char* SASURI, IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_CALLBACK_EX getDataCallbackEx, void* context, unsigned int* httpStatus, BUFFER_HANDLE httpResponse, const char* certificates, HTTP_PROXY_OPTIONS *proxyOptions)
+{
+    BLOB_RESULT result;
+    const char* hostnameBegin;
+    STRING_HANDLE blockIDList = NULL;
+    HTTPAPIEX_HANDLE httpApiExHandle = NULL;
+    char* hostname = NULL;
+    
+    /*Codes_SRS_BLOB_02_001: [ If SASURI is NULL then Blob_UploadMultipleBlocksFromSasUri shall fail and return BLOB_INVALID_ARG. ]*/
+    if (SASURI == NULL)
```
ewertons, line 308: We could reduce the code a little more by joining these two if blocks and printing one single error log with the values of the variables  (%p).

iothub_client/src/blob.c
```diff
@@ -131,257 +135,263 @@ BLOB_RESULT Blob_UploadBlock(
...
+                {
+                    /*Codes_SRS_BLOB_02_033: [ If any previous operation that doesn't have an explicit failure description fails then Blob_UploadMultipleBlocksFromSasUri shall fail and return BLOB_ERROR ]*/
+                    LogError("failed to STRING_construct");
+                    result = BLOB_HTTP_ERROR;
+                }
+                else if ((result = InvokeUserCallbackAndSendBlobs(httpApiExHandle, relativePath, blockIDList, getDataCallbackEx, context, httpStatus, httpResponse)) != BLOB_OK)
+                {
+                   ; 
+                }
+                else if (*httpStatus < 300)
```
ewertons, line 382: It should be:

```suggestion
                else if (*httpStatus >= 200 && *httpStatus < 300)
```

iothub_client/src/blob.c
```diff
@@ -131,257 +135,263 @@ BLOB_RESULT Blob_UploadBlock(
...
+                /*Codes_SRS_BLOB_02_028: [ Blob_UploadMultipleBlocksFromSasUri shall construct an XML string with the following content: ]*/
+                else if ((blockIDList = STRING_construct(blockListXmlBegin)) == NULL)
+                {
+                    /*Codes_SRS_BLOB_02_033: [ If any previous operation that doesn't have an explicit failure description fails then Blob_UploadMultipleBlocksFromSasUri shall fail and return BLOB_ERROR ]*/
+                    LogError("failed to STRING_construct");
+                    result = BLOB_HTTP_ERROR;
+                }
+                else if ((result = InvokeUserCallbackAndSendBlobs(httpApiExHandle, relativePath, blockIDList, getDataCallbackEx, context, httpStatus, httpResponse)) != BLOB_OK)
+                {
+                   ;
```
ewertons, line 380: We should have a log trace here for the error. 
We need to reach the million-log-trace mark every time we run a simple sample, otherwise we fall below SLA.

ewertons (approved):
Seems great, thanks for refactoring! 
I left some comments though, please address before merging.

iothub_client/src/blob.c
jspaith: I'd actually rather not make this change, or at least not without more research with the Hub and potentially Storage teams which would be a separate issue.  I don't want any functional changes in this (other than initializing the values to NULL, which would prevent a crash in degenerate caes).

iothub_client/src/blob.c
jspaith: Fixing.

iothub_client/src/blob.c
jspaith: Fixed.

iothub_client/src/blob.c
jspaith: Similar feedback to above, I'd rather not make functional changes that aren't strictly required in this refactor.

iothub_client/src/blob.c
jspaith: Per offline discussion, I'd rather keep the refactor isolated over any sort of functional changes even if this is righteous.  Resolving for now.

PR #1703, opened 2020-10-21
Fix for lack of throttling on registration request of DPS client.

ewertons:
This is regarding IcM ticket #209445257.

The Device Provisioning Client of the Azure IoT C SDK handles 429 and retry-after parameter for get-status operation, but not for the initial registration request. That is a bug, as shown in the logs of the IcM ticket above.

According to the DPS service team, 
"DPS has separate throttling limits for the device registration method (PUT) and the polling method (GET)
[according to] https://docs.microsoft.com/en-us/azure/iot-dps/about-iot-dps#quotas
[and] Both can and will be throttled"

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/provisioning_client/src/prov_device_ll_client.c
+++ b/provisioning_client/src/prov_device_ll_client.c
@@ -92,7 +92,7 @@ typedef struct PROV_INSTANCE_INFO_TAG
 
     TICK_COUNTER_HANDLE tick_counter;
 
-    tickcounter_ms_t status_throttle;
+    tickcounter_ms_t last_send_time;
     tickcounter_ms_t timeout_value;
     uint32_t retry_after_value;
 
@@ -930,8 +930,8 @@ PROV_DEVICE_LL_HANDLE Prov_Device_LL_Create(const char* uri, const char* id_scop
                 else
                 {
                     // Ensure that we are passed the throttling time and send on the first send
-                    (void)tickcounter_get_current_ms(result->tick_counter, &result->status_throttle);
-                    result->status_throttle += (result->retry_after_value * 1000);
+                    (void)tickcounter_get_current_ms(result->tick_counter, &result->last_send_time);
+                    result->last_send_time += (result->retry_after_value * 1000);
                 }
             }
         }
@@ -1118,29 +1118,11 @@ void Prov_Device_LL_DoWork(PROV_DEVICE_LL_HANDLE handle)
         }
         if (prov_info->is_connected || prov_info->prov_state == CLIENT_STATE_ERROR)
         {
+            tickcounter_ms_t current_time = 0;
+
             switch (prov_info->prov_state)
             {
                 case CLIENT_STATE_REGISTER_SEND:
-                    /* Codes_SRS_PROV_CLIENT_07_013: [ CLIENT_STATE_REGISTER_SEND which shall construct an initial call to the service with endorsement information ] */
-                    if (prov_info->prov_transport_protocol->prov_transport_register(prov_info->transport_handle, prov_transport_process_json_reply, prov_transport_create_json_payload, prov_info) != 0)
-                    {
-                        LogError("Failure registering device");
-                        if (prov_info->error_reason == PROV_DEVICE_RESULT_OK)
-                        {
-                            prov_info->error_reason = PROV_DEVICE_RESULT_TRANSPORT;
-                        }
-                        prov_info->prov_state = CLIENT_STATE_ERROR;
-                    }
-                    else
-                    {
-                        (void)tickcounter_get_current_ms(prov_info->tick_counter, &prov_info->timeout_value);
-                        prov_info->prov_state = CLIENT_STATE_REGISTER_SENT;
-                    }
-                    break;
-
-                case CLIENT_STATE_STATUS_SEND:
-                {
-                    tickcounter_ms_t current_time = 0;
                     if (tickcounter_get_current_ms(prov_info->tick_counter, &current_time) != 0)
                     {
                         LogError("Failure getting the current time");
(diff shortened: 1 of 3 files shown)


provisioning_client/src/prov_transport_mqtt_common.c
```diff
@@ -319,7 +319,7 @@ static void mqtt_notification_callback(MQTT_MESSAGE_HANDLE handle, void* user_ct
             size_t status_pos = strlen(MQTT_TOPIC_STATUS_PREFIX);
             if (memcmp(MQTT_TOPIC_STATUS_PREFIX, topic_resp, status_pos) == 0)
             {
-                // If the status code is > 429 then this is a transient error
+                // If the status code is >= 429 then this is a transient error
                 long status_code = atol(topic_resp + status_pos);
                 if (status_code >= PROV_STATUS_CODE_TRANSIENT_ERROR)
```
CIPop, line 324: I don't think this is true.

From embedded SDK: (https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/inc/azure/iot/az_iot_common.h) 

AZ_NODISCARD AZ_INLINE bool az_iot_status_retriable(az_iot_status status)
{
  return ((status == AZ_IOT_STATUS_THROTTLED) || (status == AZ_IOT_STATUS_SERVER_ERROR));
}

The other errors appear to be non-retriable.

provisioning_client/src/prov_device_ll_client.c
```diff
@@ -1118,29 +1118,43 @@ void Prov_Device_LL_DoWork(PROV_DEVICE_LL_HANDLE handle)
         }
         if (prov_info->is_connected || prov_info->prov_state == CLIENT_STATE_ERROR)
         {
+            tickcounter_ms_t current_time = 0;
```
CIPop, line 1121: Shouldn't we get the real time every time?

provisioning_client/src/prov_device_ll_client.c
```diff
@@ -930,8 +930,8 @@ PROV_DEVICE_LL_HANDLE Prov_Device_LL_Create(const char* uri, const char* id_scop
                 else
                 {
                     // Ensure that we are passed the throttling time and send on the first send
-                    (void)tickcounter_get_current_ms(result->tick_counter, &result->status_throttle);
-                    result->status_throttle += (result->retry_after_value * 1000);
+                    (void)tickcounter_get_current_ms(result->tick_counter, &result->last_send_time);
+                    result->last_send_time += (result->retry_after_value * 1000);
```
CIPop, line 934: Nit: likely safer and less CPU to store the times in seconds.

provisioning_client/src/prov_device_ll_client.c
```diff
@@ -1118,29 +1118,43 @@ void Prov_Device_LL_DoWork(PROV_DEVICE_LL_HANDLE handle)
...
-                    /* Codes_SRS_PROV_CLIENT_07_013: [ CLIENT_STATE_REGISTER_SEND which shall construct an initial call to the service with endorsement information ] */
-                    if (prov_info->prov_transport_protocol->prov_transport_register(prov_info->transport_handle, prov_transport_process_json_reply, prov_transport_create_json_payload, prov_info) != 0)
+                    if (tickcounter_get_current_ms(prov_info->tick_counter, &current_time) != 0)
                     {
-                        LogError("Failure registering device");
-                        if (prov_info->error_reason == PROV_DEVICE_RESULT_OK)
-                        {
-                            prov_info->error_reason = PROV_DEVICE_RESULT_TRANSPORT;
-                        }
+                        LogError("Failure getting the current time");
```
CIPop, line 1128: Re Nit above: we could adjust `current_time` to be in seconds right after getting the tick ms time.

provisioning_client/src/prov_device_ll_client.c
```diff
@@ -92,7 +92,7 @@ typedef struct PROV_INSTANCE_INFO_TAG
 
     TICK_COUNTER_HANDLE tick_counter;
 
-    tickcounter_ms_t status_throttle;
+    tickcounter_ms_t last_send_time;
```
ericwol-msft, line 95:
```suggestion
    tickcounter_ms_t last_send_time_ms;
```

CIPop (approved):
One small worry about what we consider retriable errors which appears to be different than the original spec and Hub/DPS error code spec.

provisioning_client/src/prov_device_ll_client.c
ericwol-msft: it looks like it is also used in CLIENT_STATE_STATUS_SEND... so it will be zero?

provisioning_client/src/prov_transport_mqtt_common.c
ewertons: Notice that this code keeps trying to reconnect no matter what. 
If we go for fixing the entire code to distinguish between retryable and non-retryable errors, there will be a lot more changes that this. 
This is fixing a very specific issue.

ewertons: But if we take the priorization away, I do agree with you.

provisioning_client/src/prov_device_ll_client.c
ewertons: No, I had to move it here because if you use the same var name in different switch cases the compiler (at least VS2019) will fail. We only get the time in 2 switch cases.

provisioning_client/src/prov_device_ll_client.c
ewertons: I agree with you. 
Not sure why this has been done like that, specially given that the code translate to seconds anyway when it uses the variable:
if ((current_time - prov_info->last_send_time) / 1000 > prov_info->retry_after_value)

And that happens on DoWork!!! (running at least every 100 milliseconds).

I'll make an exception to my focus on fixing the specific issue and change this to seconds. 

Thanks @CIPop !

ewertons: So I was about to agree with you, but...
tickcounter_get_current_ms only returns time in milliseconds.
There is no other tickcounter function for getting time in seconds.
The code needs to keep getting the time every time it calculates a time span (specially in DoWork). So it makes more sense to change the variable that holds the retry-after value to milliseconds, so we don't need to keep recalculating from milliseconds to seconds to compare.
LST, instead of going all the way with seconds, it's more advisable to go all the way with milliseconds.

provisioning_client/src/prov_device_ll_client.c
ewertons: Keeping it as milliseconds and changing retry_after_Value to milliseconds creates better solution.

provisioning_client/src/prov_device_ll_client.c
ewertons: Approved! Thanks

PR #1706, opened 2020-10-21
Update Root CA for Event Hub connections during testing

danewalton:


testtools/iothub_test/src/iothubtest.c
```diff
@@ -1130,7 +1155,7 @@ IOTHUB_TEST_CLIENT_RESULT IoTHubTest_ListenForEvent(IOTHUB_TEST_HANDLE devhubHan
                 else
                 {
 #ifdef SET_TRUSTED_CERT
-                    xio_setoption(tls_io, OPTION_TRUSTED_CERT, certificates);
+                    xio_setoption(tls_io, OPTION_TRUSTED_CERT, test_service_cert);
 #endif // SET_TRUSTED_CERT
```
hihigupt, line 1159: In test can we depend upon OS cert store? If yes, this can be done by simply removing this xio_setoption.

If I remember: for openssl and mbedtls when we do not configure them with personal root CA list they fallback to OS cert store. 

Is it possible to test that?

jspaith: Yes, the SET_TRUSTED_CERT which is controled by the cmake flag `use_sample_trusted_cert` does the right things here.  It may be set other places but from our main cmake lists 

```
if ((WIN32 AND ${use_openssl}) OR ${use_wolfssl} OR ${use_mbedtls} OR ${use_bearssl})
    option(use_sample_trusted_cert "Set flag in samples to use SDK's built-in CA as TrustedCerts" ON)
else()
    option(use_sample_trusted_cert "Set flag in samples to use SDK's built-in CA as TrustedCerts" OFF)
endif()
```

danewalton: I believe wolfssl and bearssl require you to set them and don't use cert store

jspaith (approved):

PR #1707, opened 2020-10-21
Make uploadToBlob E2E tests better respect throttling

jspaith:
We are at risk of false positive errors in our e2e tests for UploadToBlob.  At issue is an [S1 hub's thresholds](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling) can easily be exceeded by our checkin gates spinning up so many of these tests.  

The tests run in <1 second and many of them don't wait between runs.  Furthermore, there's a multi-threaded test that spins 5 simultaneous worker threads.  In an experiment with extra instrumentation for timing, one test by itself created 10 connections in a 10 second span.

This change (a) makes sure ALL tests wait 3 seconds between runs and (b) moves the simultaneous thread creation from 5 to 3.  This will mean the max # of connections a given test EXE can create in busiest 10 second period will be 5.

Will monitor progress on this on gate side and potentially further extend out sleep (which is now in a central location) as needed.

### DIFF
--- a/iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
+++ b/iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
@@ -199,7 +199,7 @@ static void sleep_between_upload_blob_e2e_tests(void)
     //
     // On gate runs, we have many instances of this test executable running at the same time and we cannot orchestrate
     // them.  Most individual testcases take <1 second to run.  Without a sleep, this amount of traffic
-    // will end up going over throttling maximums and causing false positive test case failures.
+    // will end up going over throttling maximums and causing test case failures.
     //
     LogInfo("Invoking sleep for %d milliseconds after test case", TEST_SLEEP_BETWEEN_UPLOAD_TO_BLOB_E2E_TESTS_MS);
     ThreadAPI_Sleep(TEST_SLEEP_BETWEEN_UPLOAD_TO_BLOB_E2E_TESTS_MS);
@@ -244,7 +244,7 @@ void e2e_uploadtoblob_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol, IOTHUB_ACC
     poll_for_upload_completion(&uploadToBlobStatus);
     check_upload_result(uploadToBlobStatus);
 
-    // We need to sleep in any event to avoid triggering IoT Hub upload trheshold limits in our E2E tests.  We need to do this
+    // We need to sleep in any event to avoid triggering IoT Hub upload threshold limits in our E2E tests.  We need to do this
     // before the client destroy - and not test runs themselves which would've been better - because of https://github.com/Azure/azure-iot-sdk-c/issues/1705.
     sleep_between_upload_blob_e2e_tests();
     IoTHubClient_Destroy(iotHubClientHandle);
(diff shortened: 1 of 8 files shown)


danewalton:
Pleading ignorance here. Is it a possible issue that we invoke ctest with multiple cores? Could there be multiple instances of upload to blob tests running at the same time which helps cause this issue?

jspaith:
> Pleading ignorance here. Is it a possible issue that we invoke ctest with multiple cores? Could there be multiple instances of upload to blob tests running at the same time which helps cause this issue?

Yup.  I looked at gates and how many run(s) each does.  The Ubuntu/clang are particularly hard on the threshold because they always have all 4 tests running really close to each other if not same time because it's just order of cmake tests.
```
Windows x86 - 1 run
Clang - 4 runs
Ubuntu16  - 4 runs
Bear - 1 run
Windowsx64  1 run
Xcodenative - 1 run
Cares - 1 run
```

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
```diff
@@ -191,6 +191,21 @@ static void check_upload_result(UPLOADTOBLOB_CALLBACK_STATUS uploadToBlobStatus)
...
 
+static void sleep_between_upload_blob_e2e_tests(void)
+{
+    // We need a Sleep() between each E2E test.  Current (as of October, 2020) throttling
+    // rules limit an S1 hub to 1.67 upload initiations per second for the entire IoT Hub (not just per device).
+    // https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling
+    //
+    // On gate runs, we have many instances of this test executable running at the same time and we cannot orchestrate
+    // them.  Most individual testcases take <1 second to run.  Without a sleep, this amount of traffic
+    // will end up going over throttling maximums and causing false positive test case failures.
```
danewalton, line 202: false negative?

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
```diff
@@ -228,7 +243,10 @@ void e2e_uploadtoblob_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol, IOTHUB_ACC
 
     poll_for_upload_completion(&uploadToBlobStatus);
     check_upload_result(uploadToBlobStatus);
-	ThreadAPI_Sleep(3000);
+
+    // We need to sleep in any event to avoid triggering IoT Hub upload trheshold limits in our E2E tests.  We need to do this
```
danewalton, line 247:
```suggestion
    // We need to sleep in any event to avoid triggering IoT Hub upload threshold limits in our E2E tests.  We need to do this
```

danewalton (approved):

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
jspaith: You're probably right but then we're getting into double negative territory.  I'm just removing this bit and moving onto "causing test failures".

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
jspaith: Yup.  Fixing.

ericwol-msft (approved):

Issue #1554, opened 2020-06-09
Failure to create Intermediate Device CA certificate


IndefiniteBen:
Using the certificate guide [tools/CACertificates/CACertificateOverview.md](https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md) with the bash commands fails for me. This has previously been discussed in #349 and #580. I have not tried the same process in PowerShell with OpenSSL.

#### Development Machine, OS, Compiler (and Other Relevant Toolchain Info)
Windows Subsystem for Linux Ubuntu 18.04

#### SDK Version
Clone from master

#### Describe the Bug

As mentioned in #349 I run command `./certGen.sh create_root_and_intermediate` and after the Root CA Certificate details are output to console, the creation of the Intermediate Device CA CSR fails as seen in the logs below.

#### Console Logs
```
$ ./certGen.sh create_root_and_intermediate
[Root CA Certificate details...]
Creating the Intermediate Device CA
-----------------------------------
Generating RSA private key, 4096 bit long modulus (2 primes)
.++++
...........................................................................................................................................++++
e is 65537 (0x010001)
Creating the Intermediate Device CA CSR
-----------------------------------
Signing the Intermediate Certificate with Root CA Cert
-----------------------------------
Using configuration from ./openssl_root_ca.cnf
Can't open ./certs/azure-iot-test-only.intermediate.cert.pem for reading, No such file or directory
140638769123776:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:72:fopen('./certs/azure-iot-test-only.intermediate.cert.pem','r')
140638769123776:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:79:
unable to load certificate
```

ericwol-msft:
I just tried this on my machine and it worked correctly (using WSL). What version of openSSL (mine OpenSSL 1.1.1f  31 Mar 2020)? what files are in the certs directory? Is this a clean directory?

IndefiniteBen:
* Master branch of SDK
* Yes it is a clean directory. I also tried in both a (windows) user directory and a new directory created in C:\
* `$ openssl version` returns `OpenSSL 1.1.1  11 Sep 2018`
*  `$ ls azure-iot-sdk-c/certs/` returns `certs.c  certs.h`, same as https://github.com/Azure/azure-iot-sdk-c/tree/master/certs 
[...]

PR #1713, opened 2020-10-23
update cert gen script to exit on failure

danewalton:
tends to #1554


jspaith (commented):
This is a great change, thanks.

jspaith (approved):

Issue #1711, opened 2020-10-22
SetOptions documentation needs cleanup


jspaith:
[Set options docs](https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/Iothub_sdk_options.md) needs spring cleaning.

Off top of my head, and I'm sure there's others:
* Need to document the `modelId` logic that was added for PnP.
* Need to document which options are applicable for UploadToBlob (which are not 100% set, because UploadToBlob uses HTTP handle whereas everything else is xio based)
* Need to document in the UploadToBlob API reference itself a note about not options being supported.
* Run through other options in C SDK and see what if anything is missing.
* Remove documentation of `XIO_RECEIVE_BUFFER_SIZE`.  (Do NOT remove its implementation, just fact it's doc'd.)  This isn't really the type of thing I want encouraging folks to use.

kiranpradeep:
@jspaith Can SetOptions documentation also be extended to include OPTION_DO_WORK_FREQUENCY_IN_MS?

jspaith:
> @jspaith Can SetOptions documentation also be extended to include OPTION_DO_WORK_FREQUENCY_IN_MS?

Yes, it'll get done as part of this too.

jspaith:
Fixed #1714.

az-iot-builder-01:
@kiranpradeep, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #1011, opened 2019-05-17
OPTION_DO_WORK_FREQUENCY_IN_MS data type mismatch


kiranpradeep:
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

Ubuntu 18.04, Intel Core i5, clang  6.0.0

**SDK Version (Please Give Commit SHA if Manually Compiling)**

apt-get install azure-iot-sdk-c-dev (latest - 0.2.0.0-11bionic)

**Protocol**

MQTT Websockets (bug is irrespective of protocol)

**Describe the Bug**

1. IoTHub C SDK Options document [[1](https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/Iothub_sdk_options.md)] doesn't document data type of OPTION_DO_WORK_FREQUENCY_IN_MS.
2. iothub_client_core.c[[2](https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/src/iothub_client_core.c#L51)] defines the type as `uint16_t`. But there are occasional casts to tickcounter_ms_t [[3](https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/src/iothub_client_core.c#L1716)] which appears to be buggy.
3. Assertion fails when used as in below sample source.

```c
#include <assert.h>
#include "azureiot/iothub.h"
#include "azureiot/iothub_device_client.h"
#include "azureiot/iothub_client_options.h"
#include "azureiot/iothubtransportmqtt_websockets.h"
#include "azure_c_shared_utility/tickcounter.h"

int main(void)
{
        char connectionString[] = "HostName=xxxxxxxxxxxxxxxxx.azure-devices.net;DeviceId=xxxxxxxx;SharedAccessKey=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
        IoTHub_Init();

        IOTHUB_DEVICE_CLIENT_HANDLE iotHubClientHandle = IoTHubDeviceClient_CreateFromConnectionString(connectionString, MQTT_WebSocket_Protocol);
        assert(iotHubClientHandle != NULL) ;

        tickcounter_ms_t timeout = 10000;
        int ret = IoTHubDeviceClient_SetOption(iotHubClientHandle, OPTION_MESSAGE_TIMEOUT, &timeout);
        assert(ret == IOTHUB_CLIENT_OK);

[...]

PR #1714, opened 2020-10-23
Documentation cleanup

jspaith:
Cleanup of some of reference documentation.  Inspired by #1711 but got bigger once started.

* Better document options setting.  Specifically customers have been confused that file upload doesn't implement the full set of options.  Also add DPS options to this doc and restructure a bunch and expand documentation for certain critical options e.g. Trusted certificate stuff.
* Mark the iothub_client.h and iothub_client_ll.h APIs as deprecated and point customers to newer ones.  (Note: We can NEVER remove these headers or implementations as we know it'd have very bad back-compat impact on customers.)
* Move samples / e2e tests to using the DeviceClient instead of the Client that's just been deprecated.
* Call out iothub_client_core.h and iothub_client_core_ll.h headers as internal.  These really should've been in `./internal` directory in 1st place and maybe we can moev them (it'd "only" cause compile breaks, which would be easy to fix).  Not moving them on this PR as enough is going on.
* TONS of fit and finish.  Like using word "network" for "wire", cleaning up sloppiness on LL APIs referenced in convenience headers where they shouldn't have been, removing unneeded use of word "module" places as that's too overloaded in Azure IoT, etc. etc.


iothub_client/inc/iothub_client.h
```diff
@@ -1,8 +1,16 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+//**********************************************************************
+// NOTE: THIS HEADER IS DEPRECATED
+// 
+// Functions in this header will be maintained for backward compatability.
+// New applications should use iothub_device_client.h, however.
```
danewalton, line 8: nit: 
```suggestion
// New applications should use iothub_device_client.h.
```

iothub_client/inc/iothub_client.h
```diff
@@ -37,354 +45,109 @@ extern "C"
...
-    *
-    *            Sample connection string:
-    *                <blockquote>
-    *                    <pre>HostName=[IoT Hub name goes here].[IoT Hub suffix goes here, e.g., private.azure-devices-int.net];DeviceId=[Device ID goes here];SharedAccessKey=[Device key goes here];</pre>
-    *                   <pre>HostName=[IoT Hub name goes here].[IoT Hub suffix goes here, e.g., private.azure-devices-int.net];DeviceId=[Device ID goes here];SharedAccessSignature=SharedAccessSignature sr=[IoT Hub name goes here].[IoT Hub suffix goes here, e.g., private.azure-devices-int.net]/devices/[Device ID goes here]&sig=[SAS Token goes here]&se=[Expiry Time goes here];</pre>
-    *                </blockquote>
-    *
-    * @return    A non-NULL @c IOTHUB_CLIENT_HANDLE value that is used when
-    *             invoking other functions for IoT Hub client and @c NULL on failure.
+    * @warning IoTHubClient_CreateFromConnectionString is deprecated.  Use IoTHubDeviceClient_CreateFromConnectionString instead.
```
danewalton, line 48: Additional option is the `@deprecated` tag in doxygen.

image

iothub_client/inc/iothub_client.h
```diff
@@ -37,354 +45,109 @@ extern "C"
...
-    *             Hub using the specified parameters.
-    *
-    * @param    config    Pointer to an @c IOTHUB_CLIENT_CONFIG structure
-    *
-    *            The API does not allow sharing of a connection across multiple
-    *            devices. This is a blocking call.
-    *
-    * @return    A non-NULL @c IOTHUB_CLIENT_HANDLE value that is used when
-    *             invoking other functions for IoT Hub client and @c NULL on failure.
+    * @warning IoTHubClient_Create is deprecated.  Use IoTHubDeviceClient_Create instead.
```
danewalton, line 53: Would we want something like this?

```c
#ifdef _MSC_VER
#define DEPRECATED __declspec(deprecated)
#elif defined(__GNUC__) | defined(__clang__)
#define DEPRECATED __attribute__((__deprecated__))
#else
#define DEPRECATED
#endif

//usage:
DEPRECATED void foo(int bar);
```

giving users a compile time warning? Or would we also get these if these are used underneath our APIs

provisioning_client/inc/azure_prov_client/prov_device_client.h
```diff
@@ -32,6 +32,8 @@ extern "C"
 {
 #endif
 
+static const char* const PROV_OPTION_DO_WORK_FREQUENCY_IN_MS = "do_work_freq_ms";
```
danewalton, line 35: If I understand correctly, if this is static, it will create an instantiation of this every time this is included (which could be wasteful from a memory perspective). Should this drop the static?

danewalton (approved):

iothub_client/inc/iothub_client.h
jspaith: Fixed.

iothub_client/inc/iothub_client.h
jspaith: I really like this.  Changing.

provisioning_client/inc/azure_prov_client/prov_device_client.h
jspaith: We need the static since it's being declared in .h.  If you look at `prov_device_client_ll.h` and `iothub_client_options.h` this is the pattern they use.  Basically - iothub_client_options.h has e.g. below.  I'm adding the STATIC_VAR_UNUSED since to prov options since it's just a matter of time before something trips up over that anyway.

```c
static STATIC_VAR_UNUSED const char* OPTION_CONNECTION_TIMEOUT = "connect_timeout";
```

iothub_client/inc/iothub_client.h
jspaith: I'd love something like that, obviously with a namespace fixup so we don't camp on `DEPRECATED`.

But the `MOCKABLE_FUNCTION` get in the way and I don't want to try and shoe-horn anything else around that even were in possible in theory.  I've put a comment in #1342 draft PR that if that ever comes in, we should reconsider this issue.

doc/Iothub_sdk_options.md
```diff
@@ -1,119 +1,178 @@
...
+| `"product_info"`                  | OPTION_PRODUCT_INFO             | const char*        | User defined Product identifier sent to the IoT Hub service
+| `"retry_interval_sec"`            | OPTION_RETRY_INTERVAL_SEC       | unsigned int*      | Number of seconds between retries when using the interval retry policy.  (Not supported for HTTP transport.)
+| `"retry_max_delay_secs"`          | OPTION_RETRY_MAX_DELAY_SECS     | unsigned int*      | Maximum number of seconds a retry delay when using linear backoff, exponential backoff, or exponential backoff with jitter policy.  (Not supported for HTTP transport.)
+| `"sas_token_lifetime"`            | OPTION_SAS_TOKEN_LIFETIME       | size_t*            | Length of time in seconds used for lifetime of SAS token.
+| `"do_work_freq_ms"`               | OPTION_DO_WORK_FREQUENCY_IN_MS  | unsigned int *     | Specifies how frequently the worker thread spun by the convenience layer will wake up, in milliseconds.  The default is 1 millisecond.  The maximum allowable value is 100.  (Convenience layer APIs only)
```
kiranpradeep, line 84: It appears the type is `tickcounter_ms_t` https://github.com/Azure/azure-iot-sdk-c/issues/1011#issuecomment-494508855. Kindly ignore if otherwise.

jspaith: Good catch.  I will fix this in next round.  I'll double check the provisioning client equivalent of this and fixup here if in correct.  Thanks

jspaith: OK - fixed locally, put a link to header that has `tickcounter_ms_t` as this isn't a standard data type, and fixed up prov client (which takes a uint16*).  Will push if/when more feedback comes &certainly before merge.

PR #1715, opened 2020-10-26
update bash build scripts

danewalton:
@jspaith


jspaith (approved):
Thank you!

danewalton:
note that the gcc4.4 build was failing due to curl not being installed on the docker container. I updated the docker image and installed curl on it.

Issue #1691, opened 2020-10-13
Unable to build sample project using vcpkg


SergioMasson:
I`m trying to build the sample project iothub_ll_telemetry_sample using a windows console application on x64 architecture. The SDK is installed using vcpkg. All header files are correctly found. However, I get multiple linker errors when trying to build the console app. 

Error list: 
aziotsharedutil.lib(platform_win32.c.obj) : error LNK2001: unresolved external symbol __imp_WSAStartup
1>aziotsharedutil.lib(platform_win32.c.obj) : error LNK2001: unresolved external symbol __imp_WSACleanup
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol __imp_CertFreeCertificateContext
1>aziotsharedutil.lib(httpapi_winhttp.c.obj) : error LNK2001: unresolved external symbol __imp_CertFreeCertificateContext
1>aziotsharedutil.lib(x509_schannel.c.obj) : error LNK2001: unresolved external symbol __imp_CertFreeCertificateContext
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol AcquireCredentialsHandleA
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol __imp_FreeCredentialsHandle
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol InitializeSecurityContextA
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol __imp_DeleteSecurityContext
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol QueryContextAttributesA
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol EncryptMessage
1>aziotsharedutil.lib(tlsio_schannel.c.obj) : error LNK2001: unresolved external symbol DecryptMessage
1>aziotsharedutil.lib(httpapi_winhttp.c.obj) : error LNK2001: unresolved external symbol __imp_WinHttpSetStatusCallback
1>aziotsharedutil.lib(httpapi_winhttp.c.obj) : error LNK2001: unresolved external symbol __imp_WinHttpOpen
[...]

PR #1721, opened 2020-10-28
Fix samples vcproj files to load from correct vcpkg path, update vc++ deps

ewertons:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/1691

# Description of the problem

Projects of the samples were using the incorrect vcpkg variable to find installed packages.

# Description of the solution
 
Updated VcpkgRoot var to VcpkgCurrentInstalledDir.
Also upgraded target on samples projects to VS2019.

### DIFF
--- a/iothub_client/samples/iothub_client_sample_upload_to_blob/windows/iothub_client_sample_upload_to_blob.vcxproj
+++ b/iothub_client/samples/iothub_client_sample_upload_to_blob/windows/iothub_client_sample_upload_to_blob.vcxproj
@@ -94,7 +94,7 @@
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>GB_MEASURE_MEMORY_FOR_THIS;GB_DEBUG_ALLOC;WIN32;_CRT_SECURE_NO_WARNINGS;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>..\;$(VcpkgRoot)include\azureiot;</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\;$(VcpkgCurrentInstalledDir)include\azureiot;</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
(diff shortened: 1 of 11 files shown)


iothub_client/samples/iothub_client_sample_upload_to_blob/windows/iothub_client_sample_upload_to_blob.vcxproj
```diff
@@ -27,26 +27,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
```
CIPop, line 30: Why are we checking in Visual Studio project files? Shouldn't these be generated by CMake?

ericwol-msft: This was added by JetstreamRoySprowl a few years ago. Good question is how this is being maintained.

ewertons: No. These have been there for a long time to make it easy for customers that want to run the samples using packages like nugets (now discontinued) and vcpkg.

CIPop: Do these reference the same source files that cmake uses or are they duplicated? 
We should think of moving these to Azure-Samples if they don't use the main build.

ericwol-msft (approved):

iothub_client/samples/iothub_client_sample_upload_to_blob/windows/iothub_client_sample_upload_to_blob.vcxproj
ewertons: These projects do not duplicate code perse, they use the code from the parent folder, which is the same "main.c" file used by cmake. There is no duplication.

CIPop (approved):

PR #1734, opened 2020-11-06
update ca certificate documentation

### LINKED ISSUE #1575: Confusing Instructions on Proof Of Possession Step
In the doc, CACertificateOverview.md, in Step 3, it says:
"Only perform this step if you're setting up CA Certificates and proof of possession. For simple device certificates, such as Edge certificates, skip to the next step"

I have an Edge device, and want to use production CA Certificates to authenticate it to the hub.  Do I need to do Step 3 or not?  The statement in the doc is very confusing and offers contradictory guidance.

It seems to imply that an Edge certificate is a 'simple device certificate' - is it?  I thought the Edge devices, for production, should also use Production, CA signed certificates to authenticate with the hub?

Please clarify whether the proof of possession step is needed in the case of installing CA-signed certificates onto an Edge device.

Also confusing, this doc is referenced by this article:
[https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-security-x509-get-started](url)

That article is for installing Production certificates.  And yet the warning at the top of CACertificateOverview.md says that "Certificates created by these scripts **MUST NOT** be used for production."  Of course that is incorrect.  These outputs of these scripts CAN be used for production, if you started with a CA-signed root, it just depends on which steps you are performing, etc.  But that warning is so emphatic it could scare off developers, assuming they are in the wrong place.

Please remove, or clarify this warning.

danewalton:
fixes #1575


danewalton:
Hey @massand long time no see! Just adding you here to hopefully validate that this is indeed correct. I had asked the Edge team about it a while back and I want to make sure this is still correct.

wduraes (approved):

massand (approved):
Yes - still true as of November 6 2020. 
Thanks for the Edge device callout in the certGen docs 🖖

Issue #1255, opened 2019-10-11
Disconnect without connection status callback called


Patrik-Berglund:
Development Machine, OS, Compiler (and Other Relevant Toolchain Info)
ESP-IDF v3.3
Compiled on Ubuntu 16.04 using GCC 5.4.0

SDK Version (Please Give Commit SHA if Manually Compiling)
Release 2019-07-01

Protocol
MQTT
Using tlsio_mbedtls and socketio_berkeley from c-utility

Describe the Bug
Connection status callback not called on disconnect after "IOTHUB_CLIENT_CONFIRMATION_MESSAGE_TIMEOUT". The client just reconnects without notification.

Error log,
```
I (133463) IOT HUB: Result Call Back Called! Result is: IOTHUB_CLIENT_CONFIRMATION_MESSAGE_TIMEOUT 
Error: Time:Fri Oct 11 09:15:38 2019 File:/components/azure-iot/pal/src/socketio_berkeley.c Func:socketio_send Line:834 Failure: socket state is not opened.
-> 09:15:38 DISCONNECT
Error: Time:Fri Oct 11 09:15:38 2019 File:/components/azure-iot/pal/src/socketio_berkeley.c Func:socketio_send Line:834 Failure: socket state is not opened.
-> 09:15:39 CONNECT | VER: 4 | KEEPALIVE: 60 | FLAGS: 192 | USERNAME: XXXXXXX.azure-devices.net/XXXXXXX/?api-version=2017-11-08-preview&DeviceClientType=iothubclient%2f1.3.4%20(berkeley_mbedtls) | PWD: XXXX | CLEAN: 0
<- 09:15:40 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
I (135153) IOT HUB: Connection Status Callback: status=0 reason=6
I (135153) IOT HUB: IOTHUB_CLIENT_CONNECTION_OK
```

ewertons:
Thank you, @Patrik-Berglund !

This indeed seems incorrect, we will verify and prioritize it accordingly.

momuno:
Hi @Patrik-Berglund, I am looking into this bug.  Can you please tell me what device you were using? Was it an esp2866, esp32, or something else?

Patrik-Berglund:
Hi @momuno 

I don't work for the company anymore where this product was developed.

[...]

PR #1735, opened 2020-11-07
Fix for Issue #1255: Disconnect without connection status callback called

momuno:
Addresses Issue #1255.

Bug occurs in logic of `IoTHubTransport_MQTT_Common_DoWork` --> `process_queued_ack_messages`, when the `RESEND_TIMEOUT_VALUE_MIN` is met and `MAX_SEND_RECOUNT_LIMIT` is reached.  Upon modifying those macro values, bug was reproduced using a modified `iothub_ll_telemetry_sample`, with similar output:

```
Confirmation callback received for message 5 with result IOTHUB_CLIENT_CONFIRMATION_MESSAGE_TIMEOUT
-> 22:11:58 DISCONNECT
Sending message 6 to IoTHub
-> 22:12:00 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: xxx | PWD: XXXX | CLEAN: 0
Sending message 7 to IoTHub
<- 22:12:06 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
The device client is connected to iothub
```

The problem is that the `connection_status_cb` is never called post `DisconnectFromClient`, as it is done other places, such as in `InitializeConnection` when the SAS token has expired.

After making the change, the callback ocurs:

```
Confirmation callback received for message 5 with result IOTHUB_CLIENT_CONFIRMATION_MESSAGE_TIMEOUT
-> 22:18:01 DISCONNECT
The device client has been disconnected
Sending message 6 to IoTHub
-> 22:18:02 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: xxx | PWD: XXXX | CLEAN: 0
Sending message 7 to IoTHub
<- 22:18:07 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
The device client is connected to iothub
```

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -2106,7 +2106,11 @@ static void process_queued_ack_messages(PMQTTTRANSPORT_HANDLE_DATA transport_dat
                 free(msg_detail_entry);
 
                 DisconnectFromClient(transport_data);
-                transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED, transport_data->transport_ctx);
+                if (!transport_data->isRetryExpiredCallbackSet) // Only call once
+                {
+                    transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED, transport_data->transport_ctx);
+                    transport_data->isRetryExpiredCallbackSet = true; // Reset to false upon successful call to sendMqttConnectMsg in new connection
+                }
             }
             else
             {
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -1063,7 +1063,7 @@ static void sendPendingGetTwinRequests(PMQTTTRANSPORT_HANDLE_DATA transportData)
 static void removeExpiredTwinRequestsFromList(PMQTTTRANSPORT_HANDLE_DATA transport_data, tickcounter_ms_t current_ms, DLIST_ENTRY* twin_list)
 {
     PDLIST_ENTRY list_item = twin_list->Flink;
-    
+
     while (list_item != twin_list)
     {
         DLIST_ENTRY next_list_item;
@@ -2106,6 +2106,7 @@ static void process_queued_ack_messages(PMQTTTRANSPORT_HANDLE_DATA transport_dat
                 free(msg_detail_entry);
 
                 DisconnectFromClient(transport_data);
+                transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED, transport_data->transport_ctx);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2106,6 +2106,7 @@ static void process_queued_ack_messages(PMQTTTRANSPORT_HANDLE_DATA transport_dat
                 free(msg_detail_entry);
 
                 DisconnectFromClient(transport_data);
+                transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED, transport_data->transport_ctx);
```
danewalton, line 2109: Curious if you can double check something here?

Looking at this, this gets called here
https://github.com/Azure/azure-iot-sdk-c/blob/26d8aed1362e4434afcf2f62267995bb92e6c262/iothub_client/src/iothubtransport_mqtt_common.c#L3272 in `IoTHubTransport_MQTT_Common_DoWork()`.

If [`InitializeConnection()`](https://github.com/Azure/azure-iot-sdk-c/blob/26d8aed1362e4434afcf2f62267995bb92e6c262/iothub_client/src/iothubtransport_mqtt_common.c#L3176) fails, this section might not have been hit:
https://github.com/Azure/azure-iot-sdk-c/blob/26d8aed1362e4434afcf2f62267995bb92e6c262/iothub_client/src/iothubtransport_mqtt_common.c#L2434-L2438
which sets the callback and sets that "set" flag to true. I'm wondering if this line you added should be wrapped in a similar `if (!transport_data->isRetryExpiredCallbackSet)` to prevent invoking a callback which is `NULL`.

Hate to be that person but there's a chance there should be a UT here for this scenario if that's true.

danewalton (changes requested):
Might also need to patch up the "expected calls" ut stuff

iothub_client/src/iothubtransport_mqtt_common.c
momuno: No you're good, thank you for your attention to detail here.  Especially since this space is new for me.  What is a UT?  And what do you mean by patching up the expected calls ut stuff?

danewalton: Unit tests. So if you add a new function call in one of our API's, you likely have to change the UT's since they have a list of "expected calls" to verify. The idea being that with a given UT, you expect the code to flow a certain, defined way. We set up "expected calls" to verify that indeed, the flow you expected was called. So in this case, there's a chance that a new call breaks the previous expected "flow".

danewalton: example I think in the pipeline here:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=72109&view=logs&j=d89e688f-0939-5d98-042b-db98508a1b62&t=8fca59f4-1111-5c89-dc70-3102283020d1&l=1755

momuno: awesome, thank you!

momuno: Q:  How does making the call `transport_data->transport_callbacks.connection_status_cb(...` set the callback?  I just see the callback called, not "set".

danewalton (approved):
This looks good to me. I'd like @ewertons or @jspaith to weigh in here since the purpose of `transport_data->isRetryExpiredCallbackSet` isn't obvious to me. Maybe the naming is just a bit odd.

PR #1741, opened 2020-11-11
Fix Valgrind issue in UploadToBlob logging

jspaith:
A Valgrind run on a fail case discovered that our logging was using the uninitialized variable for `httpStatus`.  We can only trust this value if the `Blob_UploadBlock` returned `BLOB_OK`; otherwise it may or may not have been set.
This should *only* fix logging out a potentially uninit'd variable.  There should be no functional changes otherwise.


ewertons (approved):
Valgrind complains about this joint log? If you have it, could you share the error message so I can save in my notes? Thanks John.

jspaith:
> Valgrind complains about this joint log? If you have it, could you share the error message so I can save in my notes? Thanks John.

I don't have the exact message handy as it's in some log from many runs ago, but the gist of this is `httpStatus` is not set by the caller of `InvokeUserCallbackAndSendBlobs` nor is it set by `InvokeUserCallbackAndSendBlobs` itself.  If our call to `Blob_UploadBlock` fails, there are paths where `httpStatus` will not be set so we can't trust this.  The product code is fine but the log code wasn't.

ericwol-msft (approved):

Issue #1729, opened 2020-11-04
'azure_macro_utils/macro_utils.h' file not found in Xcode 12


cbauer10:
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

MacBook Pro, iOS 14, Xcode 12.1

**SDK Version (Please Give Commit SHA if Manually Compiling)**

Release 1.3.9

**Describe the Bug**

CocoaPod no longer compiles with the following three errors:

`'azure_macro_utils/macro_utils.h' file not found: serializer_e2e.c`
`'azure_c_shared_utility/threadapi.h' file not found: simple sample_mqtt.c`
`'umock_c/umock_c.h' file not found: schemaserializer_ut.c`

Simply use the sample that you have provided to see these issues:
https://github.com/Azure-Samples/azure-iot-samples-ios

Frank-vdm:
im also having this issue, reading the pod files does not help the issue

cbauer10:
Just curious if there is any update on this issue?

jbobotek:
I am investigating it as we speak. No updates other than that those ut and e2e files shouldn't be even in the pod. Currently fresh downloads of the sample and pods are not giving me the IoTHubClient framework, so working through that first.

cbauer10:
Really appreciate it! Let me know if you need any further info from me to assist. Thank you!

Frank-vdm:
I found an ugly work around, maybe this helps somehow solve the issue more permanently, I updated the files in the box to import from `macro_utils` instead of `azure_macro_utils`.

jbobotek:
Ok, I have not been able to repro your issue. If you're using the latest Pods, you shouldn't be getting these e2e and ut errors as they are not included files. They may have been included previously many updates ago.
I will soon be updating the pods and Podfile to fix another issue found in iOS 14.1. I'll test it again after fixing this issue.

cbauer10:
@jbobotek I am not sure what else I can provide you with to assist.  If you pull the sample and try to compile it does not work.  Those samples are pointed at `1.3.9-LTS`, but if I set it to point at the most recent it has the same issue.  It should be very easy to reproduce.

Can you give an estimate on the update you are releasing?  Perhaps that will solve it.

jbobotek:
I agree, I was having issues compiling for iOS 12 as well, but I did not run into the same bug. Did you try @Frank-vdm 's fix?

cbauer10:
[...]

PR #1751, opened 2020-11-17
Updating dependencies for Xcode 12 compatibility.

cbauer10:
# Checklist

#1729 

# Description of the problem
The CocoaPod no longer compiles in Xcode 12 as the libraries are being imported from the incorrect place.

# Description of the solution
Importing the files from `macro_utils` instead of `azure_macro_utils` has resolved this issue.

### DIFF
--- a/iothub_client/inc/internal/blob.h
+++ b/iothub_client/inc/internal/blob.h
@@ -13,7 +13,7 @@
 #ifndef BLOB_H
 #define BLOB_H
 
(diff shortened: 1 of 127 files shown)


jbobotek:
Yeah, the problem is not the files, it's that the azure_macro_utils library is being imported with the wrong file structure.

cbauer10:
I see.  So are you making that change then?

jbobotek:
> I see. So are you making that change then?

Correct.

Issue #1695, opened 2020-10-15
Cannot send certain ASCII characters to IoTHub when using MQTT


momuno:
Running C-SDK sample iothub_ll_telemetry_sample, and using MQTT, the following behavior is observed:

characters '#', '&', '+', '/', '=' do not work for key or property values.
The client disconnects and either
1) no telemetry is received by IoTHub.(EX: "#":"a")
2) a null property is received  (EX: "a":"/")
```
{
  "body": "test_message",
  "enqueuedTime": "2020-10-14T20:36:18.890Z",
  "properties": {
    "a": null
  }
}
```

All other ASCII characters from [space] to '\~' work as expected.
(EX: "0123456789 qwertyuiopasdfghjklzxcvbnm QWERTYUIOPASDFGHJKLZXCVBNM !\"$%'()\*,-.:;<>?@[\\]\^\_\`{|}\~":"0123456789 qwertyuiopasdfghjklzxcvbnm QWERTYUIOPASDFGHJKLZXCVBNM !\"$%'()\*,-.:;<>?@[\\]^_\`{|}~")

IoTHub has validated that all characters [space] to '~' work when using MQTT and running .net SDKs.  It is suspected, therefore, that the server is able to receive the characters and that the C SDK may not be sending the message properly.

**To reproduce:**
In the source code for the sample, adjust the call to IoTHubMessage_SetProperty(), updating the key and property values passed in as parameters.  Rebuild the sample using instructions for building the samples (cmake --build .), and run the sample.  Use Azure IoT Explorer to observe IoTHub received telemetry messages.

A failed console log will look like the following:

```
Creating IoTHub Device handle
Sending message 1 to IoTHub
Sending message 2 to IoTHub
Sending message 3 to IoTHub
Sending message 4 to IoTHub
Sending message 5 to IoTHub
-> 13:48:05 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: momuno-hub.azure-devices.net/hub-sas-test/?api-version=2019-10-01&DeviceClientType=iothubclient%2f1.3.9%20(native%3b%20Linux%3b%20x86_64) | PWD: XXXX | CLEAN: 0
<- 13:48:05 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
The device client is connected to iothub
[...]

PR #1753, opened 2020-11-20
Turn on mqtt encode/decode option

momuno:
# Checklist

Updates existing samples that use mqtt to set the encode/decode option.  This solves the encoding failure in Issue #1695 .
It also points the developer to use this in their code by including it and turning it on as default in the other mqtt samples.


ericwol-msft (approved):

ewertons (approved):
Mollie, please make sure you run the samples (including sending D2C) with MQTT and MQTT/WS on both Windows and Linux to validate they will still work correctly.

PR #1758, opened 2020-11-24
Add jitter between test cases

jspaith:
Add 1.5 second jitter to uploadToBlob e2e test cases, in addition to the 3 second sleep between run.  The test tends to hit bursts which we think causes throttling issues.


momuno:
This is certainly worth a shot.  Looks good to me, but I'll wait for someone with more knowledge of our test environment to approve.

danewalton (approved):
We'll see how this goes, but considering the random could still place upload to blob tests together, we could get transient failures. We could decide to add another `ctest` option to run twice in case of failure to double check that the interweaving isn't what caused it to fail.

jbobotek (approved):

jspaith:
@danewalton - ack, we'll continue to keep an eye on this and I'll be some more gate stress runs hopefully to keep chasing down any tail of infra instability on all this.  I'm cautiously optimistic the random will help; we have 436 tests on Ubuntu with 16 executing in parallel, so at any given time only ~3% will run at same time.  Now there's 4 upload tests and they're more likely to run in parallel than 100% random because the UTs are fast.  But I'm hoping this puts a huge dent in this.

PR #1761, opened 2020-11-25
Add additional schedule-random to ctest

jspaith:
Initial change #1758 that added `--schedule-random` missed some of our build scripts.  Addressing here.

### DIFF
--- a/build_all/linux/build.sh
+++ b/build_all/linux/build.sh
@@ -167,10 +167,10 @@ then
   then
     #use doctored openssl
     export LD_LIBRARY_PATH=/usr/local/ssl/lib
-    ctest -j $TEST_CORES --output-on-failure
+    ctest -j $TEST_CORES --output-on-failure --schedule-random
     export LD_LIBRARY_PATH=
   else
-    ctest -j $TEST_CORES -C "Debug" --output-on-failure
+    ctest -j $TEST_CORES -C "Debug" --output-on-failure --schedule-random
   fi
 fi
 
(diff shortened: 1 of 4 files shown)


danewalton (approved):

ericwol-msft:
This is failing because we are now getting this error on Linux

Error: Time:Wed Nov 25 18:12:04 2020 File:/__w/9/s/c-utility/adapters/tlsio_openssl.c Func:openssl_static_locks_install Line:545 **Locks already initialized**
Error: Time:Wed Nov 25 18:12:04 2020 File:/__w/9/s/c-utility/adapters/tlsio_openssl.c Func:tlsio_openssl_init Line:1084 **Failed to install static locks in OpenSSL!**

ericwol-msft (approved):

jspaith:
The test failures you mention Eric are false positives in logs; we call platform_init twice in e2e test for some reason (which we should deal with but a separate issue).  These changes are increasing the % of test pass but nowhere near where we need to be.

I've just pushed a new change that adds jitter to the multi-thread tests as I've seen those failing more after my global jitter change, and removing some unneeded e2e tests.  

We could actually probably go further on these protocol test removals.  As an experiment I NULL'd out the transport handle (which is usually used for telemetry/twin/method) in a debugger on UploadE2Etests.  Other than DoWork() printing error messages about the handle being NULL, the upload test itself was fine.  The reason is the upload logic is all about a new HTTP handle for both Azure IoT & Storage, not the MQTT/etc transport.

jspaith:
Also the latest change fixes some bad behavior in tests where they were passing a variable scoped to a function to the SDK callback, but the test function may have stopped executing but the SDK thread continues.  Moved such test variables to global to be safe.

ericwol-msft (approved):

PR #1794, opened 2020-12-11
retry longhaul Device Method due to SAS refresh

ericwol-msft:
# Checklist

# Description of the problem
When the SAS token expires, we need to reconnect and if we send a device message we get an 404 error because the device is no longer connected

# Description of the solution
retry the device method again after 60 seconds to allow the device to reconnect

### DIFF
--- a/iothub_client/tests/common_longhaul/iothub_client_common_longhaul.c
+++ b/iothub_client/tests/common_longhaul/iothub_client_common_longhaul.c
@@ -1396,7 +1396,26 @@ static int invoke_device_method(const void* context)
                 MAX_DEVICE_METHOD_TRAVEL_TIME_SECS,
                 &responseStatus, &responsePayload, &responseSize)) != IOTHUB_DEVICE_METHOD_OK)
             {
-                LogError("Failed invoking device method");
+                if (responseStatus == 404)
+                {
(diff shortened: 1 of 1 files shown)


iothub_client/tests/common_longhaul/iothub_client_common_longhaul.c
```diff
@@ -1396,7 +1396,26 @@ static int invoke_device_method(const void* context)
                 MAX_DEVICE_METHOD_TRAVEL_TIME_SECS,
                 &responseStatus, &responsePayload, &responseSize)) != IOTHUB_DEVICE_METHOD_OK)
             {
-                LogError("Failed invoking device method");
+                if (responseStatus == 404)
```
ewertons, line 1399: If we are going to retry instead of reconciling the results with reconnection data, we should put this in a loop of 2 iterations instead of duplicating the code I think. What do you say?

ewertons (changes requested):

iothub_client/tests/common_longhaul/iothub_client_common_longhaul.c
ericwol-msft: made retry a loop

ewertons (approved):

Issue #1767, opened 2020-12-01
'segment fault' - after closing SSH session to a device via device streams


gerhard67:
**Development Machine, OS, Compiler (and Other Relevant Toolchain Info)**

Linux ${mydevice} 4.15.0-1098-azure #109~16.04.1-Ubuntu SMP Wed Sep 30 18:53:14 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux 

**SDK Version (Please Give Commit SHA if Manually Compiling)**

demouser@iiot-lab-edge-vm:~/azure-iot-sdk-c$ git log
commit be574df7ba495f5463c5325a6433406ce754744c
Author: xinyiz1226 <xinyiz@microsoft.com>
Date:   Fri Feb 28 06:05:41 2020 +0800

**Describe the Bug**

I have the same issue, as described here https://github.com/Azure/azure-iot-sdk-c/issues/967
Unfortunately, the issue is closed, but not described who it was fixed. 

steps 

> - start on the device "iothub_client_c2d_streaming_proxy_sample"
> - start on the client "Quickstarts\device-streams-service\proxy.js"
> - start putty to connect 
> - login 
> - exit 
> - iothub_client_c2d_streaming_proxy_sample dumps 

**log on the device** 

${myuser}@${mydevice}:~/azure-iot-sdk-c/cmake/iothub_client/samples/iothub_client_c2d_streaming_proxy_sample$ ./iothub_client_c2d_streaming_proxy_sample
-> 12:28:16 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: ${myiothubname}.azure-devices.net/${mydeviceid}/?api-version=2019-07-01-preview&DeviceClientType=iothubclient%2f1.3.6%20(native%3b%20Linux%3b%20x86_64) | PWD: XXXX | CLEAN: 0
<- 12:28:16 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
-> 12:28:16 SUBSCRIBE | PACKET_ID: 2 | TOPIC_NAME: $iothub/streams/POST/# | QOS: 0 | TOPIC_NAME: $iothub/streams/res/# | QOS: 0
<- 12:28:16 SUBACK | PACKET_ID: 2 | RETURN_CODE: 0 | RETURN_CODE: 0
[...]

PR #1800, opened 2020-12-17
Fix Device Streaming samples to check args in callbacks and avoid buffer underrun (gh#1767)

ewertons:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/1767

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/samples/iothub_client_c2d_streaming_proxy_sample/iothub_client_c2d_streaming_proxy_sample.c
+++ b/iothub_client/samples/iothub_client_c2d_streaming_proxy_sample/iothub_client_c2d_streaming_proxy_sample.c
@@ -109,6 +109,12 @@ static void on_bytes_received(void* context, const unsigned char* buffer, size_t
 {
     (void)context;
 
+    if (buffer == NULL)
+    {
+        (void)printf("[ERROR] on_bytes_received invoked with NULL buffer (unexpected)\r\n");
+        exit(1);
+    }
+
     if (g_is_uws_client_ready)
     {
         size_t send_size;
@@ -218,7 +224,21 @@ static void on_ws_peer_closed(void* context, uint16_t* close_code, const unsigne
 {
     (void)context;
     (void)extra_data_length;
-    (void)printf("on_ws_peer_closed (Code: %d, Data: %.*s)\r\n", *close_code, (int)extra_data_length, extra_data);
+
+    (void)printf("on_ws_peer_closed (");
+
+    if (close_code != NULL)
+    {
+        (void)printf("Code: %d, ", *close_code);
+    }
+
+    if (extra_data != NULL)
+    {
+        (void)printf("Data: %.*s", (int)extra_data_length, extra_data);
+    }
+
+    (void)printf(")\r\n");
+
     g_continueRunning = false;
 }
 
--- a/iothub_client/samples/iothub_client_c2d_streaming_sample/iothub_client_c2d_streaming_sample.c
+++ b/iothub_client/samples/iothub_client_c2d_streaming_sample/iothub_client_c2d_streaming_sample.c
@@ -74,7 +74,6 @@ static const char* proxy_password = NULL; // Proxy password
 
 static bool g_continueRunning = true;
 static UWS_CLIENT_HANDLE g_uws_client_handle = NULL;
-static char stream_payload[128];
 
 
 static void on_ws_open_complete(void* context, WS_OPEN_RESULT ws_open_result)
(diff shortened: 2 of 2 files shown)


iothub_client/samples/iothub_client_c2d_streaming_proxy_sample/iothub_client_c2d_streaming_proxy_sample.c
```diff
@@ -111,22 +111,29 @@ static void on_bytes_received(void* context, const unsigned char* buffer, size_t
 
     if (g_is_uws_client_ready)
     {
-        size_t send_size;
-
-        while (size > 0)
+        if (buffer == NULL)
         {
-            send_size = size <= MAX_WEBSOCKET_PAYLOAD_SIZE_IN_BYTES ? size : MAX_WEBSOCKET_PAYLOAD_SIZE_IN_BYTES;
+            printf("on_bytes_received invoked with NULL buffer (unexpected)\r\n");
```
CIPop, line 116: This is a bit unclear: if this is an error can we say "error" and maybe exit the process?
The code continues here so I assume this is just a warning?

CIPop: Also, if it's a warning, it's unclear if it is a developer issue that should be corrected.

CIPop: Missing `(void)` before `printf`.

iothub_client/samples/iothub_client_c2d_streaming_proxy_sample/iothub_client_c2d_streaming_proxy_sample.c
```diff
@@ -218,7 +225,21 @@ static void on_ws_peer_closed(void* context, uint16_t* close_code, const unsigne
 {
     (void)context;
     (void)extra_data_length;
-    (void)printf("on_ws_peer_closed (Code: %d, Data: %.*s)\r\n", *close_code, (int)extra_data_length, extra_data);
```
CIPop, line 221: Nit: we could use an inline conditional to keep the code smaller and a single `printf` call if you think it's better. (I don't have a preference.)

iothub_client/samples/iothub_client_c2d_streaming_sample/iothub_client_c2d_streaming_sample.c
```diff
@@ -88,14 +87,7 @@ static void on_ws_send_frame_complete(void* context, WS_SEND_FRAME_RESULT ws_sen
 {
     (void)context;
 
-    if (ws_send_frame_result == WS_SEND_FRAME_OK)
```
CIPop, line 91: Can't this still happen? If no I think it would be nice to have an assert on what result is expected.

iothub_client/samples/iothub_client_c2d_streaming_sample/iothub_client_c2d_streaming_sample.c
```diff
@@ -105,19 +97,36 @@ static void on_ws_frame_received(void* context, unsigned char frame_type, const
...
-    (void)uws_client_send_frame_async(g_uws_client_handle, frame_type, buffer, size, true, on_ws_send_frame_complete, NULL);
+        (void)uws_client_send_frame_async(g_uws_client_handle, frame_type, buffer, size, true, on_ws_send_frame_complete, NULL);   
+    }
 }
 
 static void on_ws_peer_closed(void* context, uint16_t* close_code, const unsigned char* extra_data, size_t extra_data_length)
 {
     (void)context;
     (void)extra_data_length;
-    (void)printf("on_ws_peer_closed\r\nCode: %d\r\nData: %s\r\n", *(int*)close_code, extra_data);
```
CIPop, line 120: Nit: same as above if we want a single `printf` call.

CIPop (approved):
Left 2 comments and a few nitpicks.

iothub_client/samples/iothub_client_c2d_streaming_proxy_sample/iothub_client_c2d_streaming_proxy_sample.c
ewertons: I agree, we could just exit the process. It is an error if a buffer is null.

iothub_client/samples/iothub_client_c2d_streaming_sample/iothub_client_c2d_streaming_sample.c
ewertons: No matter the result, the sample ends when this function is called. Then there is nothing else to be done.

PR #1807, opened 2020-12-22
Addressing the Y2k38 issue for SAS tokens

CIPop:
Changing from `size_t` (32 bit on 32bit systems) to `uint64_t` for SAS time related operations to avoid the https://en.wikipedia.org/wiki/Year_2038_problem .

(Requires https://github.com/Azure/azure-c-shared-utility/pull/494).

Note: this is changing some of the `internal` APIs. No public API has been changed (the SAS expiry option is still expecting `size_t` which is converted internally to `uint64_t`).

### DIFF
--- a/iothub_client/src/iothubtransport_amqp_cbs_auth.c
+++ b/iothub_client/src/iothubtransport_amqp_cbs_auth.c
@@ -29,7 +29,7 @@ typedef struct AUTHENTICATION_INSTANCE_TAG
     ON_AUTHENTICATION_ERROR_CALLBACK on_error_callback;
     void* on_error_callback_context;
 
-    time_t cbs_request_timeout_secs;
+    uint64_t cbs_request_timeout_secs;
 
     AUTHENTICATION_STATE state;
     CBS_HANDLE cbs_handle;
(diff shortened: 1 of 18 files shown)


iothub_client/src/iothubtransport_amqp_cbs_auth.c
```diff
@@ -29,7 +29,7 @@ typedef struct AUTHENTICATION_INSTANCE_TAG
     ON_AUTHENTICATION_ERROR_CALLBACK on_error_callback;
     void* on_error_callback_context;
 
-    size_t cbs_request_timeout_secs;
+    time_t cbs_request_timeout_secs;
```
ewertons, line 32: It would be better not to make this change.
First because the result of difftime is not of time_t type. Second because time_t is not an arithmetical type.
And also there is no guarantee that time_t will be a time in seconds, like the name of the variable suggests.

ewertons (changes requested):
There is not

iothub_client/src/iothubtransport_amqp_cbs_auth.c
CIPop: Changed to `uint64_t`

ewertons (approved):

CIPop:
C-raspberrypi fails with setup problems. (I've removed the "Required" field until we figure out the fix.)

PR #1810, opened 2020-12-28
Merge latest master into distributed tracing branch

ewertons:
# Checklist

# Description of the problem

# Description of the solution


build_all/docs/Doxyfile
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           = "Microsoft Azure IoT Device SDK for C"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 1.3.9
+PROJECT_NUMBER         = 1.6.0
```
CIPop, line 41: Shouldn't this be something higher? We are already at 1.6.0 in `master`.

iothub_client/inc/internal/iothub_internal_consts.h
```diff
@@ -9,11 +9,7 @@ extern "C"
 {
 #endif
 
-    static const char* IOTHUB_API_VERSION = "2019-10-01";
-    // TODO: https://github.com/Azure/azure-iot-sdk-c/issues/1547 tracks removing this preview 
-    // variable once the underlying logic is enabled on all IoTHubs.
-    static const char* IOTHUB_API_PREVIEW_VERSION = "2020-05-31-preview";
-
+    static const char* IOTHUB_API_VERSION = "2020-09-30";
```
CIPop, line 12: Does this have full-support for distributed tracing or do we need to use a different -preview APIversion?

CIPop (approved):
Left a couple of comments on versions.

build_all/docs/Doxyfile
ewertons: For public-preview we usually keep the same version as in master. Maybe another point to fix on the versioning strategy?

iothub_client/inc/internal/iothub_internal_consts.h
ewertons: That's something that needs to be verified. Notice that "IOTHUB_API_PREVIEW_VERSION" was introduced by John for pnp, so it is not related to distributed-tracing.

PR #1834, opened 2021-01-23
release_2021_01_21_after_bump_version

momuno:


ericwol-msft (approved):

Timple:
Any particular reason these three-letter versions are not tagged anymore?

momuno:
Hi @Timple. Today we only tag when releasing from master; version tags are not added to LTS releases.  It does appear the version was mistakenly bumped to 1.7.0.  It should be 1.6.0 and is being fixed.

momuno:
Hi @Timple. We are currently verifying this.  Thanks for bringing this to our attention.

momuno:
Hi @Timple.  It looks like the last tagged version was 1.4.0 and 1.3.9.  I'm not sure why we stopped tagging them. I am working on the next release and will make sure to tag the version! Thanks again.

Timple:
It's easy to tag in hindsight as well, so one could still do this. :slightly_smiling_face:

momuno:
> Thanks for not forgetting about it.
> It's easy to tag in hindsight as well, so one could still do this. 🙂

We're on the same wavelength :)  All should be updated.

PR #1836, opened 2021-01-27
[SDL] NULL pointer check

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution


samples/solutions/remote_monitoring_client/remote_monitoring.c
```diff
@@ -91,33 +91,37 @@ typedef struct CHILLER_TAG
 /*  Converts the Chiller object into a JSON blob with reported properties ready to be sent across the wire as a twin. */
 static char* serializeToJson(Chiller* chiller)
 {
-	char* result;
+	char* result = "{}";
```
ewertons, line 94: So is this really going to work?
If json_serialize_to_string(root_value); on line 121 returns a dynamically allocated string, now this function returns both a static and dynamic piece of mem. Whoever frees the result (if it is the case) will have trouble.
Did you check that?

ewertons (approved):
Please look at the one comment I left. Thanks.

samples/solutions/remote_monitoring_client/remote_monitoring.c
ericwol-msft: good catch

PR #1857, opened 2021-02-11
update http set option to convert http return result

### LINKED ISSUE #1843: Cipher suite option for DPS
Found out that there are no option to set cipher suite for DPS in provisioning_client src code.

While for iothub_client, I tried the option called "OPTION_OPENSSL_CIPHER_SUITE" and successfully set the hardcoded cipher suite without any issue.
IoTHubDeviceClient_LL_SetOption(iotHubClientHandle, OPTION_OPENSSL_CIPHER_SUITE, "hardcodedciphersuite")

Is there an option or does it have a way to set cipher suite for DPS? Thanks

danewalton:
fixes #1843

### DIFF
--- a/provisioning_client/src/prov_transport_http_client.c
+++ b/provisioning_client/src/prov_transport_http_client.c
@@ -1231,7 +1231,15 @@ static int prov_transport_http_set_option(PROV_DEVICE_TRANSPORT_HANDLE handle, c
         }
         else
         {
-            result = uhttp_client_set_option(http_info->http_client, option, value);
+            HTTP_CLIENT_RESULT http_result = uhttp_client_set_option(http_info->http_client, option, value);
+            if (http_result != HTTP_CLIENT_OK)
+            {
+                result = MU_FAILURE;
+            }
+            else
+            {
+                result = 0;
+            }
         }
     }
     return result;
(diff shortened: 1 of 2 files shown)


ericwol-msft (approved):

provisioning_client/src/prov_transport_http_client.c
```diff
@@ -1231,7 +1231,15 @@ static int prov_transport_http_set_option(PROV_DEVICE_TRANSPORT_HANDLE handle, c
         }
         else
         {
-            result = uhttp_client_set_option(http_info->http_client, option, value);
+            HTTP_CLIENT_RESULT http_result = uhttp_client_set_option(http_info->http_client, option, value);
+            if (http_result != HTTP_CLIENT_OK)
+            {
+                result = MU_FAILURE;
```
momuno, line 1237: This looks good. Would it be worth having a LogError for the MU_FAILURE scenario?

momuno (approved):

provisioning_client/src/prov_transport_http_client.c
danewalton: It didn't have it before so I'd like to leave as is. There are already a good amount of log messages that bubble up the stack

ericwol-msft (approved):

PR #1919, opened 2021-03-26
Refactor iothub_client CMakes including samples

momuno:
# Checklist

This is a refactor of the main CMakeLists.txt files in iothub_client and its samples.


certs/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+#Copyright (c) Microsoft. All rights reserved.
```
ewertons, line 1: Hey Mollie, what is the technical need for this? Can't the executables just continue including the source files directly?

CMakeLists.txt
```diff
@@ -131,6 +131,10 @@ if (${use_portal_azure_cn_cert})
     add_definitions(-DUSE_PORTAL_AZURE_CN_CERT)
 endif()
 
+if(${use_sample_trusted_cert})
```
ewertons, line 134: Related question below (what is the technical need for this?)

ewertons (commented):
So are we actually adding any functionality or this is just a refactoring of the cmakelists? Or is it fixing any specific bug?

CMakeLists.txt
ewertons: Also you mentioned that with these changes parson would not be linked with iothub_client anymore, just the samples that use it, unless using u2b. But I could not find any change in any non-sample CMakelists.txt's restricting the linking of parson. Was that a bug? Meaning: if u2b was disabled and I tried to build the samples that need parson, the build would fail?

certs/CMakeLists.txt
momuno: You are correct, this is not required and does not change things technically  The reason I added this is because I think it makes the Sample CMakes easier to follow without the cert if-statement in the middle.  I also prefer the design of having a CMake for each directory, because I also think its easier to follow.  Putting it forward as an idea/proposal.

CMakeLists.txt
momuno: In the prior iothub_client CMake, parson is always linked to the iothub_client library (static and dynamic). 

If samples are being built:
Those samples that depend on parson would always have it compiled via the iothub_client library regardless of if u2b or edge are enabled.  

If samples are not being built:
If u2b and edge are disabled, their associated files would not be added to the iothub_client library (static or dynamic).  So in a static build of the iothub_client library, linking to parson would have no effect (unnecessary linking).  But in a dynamic build of the iothub_client library, parson would be present, creating a larger .so library than necessary.  It is this dynamic build scenario that I want to fix -- If customers want to build iothub_client dynamically and don't want parson included, they can do so now by disabling u2b and edge (assuming their application doesn't depend on those files.)  But with this change, the samples that depend on parson won't have it always compiled in.  So the samples that depend on parson need to link to it explicitly.

momuno: the change in iothub_client CMake is at lines 320-322 for static library linking and at 368-370 for dynamic library linking

momuno: I compared the size of the iothub_client_dll (using original iothub_client CMake) with and without parson included.  A difference of about 35.7K.

momuno: I did some more reading, I was wrong.  I was under the impression that `target_link_libraries(iothub_client_dll parson)` would add parson to the .so library.  It does not.  Only the files included during `add_library(iothub_client_dll SHARED {src_files_here})` are included.  What `target_link_libraries` does in this case is adds `libparson.a` to the link line for other targets that use `target_link_libraries({target_here} iothub_client_dll)`.

I will re-examine my changes and update.

momuno:
Needs rework.  Moving to forked repo.

Issue #1647, opened 2020-09-09
C API should allow asynchronous acknowledgement of incoming messages


alaendle:
The C API design leads to poor performance and therefore isn't suitable for high throughput/streaming scenarios.

Let me explain the problem in more detail: Currently the message handler callback (https://docs.microsoft.com/en-us/azure/iot-hub/iot-c-sdk-ref/iothub-client-core-common-h#iothubclientmessagecallbackasync) needs to return the result in a synchronous way. But think of use cases where handling the incoming message involves (heavy) I/O (like storing in a database or transferring to a message broker) - in almost any cases you would use a asynchronous API to do the I/O and provide some kind of completion callback to get the information if the operation completed successfully. And here the problem arises, with the current C API design you need to wait for the completion callback to determine the correct return value for the C SDK callback.

So please add methods that allow the ack/nack of messages independently from the message handler callback and remove the result value from the callback signature (or at least add a value like NONE to indicate that information about the success will be given at a later time).

At least conceptually I see no problems to provide an asynchronous API that enables windowing, because the underlying transport protocol works inherently in a asynchronous way - and also the C# SDK provides corresponding methods (e.g. https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client.moduleclient.completeasync?view=azure-dotnet).

For now I keep short, but please don't hesitate to contact me if I was unclear or if I could contribute more/other information that may help you to understand/solve the problem. Thanks in advance.

Please note that I'm unsure if this is a bug report or a feature request, because on the one hand the performance that can be achieved with the current API is worse, on the other hand this issue requests changes to the design that could be seen as a feature. And also note that I've addressed the same problem at [feedback.azure.com](https://feedback.azure.com/forums/321918-azure-iot-hub-dps-sdks/suggestions/41349889-c-api-should-allow-asynchronous-acknowledgement-of); not to double the work at Microsoft, but because I hope that the different platforms trigger different support levels that might lead to different interpretations (e.g. a more technical perspective here at github and more a product management view on feedback.azure.com).

danewalton:
Hi @alaendle quick question here.
As an alternative design, would it be possible to use the callback as is and queue the processing of the messages. IE decouple the acknowledgement that the message is coming in and the actual handling of the message itself? (I might be saying the same thing as you here). 

Just to clarify I think: You would like to acknowledge the message at the MQTT level and at the IoT Hub layer at different points in execution?

alaendle:
[...]

PR #1994, opened 2021-06-15
Add support for sending async responses for C2D messages

ewertons:
# Checklist

https://github.com/Azure/azure-iot-sdk-c/issues/1647

# Description of the problem
Customer has requested a way to send ACK/DISPOSITION for C2D messages asynchronously, outside the stack that leads to the callback that delivers the incoming message.

# Description of the solution
Added corresponding APIs for achieving the functionality requested.


iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -43,13 +43,55 @@
...
+        messageId = "<unavailable>";
+    }
+
+    (void)printf("Sending ACK for cloud message (Message ID: %s)\r\n", messageId);
+
+    if (IoTHubDeviceClient_LL_SendMessageDisposition(device_ll_handle, message, IOTHUBMESSAGE_ACCEPTED) != IOTHUB_CLIENT_OK)
+    {
+        (void)printf("ERROR: IoTHubDeviceClient_LL_SendMessageDisposition..........FAILED!\r\n");
+        // If (and only if) IoTHubDeviceClient_LL_SendMessageDisposition fails, the IOTHUB_MESSAGE_HANDLE must be explicitly destroyed.
+        IoTHubMessage_Destroy(message);
```
CIPop, line 80: Having either the SDK or the app manage memory is confusing. 
In this case it's probably better to assert / crash instead of deallocating something that will definitely block the protocol maybe causing Hub to disconnect. 

Going out of sync on MQTT or AMQP is a dangerous situation.

ewertons: Changes made to sample and documentation.

iothub_client/samples/iothub_convenience_sample/iothub_convenience_sample.c
```diff
@@ -57,6 +57,14 @@ multithreaded API is that the calls to DoWork are abstracted away from your code
 #include "certs.h"
 #endif // SET_TRUSTED_CERT_IN_SAMPLES
 
+// Uncomment this define to use Asynchronous ACK of Cloud-to-Device messages.
+#define USE_C2D_ASYNC_ACK
```
danewalton, line 61: should this be commented out

iothub_client/samples/iothub_client_sample_module_filter/iothub_client_sample_module_filter.c
```diff
@@ -14,6 +14,13 @@
 #include "iothubtransportmqtt.h"
 #include "iothub.h"
 
+// Uncomment this define to use Asynchronous ACK of Cloud-to-Device messages.
+#define USE_C2D_ASYNC_ACK
```
danewalton, line 18: commented out?

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -43,13 +43,53 @@
...
+    if ((messageId = IoTHubMessage_GetMessageId(message)) == NULL)
+    {
+        messageId = "<unavailable>";
+    }
+
+    (void)printf("Sending ACK for cloud message (Message ID: %s)\r\n", messageId);
+
+    if (IoTHubDeviceClient_LL_SendMessageDisposition(device_ll_handle, message, IOTHUBMESSAGE_ACCEPTED) != IOTHUB_CLIENT_OK)
+    {
+        (void)printf("ERROR: IoTHubDeviceClient_LL_SendMessageDisposition..........FAILED!\r\n");
```
danewalton, line 78: delete the message here? and same comments as before about returning true/false etc

iothub_client/src/iothub_client_core.c
```diff
@@ -2635,3 +2641,45 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_GenericMethodInvoke(IOTHUB_CLIENT_CORE_HAN
...
+            {
+                result = IoTHubClientCore_LL_SendMessageDisposition(iotHubClientInstance->IoTHubClientLLHandle, message, disposition);
+
+                if (result != IOTHUB_CLIENT_OK)
+                {
+                    LogError("IoTHubClientCore_LL_SendMessageDisposition failed");
+                }
+
+                /* Codes_SRS_IOTHUBCLIENT_01_025: [IoTHubClient_SendEventAsync shall be made thread-safe by using the lock created in IoTHubClient_Create.] */
+                (void)Unlock(iotHubClientInstance->LockHandle);
```
danewalton, line 2679: put the unlock right after line 2671 as above? shortest time holding lock...

iothub_client/src/iothub_client_core.c
```diff
@@ -2635,3 +2641,45 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_GenericMethodInvoke(IOTHUB_CLIENT_CORE_HAN
...
+            }
+        }
+    }
+
+    return result;
+}
```
danewalton, line 2685: newline

iothub_client/src/iothub_module_client_ll.c
```diff
@@ -355,3 +355,18 @@ IOTHUB_CLIENT_RESULT IoTHubModuleClient_LL_ModuleMethodInvoke(IOTHUB_MODULE_CLIE
...
+IOTHUB_CLIENT_RESULT IoTHubModuleClient_LL_SendMessageDisposition(IOTHUB_MODULE_CLIENT_LL_HANDLE iotHubModuleClientHandle, IOTHUB_MESSAGE_HANDLE message, IOTHUBMESSAGE_DISPOSITION_RESULT disposition)
+{
+    IOTHUB_CLIENT_RESULT result;
+    if (iotHubModuleClientHandle != NULL)
+    {
+        result = IoTHubClientCore_LL_SendMessageDisposition(iotHubModuleClientHandle->coreHandle, message, disposition);
+    }
+    else
+    {
+        LogError("Input parameter cannot be NULL");
```
danewalton, line 368:
```suggestion
        LogError("iotHubModuleClientHandle cannot be NULL");
```

iothub_client/src/iothubtransport_amqp_common.c
```diff
@@ -467,24 +412,36 @@ static DEVICE_MESSAGE_DISPOSITION_RESULT get_device_disposition_result_from(IOTH
 
 static DEVICE_MESSAGE_DISPOSITION_RESULT on_message_received(IOTHUB_MESSAGE_HANDLE message, DEVICE_MESSAGE_DISPOSITION_INFO* disposition_info, void* context)
 {
-    AMQP_TRANSPORT_DEVICE_INSTANCE* amqp_device_instance = (AMQP_TRANSPORT_DEVICE_INSTANCE*)context;
     DEVICE_MESSAGE_DISPOSITION_RESULT device_disposition_result;
-    MESSAGE_CALLBACK_INFO* message_data;
+    MESSAGE_DISPOSITION_CONTEXT* dispositionContext;
```
danewalton, line 416: nit: can this be called 
```suggestion
    MESSAGE_DISPOSITION_CONTEXT* dispositionContextClone;
```
makes it a little easier to follow and understand what is being destroyed and set

iothub_client/src/iothubtransport_amqp_common.c
```diff
@@ -2492,57 +2419,49 @@ STRING_HANDLE IoTHubTransport_AMQP_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+        MESSAGE_DISPOSITION_CONTEXT_HANDLE device_message_disposition_info;
+
         /* Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_10_002: [If any of the messageData fields are NULL, IoTHubTransport_AMQP_Common_SendMessageDisposition shall fail and return IOTHUB_CLIENT_INVALID_ARG.] */
-        if (message_data->messageHandle == NULL || message_data->transportContext == NULL)
+        // Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_09_112: [A DEVICE_MESSAGE_DISPOSITION_INFO instance shall be created with a copy of the `link_name` and `message_id` contained in `message_data`]
+        if (IoTHubMessage_GetDispositionContext(message_handle, &device_message_disposition_info) != IOTHUB_MESSAGE_OK ||
+            device_message_disposition_info == NULL)
         {
-            LogError("Failed sending message disposition (message_data->messageHandle (%p) or message_data->transportContext (%p) are NULL)", message_data->messageHandle, message_data->transportContext);
+            // Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_09_113: [If the DEVICE_MESSAGE_DISPOSITION_INFO fails to be created, `IoTHubTransport_AMQP_Common_SendMessageDisposition()` shall fail and return IOTHUB_CLIENT_ERROR]
```
danewalton, line 2440: ?
```suggestion
            // Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_09_113: [If the DEVICE_MESSAGE_DISPOSITION_INFO fails to be created, `IoTHubTransport_AMQP_Common_SendMessageDisposition()` shall fail and return IOTHUB_CLIENT_INVALID_ARG]
```

iothub_client/src/iothubtransport_amqp_common.c
```diff
@@ -2492,57 +2419,49 @@ STRING_HANDLE IoTHubTransport_AMQP_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+            LogError("Invalid message handle (no disposition context found)");
             result = IOTHUB_CLIENT_INVALID_ARG;
```
danewalton, line 2442: or should this now be `IOTHUB_CLIENT_ERROR`

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1750,42 +1780,49 @@ static void processIncomingMessageNotification(PMQTTTRANSPORT_HANDLE_DATA transp
...
     else if (extractMqttProperties(IoTHubMessage, topic_resp, transportData->auto_url_encode_decode) != 0)
     {
         LogError("failure extracting mqtt properties.");
+        IoTHubMessage_Destroy(IoTHubMessage);
     }
     else
     {
-        MESSAGE_CALLBACK_INFO* messageData = (MESSAGE_CALLBACK_INFO*)malloc(sizeof(MESSAGE_CALLBACK_INFO));
-        if (messageData == NULL)
+        // TODO: try sending disposition if anything fails.
```
danewalton, line 1792: still a todo?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -3858,32 +3897,46 @@ STRING_HANDLE IoTHubTransport_MQTT_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+
+                if (mqtt_client_send_message_response(transport_data->mqttClient, msgDispCtx->packet_id, msgDispCtx->qos_value) != 0)
+                {
+                    LogError("Failed sending ACK for MQTT message (packet_id=%u)", msgDispCtx->packet_id);
+                    result = IOTHUB_CLIENT_ERROR;
+                }
+            }
+        }
+
+        // This will destroy also the disposition context;
```
danewalton, line 3936: nit: 
```suggestion
        // This will also destroy the disposition context;
```

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -3858,32 +3897,46 @@ STRING_HANDLE IoTHubTransport_MQTT_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+        /*Codes_SRS_IOTHUB_MQTT_TRANSPORT_10_001: [If messageData is NULL, IoTHubTransport_MQTT_Common_SendMessageDisposition shall fail and return IOTHUB_CLIENT_ERROR. ]*/
+        LogError("Invalid argument (handle=%p, messageHandle=%p", handle, messageHandle);
+        result = IOTHUB_CLIENT_INVALID_ARG;
     }
     else
     {
-        /*Codes_SRS_IOTHUB_MQTT_TRANSPORT_10_001: [If messageData is NULL, IoTHubTransport_MQTT_Common_SendMessageDisposition shall fail and return IOTHUB_CLIENT_ERROR. ]*/
-        LogError("message_data is NULL");
-        result = IOTHUB_CLIENT_ERROR;
+        result = IOTHUB_CLIENT_OK;
```
danewalton, line 3911: just double checking that the impl is correct here. If `disposition` is anything other than `IOTHUBMESSAGE_ACCEPTED`, then the message should be destroyed and the result should be `IOTHUB_CLIENT_OK`?

iothub_client/tests/iothubdeviceclient_ll_ut/iothubdeviceclient_ll_ut.c
```diff
@@ -410,5 +411,18 @@ TEST_FUNCTION(IoTHubDeviceClient_LL_UploadMultipleBlocksToBlob_Test)
...
+
+    //act
+    IOTHUB_CLIENT_RESULT result = IoTHubDeviceClient_LL_SendMessageDisposition(TEST_IOTHUB_DEVICE_CLIENT_LL_HANDLE, TEST_MESSAGE_HANDLE, IOTHUBMESSAGE_ACCEPTED);
+
+    //assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_TRUE(IOTHUB_CLIENT_OK == result);
+}
 
 END_TEST_SUITE(iothubdeviceclient_ll_ut)
```
danewalton, line 428: nit: newline even though not part of this PR

iothub_client/tests/iothubtransporthttp_ut/iothubtransporthttp_ut.c
```diff
@@ -3532,12 +3554,15 @@ TEST_FUNCTION(IoTHubTransportHttp_DoWork_happy_path_with_2_registered_devices_em
...
 //Tests_SRS_TRANSPORTMULTITHTTP_17_090: [ All the HTTP headers of the form iothub-app-name:somecontent shall be transformed in message properties {name, somecontent}. ]
 //Tests_SRS_TRANSPORTMULTITHTTP_17_123: [ After client creation, the first GET shall be allowed no matter what the value of GetMinimumPollingTime. ]
-TEST_FUNCTION(IoTHubTransportHttp_DoWork_happy_path_with_empty_waitingToSend_and_1_service_message_succeeds)
+TEST_FUNCTION(IoTHubTransportHttp_DoWork_happy_path_with_empty_waitingToSend_and_1_service_message_fail_set_disposition_fails)
 {
     //arrange
     unsigned int statusCode200 = 200;
     unsigned int statusCode204 = 204;
 
+    char* real_ETAG = (char*)malloc(sizeof(TEST_ETAG_VALUE) + 1);
```
danewalton, line 3563: nit: check for `NULL`?

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -6488,7 +6523,57 @@ TEST_FUNCTION(IoTHubTransport_MQTT_Common_MessageRecv_with_Properties_succeed)
     IoTHubTransport_MQTT_Common_DoWork(handle);
     umock_c_reset_all_calls();
 
-    setup_message_recv_with_properties_mocks(true, true, false);
+    setup_message_recv_with_properties_mocks(true, true, false, true);
```
danewalton, line 6526: nit: feel free to reject this one but it's difficult to parse what each of these represents without going to the function declaration. Consider having local vars for each which you set right above to make it easy to follow what each boolean is setting.

danewalton: here and elsewhere below

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
```diff
@@ -7976,48 +8074,109 @@ TEST_FUNCTION(IoTHubTransport_MQTT_Common_DeviceMethod_Response_fail)
...
+    umock_c_reset_all_calls();
+    STRICT_EXPECTED_CALL(IoTHubMessage_GetDispositionContext(IGNORED_PTR_ARG, IGNORED_PTR_ARG))
+        .CopyOutArgumentBuffer_dispositionContext(&msgDispCtxHandle, sizeof(&msgDispCtxHandle));
+    STRICT_EXPECTED_CALL(mqtt_client_send_message_response(IGNORED_PTR_ARG, IGNORED_NUM_ARG, IGNORED_NUM_ARG));
+    STRICT_EXPECTED_CALL(IoTHubMessage_Destroy(IGNORED_PTR_ARG));
+    STRICT_EXPECTED_CALL(gballoc_free(IGNORED_PTR_ARG));
 
     // act
-    IOTHUB_CLIENT_RESULT result = IoTHubTransport_MQTT_Common_SendMessageDisposition(md, IOTHUBMESSAGE_ACCEPTED);
+    IOTHUB_CLIENT_RESULT result = IoTHubTransport_MQTT_Common_SendMessageDisposition(devHandle, (IOTHUB_MESSAGE_HANDLE)0x42, IOTHUBMESSAGE_ACCEPTED);
```
danewalton, line 8172: 42!!!!!
🚀

danewalton (approved):
a lot of work here Ewerton thanks for doing it :)

iothub_client/inc/internal/iothub_message_private.h
```diff
@@ -0,0 +1,53 @@
...
+
+struct MESSAGE_DISPOSITION_CONTEXT_TAG;
+typedef struct MESSAGE_DISPOSITION_CONTEXT_TAG* MESSAGE_DISPOSITION_CONTEXT_HANDLE;
+
+typedef void(*MESSAGE_DISPOSITION_CONTEXT_DESTROY_FUNCTION)(MESSAGE_DISPOSITION_CONTEXT_HANDLE dispositionContext);
+
+/**
+* @brief   Sets the context for the transport layer to send a DISPOSITION or ACK for a Cloud-to-Device message.
+*
+* @param   iotHubMessageHandle                The message to have the context set on.
```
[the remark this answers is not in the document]
ewertons: Taken.

iothub_client/inc/internal/iothub_message_private.h
```diff
@@ -0,0 +1,53 @@
...
+* @param   dispositionContext                 The transport context for disposition.
+* @param   dispositionContextDestroyFunction  A function defined by the transport for destroying the context instance.
+*
+* @return  If any argument is NULL, returns a IOTHUBMESSAGE_INVALID_ARG, or IOTHUBMESSAGE_OK otherwise.
```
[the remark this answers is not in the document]
ewertons: Done.

iothub_client/inc/internal/iothubtransport_amqp_device.h
```diff
@@ -125,7 +125,8 @@ MOCKABLE_FUNCTION(, int, amqp_device_send_message_disposition, AMQP_DEVICE_HANDL
 MOCKABLE_FUNCTION(, int, amqp_device_set_retry_policy, AMQP_DEVICE_HANDLE, handle, IOTHUB_CLIENT_RETRY_POLICY, policy, size_t, retry_timeout_limit_in_seconds);
 MOCKABLE_FUNCTION(, int, amqp_device_set_option, AMQP_DEVICE_HANDLE, handle, const char*, name, void*, value);
 MOCKABLE_FUNCTION(, OPTIONHANDLER_HANDLE, amqp_device_retrieve_options, AMQP_DEVICE_HANDLE, handle);
-
+MOCKABLE_FUNCTION(, DEVICE_MESSAGE_DISPOSITION_INFO*, clone_device_message_disposition_info, DEVICE_MESSAGE_DISPOSITION_INFO*, disposition_info);
+MOCKABLE_FUNCTION(, void, destroy_device_message_disposition_info, DEVICE_MESSAGE_DISPOSITION_INFO*, disposition_info);
```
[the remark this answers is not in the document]
ewertons: yeah, that would make sense. I'll do as suggested.

iothub_client/inc/iothub_client_core.h
```diff
@@ -61,6 +61,9 @@ extern "C"
     MOCKABLE_FUNCTION(, IOTHUB_CLIENT_RESULT, IoTHubClientCore_GenericMethodInvoke, IOTHUB_CLIENT_CORE_HANDLE, iotHubClientHandle, const char*, deviceId, const char*, moduleId, const char*, methodName, const char*, methodPayload, unsigned int, timeout, IOTHUB_METHOD_INVOKE_CALLBACK, methodInvokeCallback, void*, context);
 #endif /* USE_EDGE_MODULES */
 
+    // TODO: document!
```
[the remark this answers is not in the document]
ewertons: That's done in the public api. Thanks.

iothub_client/inc/iothub_device_client.h
```diff
@@ -349,6 +349,24 @@ extern "C"
 
 #endif /* DONT_USE_UPLOADTOBLOB */
 
+    /**
+    * @brief    This API sends an acknowledgement to Azure IoT Hub that a cloud-to-device message has been received.
+    *
+    * @param    device_ll_handle                The handle created by a call to a create function.
```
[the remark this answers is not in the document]
ewertons: D'Oh!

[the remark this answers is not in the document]
ewertons: Done.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: yeah, that's how the upper layers are using the API, so since it's internal I kept it as is.

iothub_client/tests/iothubdeviceclient_ll_ut/iothubdeviceclient_ll_ut.c
ewertons: Fixed.

iothub_client/tests/iothubtransporthttp_ut/iothubtransporthttp_ut.c
ewertons: On tests we are more lax, since they are usually run on higher capacity platforms and malloc failures would be really unexpected.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
ewertons: I agree. Unfortunately that is disseminated in the C SDK tests, so I rather keep that as a separate task if we end up having time to document.

iothub_client/tests/iothubtransport_mqtt_common_ut/iothubtransport_mqtt_common_ut.c
ewertons: Hooray!

iothub_client/inc/internal/iothubtransport_amqp_device.h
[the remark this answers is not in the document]
ewertons: Done.

iothub_client/inc/iothub_device_client_ll.h
```diff
@@ -374,6 +374,24 @@ typedef struct IOTHUB_CLIENT_CORE_LL_HANDLE_DATA_TAG* IOTHUB_DEVICE_CLIENT_LL_HA
 
 #endif /*DONT_USE_UPLOADTOBLOB*/
 
+    /**
+    * @brief    This API sends an acknowledgement to Azure IoT Hub that a cloud-to-device message has been received.
```
jspaith, line 378: Per comments about need to call this for freee, I'd put this into the @brief because it's a critical part of what this function is doing.

`This API sends an acknowledgement to Azure IoT Hub that a cloud-to-device message has been received and frees resources associated with the message` 
?

iothub_client/inc/iothub_device_client_ll.h
```diff
@@ -374,6 +374,24 @@ typedef struct IOTHUB_CLIENT_CORE_LL_HANDLE_DATA_TAG* IOTHUB_DEVICE_CLIENT_LL_HA
...
+    *
+    * @param    device_ll_handle                The handle created by a call to a create function.
+    * @param    message                         The cloud-to-device message received through the callback provided to IoTHubDeviceClient_LL_SetMessageCallback.
+    * @param    disposition                     Acknowledgement option for the message.
+    *
+    * @warning  This function is to be used only when IOTHUBMESSAGE_ASYNC_ACK is used in the callback for incoming Cloud-to-Device messages.
+    * @remarks
+    *           For Cloud-to-Device messages, the IOTHUB_MESSAGE_HANDLE instance is allocated by the Azure IoT C SDK
```
jspaith, line 386: OK so I *love* this note and we're thinking about memory lifetimes and helping customer understand.  I just worry that the who own what isn't entirely clear, along same lines as Dane feedback of "it" unclear.  

I think we should maybe make it more explicit what we want the app to do (namely call this function) so we lead with headline and not get into the details as much.  (We're open source so if they're curious on this level of detail they can read themselves:)).

Maybe something like
* If your cloud-to-device message callback returned IOTHUBMESSAGE_ASYNC_ACK, it MUST call this API eventually.
*  Beyond sending acknowledgment to the service, this method also handles freeing message's memory.
* Not calling this function will result in memory leaks.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1750,42 +1780,49 @@ static void processIncomingMessageNotification(PMQTTTRANSPORT_HANDLE_DATA transp
...
+        // TODO: try sending disposition if anything fails.
+        MESSAGE_DISPOSITION_CONTEXT* dispositionContext = createMessageDispositionContext(msgHandle);
+
+        if (dispositionContext == NULL)
         {
-            LogError("malloc failed");
+            LogError("failed creating message disposition context");
+            IoTHubMessage_Destroy(IoTHubMessage);
+        }
+        else if (IoTHubMessage_SetDispositionContext(IoTHubMessage, dispositionContext, destroyMessageDispositionContext) != IOTHUB_MESSAGE_OK)
```
jspaith, line 1800: I recommend changing how the memory/owership model here works.  Basically MQTT (and other layers) do initial malloc() but then do a de facto type "release()" of memory into the IoTHub message handle.  But because Hub layer didn't do initial malloc(), it means that the free() has to be delegated back to the layer here.  I just worry this really produces too tight a coupling between everything, and cleanup issues tend to crop up.

I'd recommend - fairly strongly actually - keeping with the existing IoTHubMessage pattern.  So we leave IoTHubMesasage itself to make a copy of whatever data there is and the caller only ever has to call IoTHub_MessageDestroy() to make it go away.  

So basically API call turns into something like below, and yes I know my naming of vars is awful.

```c
IoTHubMessage_SetDispositionContext(hubMessage, unsigned const char* data, size_t dataLen) {
    hubMessage->disposition = malloc(dataLen); memcpy(hubMessage->disposition, dataLen);
}
```

So this is super easy on MQTT beacues it's just a flat buffer.  If we keep HTTP and/or AMQP I looked at that code and it's not quite as easy as there's pointers.  But I think you could do something like

```c
struct AMQP_CONTEXT {
   int whateverDataWeNeed;
   char extraData[]
};
AMQP_CONTEXT*ctxt = (AMQP_CONTEXT*)malloc(sizeof(AMQP_CONTEXT) + strlen(extraDataToCopy) + 1);
ctxt->whateverDataWeNeed = intValueStuff;
strcpy(ctxt->extraData, extraDataToCopy);

MessageSetDisposition(hnadle, ctxt, ctxtLenWeMallocdBasedOnAbove);
```

You'll have to be careful because IIRC some compliers don't really like the [] business at end of struct, so it may need to be a [1].  

And yeah I 100% get this is more complicated on the initial allocation and also on the dereference.  But the complexity is isolated to alloc/use, and there's not extra interdpendencies between the layers that will be way harder to debug if anything goes wrong.

iothub_client/src/iothub_client_core_ll.c
```diff
@@ -633,21 +637,15 @@ static const char* IoTHubClientCore_LL_GetModelId(void* ctx)
     return result;
 }
 
-static bool IoTHubClientCore_LL_MessageCallbackFromInput(MESSAGE_CALLBACK_INFO* messageData, void* ctx)
+static bool IoTHubClientCore_LL_MessageCallbackFromInput(IOTHUB_MESSAGE_HANDLE messageHandle, void* ctx)
 {
     bool result;
     IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* handleData = (IOTHUB_CLIENT_CORE_LL_HANDLE_DATA*)ctx;
```
jspaith, line 643: Nit - because we have two handles now - the message handle and core_ll_handle - I think `handleData` is ambiguous.  I'd think something like `clientHandleData` for this variable?

Similiar feedback in `IoTHubClientCore_LL_MessageCallback`.

jspaith (changes requested):
OK I've gone through API/MQTT/clientCore layers/sample.  I have other offline questions RE AMQP|HTTP before I get too deeply into either of those.

Requesting changes mostly for stuff from my previous comments.

iothub_client/src/iothub_client_core_ll.c
ewertons: Makes sense. Done.

iothub_client/src/iothub_module_client_ll.c
ewertons: Done.

iothub_client/src/iothubtransport_amqp_common.c
ewertons: Sure. Done.

iothub_client/src/iothubtransport_amqp_common.c
ewertons: Done

iothub_client/src/iothubtransport_amqp_common.c
ewertons: It makes sense for it to be a INVALID_ARG, since this is not expected to happen because of an internal SDK error.
The only situation when this would happen is if the customer creates or clones an IOTHUB_MESSAGE_HANDLE and passes it down to *SendMessageDisposition(...), which would be an invalid arg.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: As we talked offline, this async disposition construct is also implemented for the AMQP transport, which has 2 pieces of data for the disposition to be sent (link name and message id).
If we decide to change this design, we will need to change how amqp passes the disposition context (maybe serialize the 2 parameters and set that into the IOTHUB_MESSAGE_HANDLE, which would be easy to free() when IoTHubMessage_Destroy is called).

However... All that this design is doing is letting the transports to pass a function that knows how to free the disposition context they are setting in the IOTHUB_MESSAGE_HANDLE. There is no other action done by these callbacks. Please reconsider keeping this design, which is not as convoluted as it might first look.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Taken.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -1879,70 +1908,58 @@ static void processIncomingMessageNotification(PMQTTTRANSPORT_HANDLE_DATA transp
...
+                if (!transportData->transport_callbacks.msg_input_cb(IoTHubMessage, transportData->transport_ctx))
                 {
-                    // Cleanup of allocated memory happens at end of function as pointers are non-NULL.
                     LogError("IoTHubClientCore_LL_MessageCallbackFromInput returned false");
-                }
-                else
-                {
+                    // This will destroy the dispostion context;
+                    IoTHubMessage_Destroy(IoTHubMessage);
                     IoTHubMessage = NULL;
```
jspaith, line 1938: Please remove this `IoTHubMessage = NULL` and the one in block a bit below.  This was needed the way code was previously structured where a non-NULL would get freed in closing block, but not now.

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -4006,32 +4025,57 @@ STRING_HANDLE IoTHubTransport_MQTT_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
-
     IOTHUB_CLIENT_RESULT result;
-    if (message_data)
+    if (handle == NULL || messageHandle == NULL)
+    {
+        /*Codes_SRS_IOTHUB_MQTT_TRANSPORT_10_001: [If messageData is NULL, IoTHubTransport_MQTT_Common_SendMessageDisposition shall fail and return IOTHUB_CLIENT_ERROR. ]*/
+        LogError("Invalid argument (handle=%p, messageHandle=%p", handle, messageHandle);
+        result = IOTHUB_CLIENT_INVALID_ARG;
+    }
+    else if (disposition == IOTHUBMESSAGE_ASYNC_ACK)
```
jspaith, line 4037: Is `IOTHUBMESSAGE_ASYNC_ACK` ever legit to be set at this point for any protocol?  I don't think so.  If so could we move this check to the clientCore layer, since I don't see HTTP or AMQP guarding against this (and clientCore could for all if I'm correct).

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -4006,32 +4025,57 @@ STRING_HANDLE IoTHubTransport_MQTT_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+    {
+        LogError("IOTHUBMESSAGE_ASYNC_ACK is not a valid disposition value for this function");
+        result = IOTHUB_CLIENT_INVALID_ARG;
+    }
+    else
     {
-        if (message_data->messageHandle)
+        if (disposition == IOTHUBMESSAGE_ACCEPTED)
```
jspaith, line 4044: Nit - I think this is moot because of my comments removing IOTHUBMESSAGE_ASYNC_ACK check in this layer above.  But I recommend the pattern of

if (...)
else if (...)
else if(...)

As much as possible, or else the nesting of else's to my mind is harder to follow.  So that'd suggest having the `else if (disposition == IOUTBMESSAGE_ACCEPTED)` as part of "root else".

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -4006,32 +4025,57 @@ STRING_HANDLE IoTHubTransport_MQTT_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
+                    result = IOTHUB_CLIENT_OK;
+                }
+            }
         }
         else
         {
-            /*Codes_SRS_IOTHUB_MQTT_TRANSPORT_10_002: [If any of the messageData fields are NULL, IoTHubTransport_MQTT_Common_SendMessageDisposition shall fail and return IOTHUB_CLIENT_ERROR. ]*/
-            LogError("message handle is NULL");
-            result = IOTHUB_CLIENT_ERROR;
+            result = IOTHUB_CLIENT_OK;
```
jspaith, line 4072: **Comment 1 - important** - How does IoT Hub handle it when app does a IOTHUBMESSAGE_ABANDONED|IOTHUBMESSAGE_REJECTED?  If device never responds is it going to be in some sort of infinite loop of trying to resend the message?

Same question with IoT Edge, since it's "hub/MQTT" implementation is slightly different than Hub's so this may not be same answer.

I get MQTT doesn't have a way to signal this but I worry a lot about infinite loop case, unless hub/Edge is fine with no ACK?

**COMMENT2 - nit** Assuming we're OK just eating this message, I feel that a comment (no need to LogInfo()) that MQTT doesn't define a way to NACK really so all we can do is to just not send data.  A non-MQTT expert - which SDK tries to support by hiding all this stuff - who had to debug into code for whatever reason may not appreciate that nuance at 1st.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -2023,111 +2023,83 @@ static bool abandonOrAcceptMessage(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTR
...
+    {
+        free(dispositionContext->etagValue);
+        free(dispositionContext);
+    }
+}
+
+static MESSAGE_DISPOSITION_CONTEXT* CreateMessageDispositionContext(const char* etagValue)
 {
-    MESSAGE_CALLBACK_INFO* result = (MESSAGE_CALLBACK_INFO*)malloc(sizeof(MESSAGE_CALLBACK_INFO));
+    MESSAGE_DISPOSITION_CONTEXT* result = malloc(sizeof(MESSAGE_DISPOSITION_CONTEXT));
```
jspaith, line 2089: I recommend a `calloc` here.  I can think of two CVEs where uninited memory was root cause so even though this code looks file as-is for now, as defense in depth recommend this.

iothub_client/src/iothubtransporthttp.c
```diff
@@ -2262,57 +2234,59 @@ static void DoMessages(HTTPTRANSPORT_HANDLE_DATA* handleData, HTTPTRANSPORT_PERD
...
-                                                {
-                                                    (void)IoTHubTransportHttp_SendMessageDisposition(messageData, IOTHUBMESSAGE_ABANDONED);
-                                                }
                                             }
                                         }
-                                        IoTHubMessage_Destroy(receivedMessage);
+
+                                        if (abandon)
+                                        {
+                                            IoTHubMessage_Destroy(receivedMessage);
```
jspaith, line 2279: The same way you added MQTT comment comments about ownership, recommend something along lines of
```c
// If IoTHubMessage_SetDispositionContext succeeds above, it transitions ownership of dispositionContext to the receivedMessage.  IoTHubMessage_Destroy() below will handle freeing it.
```

jspaith (commented):
Review still in progress - I've made it through core/MQTT/HTTP - general looks good though please see my comments around how MQTT is supposed to handle REJECT|ABANDON to avoid Hub/Edge doing infinite loops.

Will look at AMQP side in a bit but want next round of comments out.

iothub_client/src/iothubtransport_amqp_common.c
```diff
@@ -2492,57 +2419,49 @@ STRING_HANDLE IoTHubTransport_AMQP_Common_GetHostname(TRANSPORT_LL_HANDLE handle
...
-
-                // Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_09_114: [`IoTHubTransport_AMQP_Common_SendMessageDisposition()` shall destroy the DEVICE_MESSAGE_DISPOSITION_INFO instance]
-                destroy_device_message_disposition_info(device_message_disposition_info);
+                result = IOTHUB_CLIENT_OK;
             }
         }
 
-        MESSAGE_CALLBACK_INFO_Destroy(message_data);
+        // Codes_SRS_IOTHUBTRANSPORT_AMQP_COMMON_09_114: [`IoTHubTransport_AMQP_Common_SendMessageDisposition()` shall destroy the DEVICE_MESSAGE_DISPOSITION_INFO instance]
+        IoTHubMessage_Destroy(message_handle);
```
jspaith, line 2464: If `handle==NULL` but `message_handle != NULL`, we will not execute this block (because of `handle==NULL`) but we need to because we have responsibility for freeing any non-NULL `message_handle`.

Because `IoTHubMessage_Destroy` will happily ignore NULL passed into it, I think we should unconditionally call `IoTHubMessage_Destroy` here in this function at the end so we're guaranteed to free(), especially if parameter checking becomes more complex in future.

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: ok i've reviewed this can work as-is.

jspaith (commented):
Put some hopefully final comments in.  This is a really big PR though so I need to re-review and sleep over one more night.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -43,13 +43,53 @@
 #include "certs.h"
 #endif // SET_TRUSTED_CERT_IN_SAMPLES
 
+// Uncomment this define to use Asynchronous ACK of cloud-to-device messages.
```
jspaith, line 46: I think we should give a bit more guidance to user so they're not necessarily guessing.

How about extra lines underneath the "Uncomment..."
// Most applications should leave USE_C2D_ASYNC_ACK undefined.  The default sample behavior is easier for the application to manage and meets the vast majority of application scenarios.
// The pattern sampled when USE_C2D_ASYNC_ACK is defined should be used when the application does 
// not want to immediately acknowledge receipt of the message to Azure IoT Hub during the callback.  It wants
// instead to defer the acknowledgement until a later time with a call to IoTHubDeviceClient_LL_SendMessageDisposition.

jspaith: Oh - looks like i said something similiar in the "I really think we should add some additional lines" below, too.  I actually tihnk my write up above is better since it's more to point and it starts with common case of discouraging apps from using this more advanced feature which is what we want.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -43,13 +43,53 @@
...
+
 /* Paste in the your iothub connection string  */
 static const char* connectionString = "[device connection string]";
 
 #define MESSAGE_COUNT        3
 static bool g_continueRunning = true;
 static size_t g_message_recv_count = 0;
 
+#ifdef USE_C2D_ASYNC_ACK
+static SINGLYLINKEDLIST_HANDLE g_cloudMessages;
```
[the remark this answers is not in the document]
jspaith: Recommend not using this now instead use the writeup i just added above.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -103,13 +143,21 @@ static IOTHUBMESSAGE_DISPOSITION_RESULT receive_msg_callback(IOTHUB_MESSAGE_HAND
     }
     g_message_recv_count++;
 
+#ifdef USE_C2D_ASYNC_ACK
+    (void)singlylinkedlist_add(g_cloudMessages, message);
+    return IOTHUBMESSAGE_ASYNC_ACK;
```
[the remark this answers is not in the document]
jspaith: I see these are still open but @ewertons please incorporate them (or re-wordsmith if needed), otherwise I think this will be tricky for folks to figure out.

jspaith (approved):
Added some more comments, especially around samples.

iothub_client/samples/iothub_client_sample_module_filter/iothub_client_sample_module_filter.c
```diff
@@ -70,6 +77,40 @@ static void PrintMessageInformation(IOTHUB_MESSAGE_HANDLE message)
...
+        messageId = "<unavailable>";
+    }
+
+    (void)printf("Sending ACK for cloud message (Message ID: %s)\r\n", messageId);
+
+    if (IoTHubModuleClient_LL_SendMessageDisposition(module_ll_handle, message, IOTHUBMESSAGE_ACCEPTED) != IOTHUB_CLIENT_OK)
+    {
+        (void)printf("ERROR: IoTHubModuleClient_LL_SendMessageDisposition..........FAILED!\r\n");
+        // If (and only if) IoTHubModuleClient_LL_SendMessageDisposition fails, the IOTHUB_MESSAGE_HANDLE must be explicitly destroyed.
+        IoTHubMessage_Destroy(message);
```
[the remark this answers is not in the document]
ewertons: The changes in this sample have been reverted.

iothub_client/samples/iothub_convenience_sample/iothub_convenience_sample.c
ewertons: Done.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
ewertons: This is a call like any other to the public API, if it fails it should print a message so the user knows something did not work.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Removed.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: I changed the code as per my last comment above. Thanks for initial poke, that brought my attention to that deficiency.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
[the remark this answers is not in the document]
ewertons: Done. Thanks for putting this together, John.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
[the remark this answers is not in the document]
ewertons: I just added that to the changes. Thanks John.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Thanks John.

iothub_client/src/iothub_client_core.c
```diff
@@ -2635,3 +2641,45 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_GenericMethodInvoke(IOTHUB_CLIENT_CORE_HAN
...
+            if (Lock(iotHubClientInstance->LockHandle) != LOCK_OK)
+            {
+                result = IOTHUB_CLIENT_ERROR;
+                LogError("Could not acquire lock");
+            }
+            else
+            {
+                result = IoTHubClientCore_LL_SendMessageDisposition(iotHubClientInstance->IoTHubClientLLHandle, message, disposition);
+
+                /* Codes_SRS_IOTHUBCLIENT_01_025: [IoTHubClient_SendEventAsync shall be made thread-safe by using the lock created in IoTHubClient_Create.] */
```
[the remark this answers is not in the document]
ewertons: Good catch. Thanks.

iothub_client/src/iothub_client_core.c
```diff
@@ -2635,3 +2641,45 @@ IOTHUB_CLIENT_RESULT IoTHubClientCore_GenericMethodInvoke(IOTHUB_CLIENT_CORE_HAN
...
+                (void)Unlock(iotHubClientInstance->LockHandle);
+
+                if (result != IOTHUB_CLIENT_OK)
+                {
+                    LogError("IoTHubClientCore_LL_SendMessageDisposition failed");
```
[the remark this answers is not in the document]
ewertons: I agree about the result. Now about moving it to inside the lock section, that's subject for talk.
With moving like you suggest we are writing it thinking more about the developers than the machine where the code will run, which is a lot more frequent. For that reason I will keep it outside the lock block. But let's discuss offline, I'm fine changing this later if we decide to do so.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Hmmm. Indeed. Thanks.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: That's a good idea.

ewertons: Done.

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: I agree 100%, except when there is a common code at the end of the block that would be common to the last else and a previous else-if. In that case this constructions keeps it cleaner. But we can chat about it later.
And indeed, the IOTHUBMESSAGE_ASYNC_ACK has been removed.

iothub_client/src/iothubtransporthttp.c
ewertons: That makes all the sense, thanks for that. Changed to calloc.

iothub_client/src/iothubtransporthttp.c
ewertons: Done.

iothub_client/src/iothubtransport_amqp_device.c
```diff
@@ -1821,3 +1821,37 @@ int amqp_device_get_twin_async(AMQP_DEVICE_HANDLE handle, DEVICE_TWIN_UPDATE_REC
 
     return result;
 }
+
+DEVICE_MESSAGE_DISPOSITION_INFO* amqp_device_clone_message_disposition_info(DEVICE_MESSAGE_DISPOSITION_INFO* disposition_info)
+{
+    DEVICE_MESSAGE_DISPOSITION_INFO* clone;
+
+    if ((clone = malloc(sizeof(DEVICE_MESSAGE_DISPOSITION_INFO))) == NULL)
```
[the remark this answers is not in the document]
ewertons: Done.

iothub_client/src/iothubtransport_amqp_common.c
ewertons: Indeed. That's all I can say. Thanks.

iothub_client/samples/iothub_client_sample_module_filter/iothub_client_sample_module_filter.c
```diff
@@ -110,6 +110,11 @@ static IOTHUBMESSAGE_DISPOSITION_RESULT DefaultMessageCallback(IOTHUB_MESSAGE_HA
 
 // SendConfirmationCallbackFromFilter is invoked when the message that was forwarded on from 'InputQueue1FilterCallback'
 // pipeline function is confirmed.
+// The Azure IoT C SDK allows developers to acknowledge cloud-to-device messages
```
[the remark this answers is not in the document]
ewertons: This sample has been reverted to remove these changes.

iothub_client/samples/iothub_convenience_sample/iothub_convenience_sample.c
```diff
@@ -73,6 +73,11 @@ static const char* proxy_password = NULL; // Proxy password
 const char* applicationCustomPropertyKey1 = "appCustomProperty1";
 const char* applicationCustomPropertyKey2 = "appCustomProperty2";
 
+// The Azure IoT C SDK allows developers to acknowledge cloud-to-device messages
```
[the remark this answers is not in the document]
ewertons: This sample has been reverted to remove all these changes. The only sample for async C2D disposition now is the iothub_ll_c2d_sample.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
ewertons: Alright, I'll use this one then. Thanks John.

iothub_client/inc/iothub_module_client_ll.h
```diff
@@ -385,6 +385,24 @@ extern "C"
 
 #endif /*USE_EDGE_MODULES*/
 
+    /**
+    * @brief    This API sends an acknowledgement to Azure IoT Hub that a cloud-to-device message has been received.
```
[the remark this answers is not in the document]
ewertons: Thanks John. Fixed.

iothub_client/src/iothub_module_client.c
```diff
@@ -129,3 +129,7 @@ IOTHUB_CLIENT_RESULT IoTHubModuleClient_ModuleMethodInvokeAsync(IOTHUB_MODULE_CL
 
 #endif /*USE_EDGE_MODULES*/
 
+IOTHUB_CLIENT_RESULT IoTHubModuleClient_SendMessageDisposition(IOTHUB_MODULE_CLIENT_HANDLE iotHubModuleHandle, IOTHUB_MESSAGE_HANDLE message, IOTHUBMESSAGE_DISPOSITION_RESULT disposition)
```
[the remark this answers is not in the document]
ewertons: I have no arguments against that. Done.

iothub_client/samples/iothub_ll_c2d_sample/iothub_ll_c2d_sample.c
```diff
@@ -43,13 +43,53 @@
...
+    {
+        (void)printf("ERROR: IoTHubDeviceClient_LL_SendMessageDisposition..........FAILED!\r\n");
+    }
+
+    *continue_processing = true;
+
+    return true;
+}
+
+static void acknowledge_cloud_messages(IOTHUB_DEVICE_CLIENT_LL_HANDLE device_ll_handle)
```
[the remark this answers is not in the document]
ewertons: Done. I added a more expanded explanation:

// `ack_and_remove_message` is a function that is executed by `singlylinkedlist_remove_if` for each list element.
// In this implementation it is used to send a delayed acknowledgement of receipt to Azure IoT Hub for each message.

and right above the call to `IoTHubDeviceClient_LL_SendMessageDisposition` this was added:

    // If using AMQP protocol, this function results in sending a MESSAGE DISPOSITION (ACCEPTED) for the given cloud-to-device message.
    // If using MQTT protocol, a PUBACK is sent for the cloud-to-device message (only) if `IOTHUBMESSAGE_ACCEPTED` is used.
    // If using HTTP protocol no delayed acknowledgement is sent for the cloud-to-device message, as this protocol does not support that.
    // Independent of the protocol used, this function MUST be called by the user application if using delayed acknowledgement of 
    // cloud-to-device messages, as it will free the memory allocated for each of those messages received.

iothub_client/inc/iothub_device_client_ll.h
ewertons: I ended up adding the following documentation:

```c
    * @remarks
    *           If your cloud-to-device message callback returned IOTHUBMESSAGE_ASYNC_ACK, it MUST call this API eventually.
    *           Beyond sending acknowledgment to the service, this method also handles freeing message's memory.
    *           Not calling this function will result in memory leaks.
    *           Depending on the protocol used, this API will acknowledge cloud-to-device messages differently:
    *           AMQP: A MESSAGE DISPOSITION is sent using the `disposition` option provided.
    *           MQTT: A PUBACK is sent if `disposition` is `IOTHUBMESSAGE_ACCEPTED`. Passing any other option results in no PUBACK sent for the message.
    *           HTTP: A HTTP request is sent using the `disposition` option provided.
```

iothub_client/inc/iothub_device_client_ll.h
```diff
@@ -374,6 +374,24 @@ typedef struct IOTHUB_CLIENT_CORE_LL_HANDLE_DATA_TAG* IOTHUB_DEVICE_CLIENT_LL_HA
 
 #endif /*DONT_USE_UPLOADTOBLOB*/
 
+    /**
+    * @brief    This API sends an acknowledgement to Azure IoT Hub that a cloud-to-device message has been received.
+    *
+    * @param    device_ll_handle                The handle created by a call to a create function.
+    * @param    message                         The cloud-to-device message received through the callback provided to IoTHubDeviceClient_LL_SetMessageCallback.
+    * @param    disposition                     Acknowledgement option for the message.
```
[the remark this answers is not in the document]
ewertons: I ended up adding the following documentation:

```c
    * @remarks
    *           If your cloud-to-device message callback returned IOTHUBMESSAGE_ASYNC_ACK, it MUST call this API eventually.
    *           Beyond sending acknowledgment to the service, this method also handles freeing message's memory.
    *           Not calling this function will result in memory leaks.
    *           Depending on the protocol used, this API will acknowledge cloud-to-device messages differently:
    *           AMQP: A MESSAGE DISPOSITION is sent using the `disposition` option provided.
    *           MQTT: A PUBACK is sent if `disposition` is `IOTHUBMESSAGE_ACCEPTED`. Passing any other option results in no PUBACK sent for the message.
    *           HTTP: A HTTP request is sent using the `disposition` option provided.
```

iothub_client/src/iothubtransport_mqtt_common.c
ewertons: Comment 1: As per design, an ABANDONED message is requeued to be delivered to the device again when it reconnects. A REJECTED message is simply dequeued by the hub and never re-delivered.
If the customer does ABANDON a message, it is expected (no matter if using IoT Hub or Edge) that the message will be redelivered upon reconnection. So it will be up to the customer to control that.

Note: Currently for MQTT we do not send a PUBACK for a rejected message. I now believe that we should send  a PUBACK, to simulate the scenario of rejecting the message (so it does not get redelivered later).

Comment 2: Sure, that makes sense. Suggested comment:

```c
// Azure IoT Hub (and Edge) support the following responses for devicebound messages:
// ACCEPTED: message shall be dequeued by the hub and consider the delivery completed.
// ABANDONED: message shall remain in the hub queue, and re-delivered upon client re-connection.
// REJECTED: message shall be dequeued by the hub and not not re-delivered anymore.
// The MQTT protocol by design only supports acknowledging the receipt of a message (through PUBACK).
// To simulate the message responses above, this function behaves as follows:
// ACCEPTED: a PUBACK is sent to the Hub/Edge.
// ABANDONED: no response is sent to the Hub/Edge.
// REJECTED: a PUBACK is sent to the Hub/Edge (same as for ACCEPTED).
```

Issue #1995, opened 2021-06-15
Fail to upload long Blob using mbedtls+compact


maallahatem:
When the sdk is compiled using **httpAPI_compact** and **mbedtls**, often the blob upload by chunk fails starting from 512Kb (chunk size).
The following log describes the HTTP failure during the operation

````
[DEBUG] -> 14:54:23 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: edge-ih-dev.azure-devices.net/myUserName/?api-version=2020-09-30&DeviceClientType=iothubclient%2f1.6.0%20(native%3b%20Linux%3b%20x86_64) | PWD: XXXX | CLEAN: 0
[DEBUG] <- 14:54:23 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
[DEBUG] -> 14:54:23 SUBSCRIBE | PACKET_ID: 2 | TOPIC_NAME: devices/myUserName/messages/devicebound/# | QOS: 1 | TOPIC_NAME: $iothub/twin/res/# | QOS: 0 | TOPIC_NAME: $iothub/methods/POST/# | QOS: 0
[DEBUG] <- 14:54:23 SUBACK | PACKET_ID: 2 | RETURN_CODE: 1 | RETURN_CODE: 0 | RETURN_CODE: 0
[DEBUG] -> 14:54:23 PUBLISH | IS_DUP: false | RETAIN: 0 | QOS: DELIVER_AT_MOST_ONCE | TOPIC_NAME: $iothub/twin/GET/?$rid=3
[DEBUG] <- 14:54:23 PUBLISH | IS_DUP: false | RETAIN: 0 | QOS: QOS_VALUE_INVALID | TOPIC_NAME: $iothub/twin/res/200/?$rid=3 | PAYLOAD_LEN: 1248
[DEBUG] -> 14:54:23 SUBSCRIBE | PACKET_ID: 4 | TOPIC_NAME: $iothub/twin/PATCH/properties/desired/# | QOS: 0
[DEBUG] -> 14:54:23 PUBLISH | IS_DUP: false | RETAIN: 0 | QOS: DELIVER_AT_MOST_ONCE | TOPIC_NAME: $iothub/twin/PATCH/properties/reported/?$rid=5 | PAYLOAD_LEN: 64
[DEBUG] <- 14:54:24 SUBACK | PACKET_ID: 4 | RETURN_CODE: 0
[...]

Issue #536, opened 2018-06-28
Problem building a client app on Windows


dominicbetts:
I'm using Windows 10 and Visual Studio 2017.

I'm stepping through this tutorial: https://docs.microsoft.com/en-us/azure/iot-accelerators/iot-accelerators-connecting-devices
I've added the following Nuget packages to the project:

- Microsoft.Azure.IoTHub.Serializer
- Microsoft.Azure.IoTHub.IoTHubClient
- Microsoft.Azure.IoTHub.MqttTransport

I have the following include statements:

```c
#include "iothubtransportmqtt.h"
#include "schemalib.h"
#include "iothub_client.h"
#include "serializer_devicetwin.h"
#include "schemaserializer.h"
#include "azure_c_shared_utility/threadapi.h"
#include "azure_c_shared_utility/platform.h"
#include <string.h>
```

With version 1.2.4 or earlier of the **Microsoft.Azure.IoTHub.IoTHubClient** package, the project builds. Any later version fails to build with:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error (active)	E1696	cannot open source file "iothub_client_core_ll.h"	RMDevice	c:\temp\RMDevice\packages\Microsoft.Azure.IoTHub.IoTHubClient.1.2.5\build\native\include\iothub_client.h	24	
Error (active)	E1696	cannot open source file "iothub_client_core.h"	RMDevice	c:\temp\RMDevice\packages\Microsoft.Azure.IoTHub.IoTHubClient.1.2.5\build\native\include\iothub_client.h	25	

...

Error	C1083	Cannot open include file: 'iothub_client_core_ll.h': No such file or directory	RMDevice	c:\temp\rmdevice\packages\microsoft.azure.iothub.iothubclient.1.2.5\build\native\include\iothub_client.h	24	
```

ewertons:
Hi @dominicbetts ,
apologize for the issue, it indeed seems to be a bug either in the nuget (less likely) or in the documentation. We will address that briefly.

yzhong94:
@ewertons any update on this issue?  Is it an issue with the sample/bug?

ewertons:
We are releasing a new set of nugets today, then I'll check the state of the failure.

ewertons:
Hi @dominicbetts ,
could you try again, using the new nuget packages released today (1.2.8)?

For reference:
[...]

PR #2004, opened 2021-06-18
Fail to upload long Blob using mbedtls+compact(Issue #1995)

maallahatem:
Hello,
The aim of this pull request is to complete the PR Fail to upload long Blob using mbedtls+compact(Issue #1995) #536

Regards
Hatem


ericwol-msft:
> /azp run

@danewalton We need to pull in changes from cutil first

danewalton:
> > /azp run
> 
> @danewalton We need to pull in changes from cutil first

Ah I see I didn't realize it was two different PR's I thought I was looking at the same one.

PR #2006, opened 2021-06-23
Add simple thermostat sample for new PnP API

jspaith:
Adds simple thermostat.

Note that there was pre-existing code using the "sample" API before which is being changed in place.  It's also being substantially cleaned up.  Some is because it's needed to fix up, and also bringing this implementation more inline with the thermostat component as part of more complex temp sensor.

For the ./common directory, we either:
* Deprecated stuff we really don't want to use anymore, like the pnp_protocol, is clearly deprecated both in files and in readme.md.
* Add a few more files for helpers.
* Update the DPS implementation, which should remain common, but needed cleaning up.

Checkin flow:
* Stage1 Checked in the public header files.
* Stage2a **This checkin** adds the updated samples for simple thermostat.
* Stage2b will add the updated samples for more complex temperature sensor.
* Stage3 will add code implementation.
* Stage4 will add unit tests.

### DIFF
--- a/iothub_client/samples/pnp/common/pnp_sample_config.c
+++ b/iothub_client/samples/pnp/common/pnp_sample_config.c
@@ -1,0 +1,132 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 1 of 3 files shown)


jspaith:
Closing this was to deal with git issues.  #2007 is correct fix courtesy of @danewalton (thx!)

PR #2033, opened 2021-07-27
Have multiple PnP component sample use official PnP API

jspaith:
*Bulk of this change.*

* Rework PnP sample to stop using the "helper sample" and instead use official API.  
  
  I tried to keep the multi-component sample as close as feasible to the single component sample, which we've reviewed a bunch already.  This is for consistency for customers mostly and also to leverage all the PR muscle put into the initial review.

*Additionally...*

* `pnp_simple_thermostat.c` fixes, both bug fixes (like strtod and not strtol) and how DPS portions are optionally included in CMake.

* Fix `pnp_dps_ll.c`.  I had previously contemplated some sort of helper API that would generate the `{"modelId":"<modelId>"}` JSON blob.  We could add later if we wanted, but... this JSON is pretty straightforward and I think vast majority of time should just be a `static const char` (because ModelId doesn't change much).  So I removed the references to this JSON builder API (which was never implemented fortunately) and just showed a more direct C string build up.

### DIFF
--- a/iothub_client/samples/pnp/common/pnp_dps_ll.c
+++ b/iothub_client/samples/pnp/common/pnp_dps_ll.c
@@ -53,7 +53,7 @@ static char* g_dpsDeviceId;
 // maximum number of characters for an ID is 128.  We reserve extra space for the JSON envelope.
 #define PNP_DPS_REGISTRATION_MAX_PAYLOAD_SIZE (128 + 16)
 // snprintf style format for building the payload for Prov_Device_LL_Set_Provisioning_Payload() call.
-static const char PNP_DPS_REGISTRATION_FMT[] =  "{\"modelId\":\"%s\"}";
+static const char PNP_DPS_REGISTRATION_FMT[] = "{\"modelId\":\"%s\"}";
 
 //
 // provisioningRegisterCallback is called by the DPS client when the DPS server has either succeeded or failed the DPS
@@ -98,12 +98,7 @@ IOTHUB_DEVICE_CLIENT_LL_HANDLE PnP_CreateDeviceClientLLHandle_ViaDps(const PNP_D
     g_pnpDpsRegistrationStatus = PNP_DPS_REGISTRATION_NOT_COMPLETE;
 
     // Initial DPS setup and handle creation.
-    if (snprintf(modelIdPayload, sizeof(modelIdPayload), PNP_DPS_REGISTRATION_FMT, pnpDeviceConfiguration->modelId) < 0)
-    {
-        LogError("building modelId payload unsuccessful");
-        result = false;
-    }
-    else if ((prov_dev_set_symmetric_key_info(pnpDeviceConfiguration->u.dpsConnectionAuth.deviceId, pnpDeviceConfiguration->u.dpsConnectionAuth.deviceKey) != 0))
+    if ((prov_dev_set_symmetric_key_info(pnpDeviceConfiguration->u.dpsConnectionAuth.deviceId, pnpDeviceConfiguration->u.dpsConnectionAuth.deviceKey) != 0))
     {
         LogError("prov_dev_set_symmetric_key_info failed.");
         result = false;
@@ -121,6 +116,11 @@ IOTHUB_DEVICE_CLIENT_LL_HANDLE PnP_CreateDeviceClientLLHandle_ViaDps(const PNP_D
     else if ((provDeviceResult = Prov_Device_LL_SetOption(provDeviceClient, PROV_OPTION_LOG_TRACE, &pnpDeviceConfiguration->enableTracing)) != PROV_DEVICE_RESULT_OK)
     {
         LogError("Setting provisioning tracing on failed, error=%d", provDeviceResult);
+        result = false;
+    }
+    else if (snprintf(modelIdPayload, sizeof(modelIdPayload), PNP_DPS_REGISTRATION_FMT, pnpDeviceConfiguration->modelId) < 0)
+    {
+        LogError("building modelId payload unsuccessful");
         result = false;
     }
     else if ((provDeviceResult = Prov_Device_LL_Set_Provisioning_Payload(provDeviceClient, modelIdPayload)) != PROV_DEVICE_RESULT_OK)
--- a/iothub_client/samples/pnp/common/pnp_dps_ll.c
+++ b/iothub_client/samples/pnp/common/pnp_dps_ll.c
@@ -17,7 +17,6 @@
 #include "azure_prov_client/prov_device_client.h"
 #include "azure_prov_client/prov_transport_mqtt_client.h"
 #include "azure_prov_client/prov_security_factory.h"
-#include "azure_prov_client/prov_client_model.h"
 #include "iothub_device_client_ll.h"
 
 // DPS and IoT core utility header files.
@@ -48,6 +47,13 @@ static const int g_dpsRegistrationPollSleep = 1000;
 static char* g_dpsIothubUri;
 // DeviceId for this device as determined by the DPS client runtime.
 static char* g_dpsDeviceId;
+
+// Maximum size ModelId DPS payload can be in this sample.  This comes from
+// the DTDLv2 limits (see https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#interface), where the 
+// maximum number of characters for an ID is 128.  We reserve extra space for the JSON envelope.
+#define PNP_DPS_REGISTRATION_MAX_PAYLOAD_SIZE (128 + 16)
+// snprintf style format for building the payload for Prov_Device_LL_Set_Provisioning_Payload() call.
+static const char PNP_DPS_REGISTRATION_FMT[] =  "{\"modelId\":\"%s\"}";
 
 //
 // provisioningRegisterCallback is called by the DPS client when the DPS server has either succeeded or failed the DPS
@@ -86,13 +92,18 @@ IOTHUB_DEVICE_CLIENT_LL_HANDLE PnP_CreateDeviceClientLLHandle_ViaDps(const PNP_D
 
     PROV_DEVICE_RESULT provDeviceResult;
     PROV_DEVICE_LL_HANDLE provDeviceClient = NULL;
-    char* modelIdPayload = NULL;
+    char modelIdPayload[PNP_DPS_REGISTRATION_MAX_PAYLOAD_SIZE];
 
     LogInfo("Initiating DPS client to retrieve IoT Hub connection information");
     g_pnpDpsRegistrationStatus = PNP_DPS_REGISTRATION_NOT_COMPLETE;
 
     // Initial DPS setup and handle creation.
-    if ((prov_dev_set_symmetric_key_info(pnpDeviceConfiguration->u.dpsConnectionAuth.deviceId, pnpDeviceConfiguration->u.dpsConnectionAuth.deviceKey) != 0))
+    if (snprintf(modelIdPayload, sizeof(modelIdPayload), PNP_DPS_REGISTRATION_FMT, pnpDeviceConfiguration->modelId) < 0)
+    {
+        LogError("building modelId payload unsuccessful");
+        result = false;
+    }
+    else if ((prov_dev_set_symmetric_key_info(pnpDeviceConfiguration->u.dpsConnectionAuth.deviceId, pnpDeviceConfiguration->u.dpsConnectionAuth.deviceKey) != 0))
     {
         LogError("prov_dev_set_symmetric_key_info failed.");
         result = false;
@@ -110,14 +121,6 @@ IOTHUB_DEVICE_CLIENT_LL_HANDLE PnP_CreateDeviceClientLLHandle_ViaDps(const PNP_D
     else if ((provDeviceResult = Prov_Device_LL_SetOption(provDeviceClient, PROV_OPTION_LOG_TRACE, &pnpDeviceConfiguration->enableTracing)) != PROV_DEVICE_RESULT_OK)
     {
         LogError("Setting provisioning tracing on failed, error=%d", provDeviceResult);
-        result = false;
-    }
-    // The next steps indicate the ModelId of the device to DPS.  This allows the service to perform custom operations,
-    // such as allocating a different IoT Hub to devices based on their ModelId.  
-    // When connecting to IoT Central with DPS, this step is required.
-    else if ((provDeviceResult = Prov_Client_Create_ModelIdPayload(pnpDeviceConfiguration->modelId, &modelIdPayload)) != PROV_DEVICE_RESULT_OK)
-    {
-        LogError("Allocating custom DPS payload for modelId failed, error=%d.", provDeviceResult);
         result = false;
     }
     else if ((provDeviceResult = Prov_Device_LL_Set_Provisioning_Payload(provDeviceClient, modelIdPayload)) != PROV_DEVICE_RESULT_OK)
@@ -186,7 +189,6 @@ IOTHUB_DEVICE_CLIENT_LL_HANDLE PnP_CreateDeviceClientLLHandle_ViaDps(const PNP_D
 
     free(g_dpsIothubUri);
     free(g_dpsDeviceId);
-    Prov_Client_Free_ModelIdPayload(modelIdPayload);
 
     return deviceClient;
 }
--- a/iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
+++ b/iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
@@ -67,7 +67,7 @@ static const char g_targetTemperaturePropertyName[] = "targetTemperature";
 static const char g_maxTempSinceLastRebootPropertyName[] = "maxTempSinceLastReboot";
 
 // Name of command this component supports to get report information
-static const char g_getMaxMinReport[] = "getMaxMinReport";
+static const char g_getMaxMinReportCommandName[] = "getMaxMinReport";
 
 // An empty JSON body for PnP command responses.
 static const char g_JSONEmpty[] = "{}";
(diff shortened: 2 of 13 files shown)


iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -6,47 +6,66 @@
...
+// Maximum component size.  The spec (https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#component)
+// defines the maximum component length as 64 to which we need to reserve the null-terminator.
+#define MAX_COMPONENT_NAME_LENGTH (64 + 1)
+
+// Names of properties for desired/reporting.
 static const char g_targetTemperaturePropertyName[] = "targetTemperature";
 static const char g_maxTempSinceLastRebootPropertyName[] = "maxTempSinceLastReboot";
 
+// Name of command this component supports to retrieve a report about the component.
+static const char g_getMaxMinReport[] = "getMaxMinReport";
```
ewertons, line 38: Just as a minor comment, but there are other variables below that are named with suffix "BodyFormat",  "ResponseFormat", etc. That's good, it helps make the code clearer.
Suggestion for variables like this: Add a suffix "CmdName". E.g.: `g_getMaxMinReportCmdName`.

iothub_client/samples/pnp/common/pnp_dps_ll.c
```diff
@@ -49,6 +48,13 @@ static char* g_dpsIothubUri;
 // DeviceId for this device as determined by the DPS client runtime.
 static char* g_dpsDeviceId;
 
+// Maximum size ModelId DPS payload can be in this sample.  This comes from
+// the DTDLv2 limits (see https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#interface), where the 
+// maximum number of characters for an ID is 128.  We reserve extra space for the JSON envelope.
+#define PNP_DPS_REGISTRATION_MAX_PAYLOAD_SIZE (128 + 16)
+// snprintf style format for building the payload for Prov_Device_LL_Set_Provisioning_Payload() call.
+static const char PNP_DPS_REGISTRATION_FMT[] =  "{\"modelId\":\"%s\"}";
```
hihigupt, line 56: >  [](http://example.com/codeflow?start=46&length=1)

nit : extra space

iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c
```diff
@@ -12,16 +12,17 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <limits.h>
```
hihigupt, line 15: >limits [](http://example.com/codeflow?start=10&length=6)

only api that is changed in this file is strtod, which comes from #include <stdlib.h> any reason this lib is added? Is this added for some warning?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_deviceinfo_component.c
```diff
@@ -40,47 +39,34 @@ static const char PnPDeviceInfo_TotalStoragePropertyValue[] = "10000";
...
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_ManufacturerPropertyName, PnPDeviceInfo_ManufacturerPropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_ModelPropertyName, PnPDeviceInfo_ModelPropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_OsNamePropertyName, PnPDeviceInfo_OsNamePropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_ProcessorArchitecturePropertyName, PnPDeviceInfo_ProcessorArchitecturePropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_ProcessorManufacturerPropertyName, PnPDeviceInfo_ProcessorManufacturerPropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_TotalStoragePropertyName, PnPDeviceInfo_TotalStoragePropertyValue},
+        {IOTHUB_CLIENT_REPORTED_PROPERTY_STRUCT_VERSION_1, PnPDeviceInfo_TotalMemoryPropertyName, PnPDeviceInfo_TotalMemoryPropertyValue}
+    };
+
+    const int numProperties = sizeof(properties) / sizeof(properties[0]);
```
hihigupt, line 58: >const int [](http://example.com/codeflow?start=4&length=9)

const size_t?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -5,57 +5,48 @@
...
-// Environment variable used to optionally specify this application's DPS id scope
-static const char g_dpsEndpointEnvironmentVariable[] = "IOTHUB_DEVICE_DPS_ENDPOINT";
+// Headers that provide implementation for subcomponents (the two thermostat components and DeviceInfo)
+#include "pnp_thermostat_component.h"
+#include "pnp_deviceinfo_component.h"
 
-// Global provisioning endpoint for DPS if one is not specified via the environment
-static const char g_dps_DefaultGlobalProvUri[] = "global.azure-devices-provisioning.net";
-#endif
+#define CURRENT_WORKING_SET_BUFFER_SIZE 64
```
hihigupt, line 49: >CURRENT_WORKING_SET_BUFFER_SIZE [](http://example.com/codeflow?start=8&length=31)

how about making these const int or converting all global const like on line 56 to macro?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -148,65 +144,87 @@ static void SetEmptyCommandResponse(unsigned char** response, size_t* responseSi
...
+    
+    return result;
+}
+
+//
+// PnP_TempControlComponent_CommandCallback is invoked by IoT SDK when a command arrives.
 //
-static int PnP_TempControlComponent_DeviceMethodCallback(const char* methodName, const unsigned char* payload, size_t size, unsigned char** response, size_t* responseSize, void* userContextCallback)
+static int PnP_TempControlComponent_CommandCallback(const char* componentName, const char* commandName, const unsigned char* payload, size_t size, const char* payloadContentType, unsigned char** response, size_t* responseSize, void* userContextCallback)
 {
```
hihigupt, line 173: nit: need line break

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -470,24 +521,34 @@ static IOTHUB_DEVICE_CLIENT_LL_HANDLE CreateDeviceClientAndAllocateComponents(vo
...
+    return result;
 }
 
 int main(void)
 {
     IOTHUB_DEVICE_CLIENT_LL_HANDLE deviceClient = NULL;
 
-    if ((deviceClient = CreateDeviceClientAndAllocateComponents()) == NULL)
+    g_pnpDeviceConfiguration.modelId = g_temperatureControllerModelId;
+    g_pnpDeviceConfiguration.enableTracing = g_hubClientTraceEnabled;
```
hihigupt, line 536: should this come from Environment setting?

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -470,24 +521,34 @@ static IOTHUB_DEVICE_CLIENT_LL_HANDLE CreateDeviceClientAndAllocateComponents(vo
...
+    else if (AllocateThermostatComponents() == false)
+    {
+        LogError("Failure allocating thermostat components");
+    }
+    // Create a handle to device client handle.  Note that this call may block
+    // for extended periods of time when using DPS.
+    else if ((deviceClient = CreateAndConfigureDeviceClientHandleForPnP()) == NULL)
     {
         LogError("Failure creating IotHub device client");
     }
```
hihigupt, line 554: PnP_ThermostatComponent_Destroy(g_thermostatHandle1);
        PnP_ThermostatComponent_Destroy(g_thermostatHandle2);

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -276,184 +327,184 @@ static void PnP_TempControlComponent_DeviceTwinCallback(DEVICE_TWIN_UPDATE_STATE
...
+        LogError("Unable to send reported state, error=%d", clientResult);
     }
 
-    STRING_delete(jsonToSend);
+    IoTHubClient_Serialize_Properties_Destroy(serializedProperties);
 }
 
 //
-// GetConnectionStringFromEnvironment retrieves the connection string based on environment variable
+// CreateDeviceClientLLHandle does the creates the IOTHUB_DEVICE_CLIENT_LL_HANDLE based on environment configuration.
```
danewalton, line 389:
```suggestion
// CreateDeviceClientLLHandle creates the IOTHUB_DEVICE_CLIENT_LL_HANDLE based on environment configuration.
```

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -276,184 +327,184 @@ static void PnP_TempControlComponent_DeviceTwinCallback(DEVICE_TWIN_UPDATE_STATE
...
+    IOTHUB_DEVICE_CLIENT_LL_HANDLE deviceClient = NULL;
 
-    if ((g_pnpDeviceConfiguration.u.connectionString = getenv(g_connectionStringEnvironmentVariable)) == NULL)
+    if (g_pnpDeviceConfiguration.securityType == PNP_CONNECTION_SECURITY_TYPE_CONNECTION_STRING)
     {
-        LogError("Cannot read environment variable=%s", g_connectionStringEnvironmentVariable);
-        result = false;
+        if ((deviceClient = IoTHubDeviceClient_LL_CreateFromConnectionString(g_pnpDeviceConfiguration.u.connectionString, MQTT_Protocol)) == NULL)
+        {
+            LogError("Failure creating IotHub client.  Hint: Check your connection string");
```
danewalton, line 400:
```suggestion
            LogError("Failure creating Iot Hub client.  Hint: Check your connection string");
```

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -510,22 +571,26 @@ int main(void)
...
+                PnP_ThermostatComponent_SendCurrentTemperature(g_thermostatHandle2, deviceClient);
             }
 
             IoTHubDeviceClient_LL_DoWork(deviceClient);
             ThreadAPI_Sleep(g_sleepBetweenPollsMs);
             numberOfIterations++;
         }
 
+        // The remainder of the code is used for cleaning up our allocated resources. It won't be executed in this 
+        // sample (because the loop above is infinite and // is only broken out of by Control-C of the program), but
```
danewalton, line 584:
```suggestion
        // sample (because the loop above is infinite and is only broken out of by Control-C of the program), but 
```

danewalton (approved):

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Yeah I generally prefer const over preprocessor too.  But here -- 

The difference is `g_sleepBetweenPollsMs` on 56 is just used as a parameter in a function call.  But the `CURRENT_WORKING_SET_BUFFER_SIZE` is used when initializing the max size of a static array.

I haven't researched, but I'm almost positive I've seen at least some C99 compilers get very unhappy if an array size is set with a variable - even a const - so it demands the pre-processor come in and stitch it up 1st.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -78,54 +73,55 @@ PNP_THERMOSTAT_COMPONENT_HANDLE g_thermostatHandle2;
...
+// Values of connection / security settings read from environment variables and/or DPS runtime.
+PNP_DEVICE_CONFIGURATION g_pnpDeviceConfiguration;
+
 //
-// PnP_TempControlComponent_InvokeRebootCommand processes the reboot command on the root interface
+// CopyPayloadToString creates a null-terminated string from the payload buffer.
+// payload is not guaranteed to be null-terminated by the IoT Hub device SDK.
 //
-static int PnP_TempControlComponent_InvokeRebootCommand(JSON_Value* rootValue)
+static char* CopyPayloadToString(const unsigned char* payload, size_t size)
```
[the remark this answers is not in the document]
jspaith: This is interesting.  I think we should discuss this one further offline as there's some nuance here.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: @hihigupt  - I'm assuming what you mean here is that this line runs on WAY too long and that the params should be split between lines?  I agree and am fixing.

If you mean something else - like CRLF | LF or whatever type line breaks are screwed up - please let me know.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Fixing here and the other places in logs/comments that this should've been separate.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
```diff
@@ -276,184 +327,184 @@ static void PnP_TempControlComponent_DeviceTwinCallback(DEVICE_TWIN_UPDATE_STATE
...
+        result = true;
+    }
+
+    if (result == false)
+    {
+        IoTHubDeviceClient_LL_Destroy(deviceClient);
+        deviceClient = NULL;
+    }
+
+    if ((result == false) &&  (iothubInitResult == 0))
```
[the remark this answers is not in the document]
jspaith: agree your proposal is cleaner.  Fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: We could, but I *think* it's OK as-is.  The other stuff like conn strings/DPS scopeIds we want as environment vars because we don't want people hard-coding secrets in .c files and this makes experimentation a bit easier.  Generally when you (the developer) compile a sample EXE though you have a good sense of whether you want it with full debug logs blasting (which is the default since we're not that verbose) or not.

Also complicating this is we'd need to coordinate potentially with all the non-embedded SDKs, which follow this pattern too.

So my inclination is to won't-fix this, though if we make future SDK work that's totaly fresh we should reconsider this and anything else that could be from env.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Good catch.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Fixing.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -71,7 +90,7 @@ typedef struct PNP_THERMOSTAT_COMPONENT_TAG
 PNP_THERMOSTAT_COMPONENT;
 
 //
-// BuildUtcTimeFromCurrentTime writes the current time, in ISO 8601 format, into the specified buffer
+// BuildUtcTimeFromCurrentTime writes the current time, in ISO 8601 format, into the specified buffer.
 //
 static bool BuildUtcTimeFromCurrentTime(char* utcTimeBuffer, size_t utcTimeBufferSize)
```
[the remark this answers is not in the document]
jspaith: Yes, potentially.  We should chat on this, too.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
```diff
@@ -95,14 +114,13 @@ static bool BuildUtcTimeFromCurrentTime(char* utcTimeBuffer, size_t utcTimeBuffe
     return result;
 }
 
-
 PNP_THERMOSTAT_COMPONENT_HANDLE PnP_ThermostatComponent_CreateHandle(const char* componentName)
 {
     PNP_THERMOSTAT_COMPONENT* thermostatComponent;
 
-    if (strlen(componentName) > PNP_MAXIMUM_COMPONENT_LENGTH)
+    if (strlen(componentName) > 64)
```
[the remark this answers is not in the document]
jspaith: Oh my.  Just when I think there's progress in the world I do this.  

Changing this to use `MAX_COMPONENT_NAME_LENGTH`, which was in file.  And cleaning up `MAX_COMPONENT_NAME_LENGTH` so it doesn't pre-include the \0 terminator but lets place that defines buffer do this.

iothub_client/samples/pnp/pnp_temperature_controller/pnp_thermostat_component.c
jspaith: Yup, like this too.  Fixing.

hihigupt (approved):

iothub_client/samples/pnp/pnp_temperature_controller/pnp_temperature_controller.c
jspaith: Pretty sure its the too long and needing an extra line.  I've fixed this as suggested, so resolving.

PR #2041, opened 2021-08-11
Final implementation for PnP public-preview.

jspaith:
Omnibus change that contains a number of smallish changes.  (The vast # of lines here are in test code and cleaning up spacing, which makes size of change misleading at 1st look.)  Should wrap up feature (and test and sample work) on PnP heading into officially cutting a preview for it.

* Implements subscribeToCommand for the convenience layer.
* No longer require the componentList to be passed in when parsing properties.  After a lot
  of discussion we're OK taking a dependency on "__t:c" being sent even in a PATCH update so
  the SDK can index off that.
* Add wrappers for device|module clients.
* Move to MOCKABLE framework for intra header consistency.
* Move the original sampler helper code to a ./deprecated directory so new customers won't use but existing ones don't have to dig through git history as they transition.
* E2E tests for subscribeToCommand at convenience layer.


jspaith:
I overcomplicated an already complicated merge on this one.  Taking down the PR and the branch and will get an updated, clean-merge out later today.

PR #2042, opened 2021-08-11
Bump version to 1.8.0 and update LTS documentation for release.

momuno:
Version on main was 1.7.0, now 1.8.0.  Five files missed the last update so you will see 1.6.0 -> 1.8.0.
LTS documentation updated for LTS release posting today.


provisioning_client/inc/azure_prov_client/prov_client_const.h
```diff
@@ -4,7 +4,7 @@
 #ifndef PROV_CLIENT_CONST_H
 #define PROV_CLIENT_CONST_H
 
-#define PROV_DEVICE_CLIENT_VERSION "1.6.0"
```
ericwol-msft, line 7: how id the ut pass with 1.7.0??

ericwol-msft (approved):

provisioning_client/inc/azure_prov_client/prov_client_const.h
momuno: 1.7.0 was only updated in version.txt.  I did not know about the other 5 files to update back in January for LTS_01_2021.  My best guess is the ut checks agains the version value in iothub_client_verision.h.  Those were the same, so the test passed.

I really don't like having 6 files that independently control the version.

PR #2078, opened 2021-09-09
send IOTHUB_CLIENT_CONNECTION_COMMUNICATION_ERROR of failed PUBACK DC messages

ericwol-msft:
# Checklist

# Description of the problem

Fix for issue #1255 was partially correct. There was a missing connection status callback in the affected code, however the status to be raised should be COMMUNICATION_ERROR, not RETRY_EXPIRED. RETRY_EXPIRED must only be raised on the code path that actually deals with retry policy checks.

# Description of the solution
Replace the connection status callback value of RETRY_EXPIRED by COMMUNICATION_ERROR.


iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2418,11 +2418,7 @@ static void ProcessPendingTelemetryMessages(PMQTTTRANSPORT_HANDLE_DATA transport
                 free(msg_detail_entry);
 
                 DisconnectFromClient(transport_data);
-                if (!transport_data->isRetryExpiredCallbackCalled) // Only call once
-                {
-                    transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED, transport_data->transport_ctx);
-                    transport_data->isRetryExpiredCallbackCalled = true;
-                }
+                transport_data->transport_callbacks.connection_status_cb(IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED, IOTHUB_CLIENT_CONNECTION_COMMUNICATION_ERROR, transport_data->transport_ctx);
```
danewalton, line 2421: just double checking here.....we no longer have the `if()` condition on calling it once since this is for a different reason?

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -2769,7 +2765,7 @@ static int UpdateMqttConnectionStateIfNeeded(PMQTTTRANSPORT_HANDLE_DATA transpor
             }
             else if (retry_action == RETRY_ACTION_STOP_RETRYING)
             {
-                // Set callback if retry expired
+                // send callback if retry expired
                 if (!transport_data->isRetryExpiredCallbackCalled)
```
danewalton, line 2769: if the above is true does this need to be protected anymore? is this only called in one place?

danewalton (approved):

iothub_client/src/iothubtransport_mqtt_common.c
ericwol-msft: correct. the problem was that IOTHUB_CLIENT_CONNECTION_RETRY_EXPIRED was also fired in another place, so two events were fired with the same value.

iothub_client/src/iothubtransport_mqtt_common.c
ericwol-msft: Yes it does. When you queue a new message this event will keep firing for every new message. I initially removed it, but broke the unit test. I added it back. This was previous code, not related to the new event that was added above.

ewertons, line 2769: I think we talked about removing the isRetryExpiredCallbackCalled entirely. Was there a reason to keep it?

ericwol-msft: failed a unit test, because every message sent would cause a connection notification. This was added outside mollie changes.

ewertons (approved):

momuno (approved):

PR #2097, opened 2021-10-01
pass empty json on null method payload

### LINKED ISSUE #1924: Convenience layer device method callbacks aren't passed empty payload
First big thanks to @digimaun for finding this.

If IoT Hub service client send no payload to a convenience layer device method callback, the SDK wrongly ignores it and doesn't pass it on.

Repo:
1 - Install `az` tool.  It needs to be a newer version; old versions won't repo.
2 - Do `az` setup - like `az login` and `az account set --subscription <yourIoTHub subscription>`.
3 - Build `iothub_client\samples\iothub_client_device_twin_and_methods_sample` from this repo, including setting connection string of device.
4 - Run `iothub_client_device_twin_and_methods_sample`.
5 - Run `az iot hub "invoke-device-method" --device-id <deviceToTestAgainst> --hub-name <hubName> --method-name "getCarVin" --mp null` 

Expected: The callback should be invoked.
Actual: The attempt to duplicate a NULL pointer causes our SDK to get confused and reject request.

```
Error: Time:Thu Apr  1 20:07:27 2021 File:C:\iot2\azure-iot-sdk-c\c-utility\src\buffer.c Func:_BUFFER_create Line:61 invalid parameter source: 00000000
Error: Time:Thu Apr  1 20:07:27 2021 File:C:\iot2\azure-iot-sdk-c\iothub_client\src\iothub_client_core.c Func:_make_method_calback_queue_context Line:350 BUFFER_create failed
Error: Time:Thu Apr  1 20:07:27 2021 File:C:\iot2\azure-iot-sdk-c\iothub_client\src\iothub_client_core.c Func:_iothub_ll_device_method_callback Line:391 construction of method calback queue context failed
```

```c
static int make_method_calback_queue_context(USER_CALLBACK_INFO* queue_cb_info, const char* method_name, const unsigned char* payload, size_t size, METHOD_HANDLE method_id, IOTHUB_QUEUE_CONTEXT* queue_context)
{
    [snip…]
    else
    {
        if ((queue_cb_info->iothub_callback.method_cb_info.payload = BUFFER_create(payload, size)) == NULL)
        {
```

The payload here is NULL with size=0 so BUFFER_create fails.  But this is after LL layer has handed off the next phase to convenience.

danewalton:
closes #1924

### DIFF
--- a/iothub_client/src/iothub_client_core.c
+++ b/iothub_client/src/iothub_client_core.c
@@ -387,10 +387,10 @@ static int iothub_ll_device_method_callback(const char* method_name, const unsig
         queue_cb_info.type = CALLBACK_TYPE_DEVICE_METHOD;
 
         // A NULL payload was sent from the service. Use empty JSON to signal to the user no payload.
-        if(size == 0)
+        if (size == 0)
         {
           payload = (const unsigned char*) CLIENT_CORE_METHOD_EMPTY_PAYLOAD;
-          size = strlen(CLIENT_CORE_METHOD_EMPTY_PAYLOAD);
+          size = sizeof(CLIENT_CORE_METHOD_EMPTY_PAYLOAD) - 1;
         }
 
         result = make_method_calback_queue_context(&queue_cb_info, method_name, payload, size, method_id, queue_context);
--- a/iothub_client/src/iothub_client_core.c
+++ b/iothub_client/src/iothub_client_core.c
@@ -26,6 +26,7 @@
 
 #define DO_WORK_FREQ_DEFAULT 1
 #define DO_WORK_MAXIMUM_ALLOWED_FREQUENCY 100
+#define CLIENT_CORE_METHOD_EMPTY_PAYLOAD "{}"
 
 struct IOTHUB_QUEUE_CONTEXT_TAG;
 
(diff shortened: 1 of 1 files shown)


danewalton:
I went through the process you (@jspaith) detailed which caused the issue. I'm now getting a successful response.
image

iothub_client/src/iothub_client_core.c
```diff
@@ -385,6 +386,13 @@ static int iothub_ll_device_method_callback(const char* method_name, const unsig
         USER_CALLBACK_INFO queue_cb_info;
         queue_cb_info.type = CALLBACK_TYPE_DEVICE_METHOD;
 
+        // A NULL payload was sent from the service. Use empty JSON to signal to the user no payload.
+        if(size == 0)
```
ericwol-msft, line 390:
```suggestion
        if (size == 0)
```

iothub_client/src/iothub_client_core.c
```diff
@@ -385,6 +386,13 @@ static int iothub_ll_device_method_callback(const char* method_name, const unsig
...
+        {
+          payload = (const unsigned char*) CLIENT_CORE_METHOD_EMPTY_PAYLOAD;
+          size = strlen(CLIENT_CORE_METHOD_EMPTY_PAYLOAD);
```
ericwol-msft, line 393: sizeof?

ericwol-msft (approved):

PR #2117, opened 2021-10-20
Add retry for device|module creation in tests; code refactor

jspaith:
All E2E tests run through a common helper file that needed some attention:

* Primary change is having device and module creation be retried on failure.  Rough analysis of our last few week gate runs suggests a 3-5% false positive rate on tests because Hub refuses device creation (likely due to test stress, and their SLA IIRC suggests retry).  New code arbitrarily choses a max of 3 retries with a 30 second wait between attempts, though that's trivial to change if we want.

  Note that these changes are in test helper, NOT the core service SDK (which would be much more invasive and a functional change on customers).

* Remove `MBED_BUILD_TIMESTAMP` #ifdef.  Even cursory examination of `IoTHubAccount_Init_With_Config` shows that this flag doesn't work.

* Cleanup some of the functions layouts.  Especially get away from very deeply nested `else` blocks and move `free` code in one place instead of duplicated.

* Remove unused global variables / functions and move constants to top of file.

### DIFF
--- a/.github/scripts/build_doxygen.sh
+++ b/.github/scripts/build_doxygen.sh
@@ -22,4 +22,4 @@ popd
 popd
 popd
 
-doxygen docs/Doxyfile
+doxygen doc/Doxyfile


jspaith:
@ericwol-msft - thanks for feedback!  I've taken them all in latest commit.

PR #2138, opened 2021-11-01
Remove dead code 

jspaith:
Removes various instances of dead-code:
* Remove many "#if 0" instances.  
  * In UT specifically this tended to be funcitons we never actually productized.
  * For `iothubtransporthttp_ut.c`, removed the #if 0'd failure tests (which were done a failure per test case, which is tedious) because more modern `IoTHubTransportHttp_Register_HappyPath_with_deviceKey_fail`.
  * There is one #if 0 left in `iothubtransporthttp_ut.c` after this PR (only such one).  There's thousands of lines of tests that have never been compiled.  This will need more investigation to bring back/remove based on whether needed or no.

* Delete files that were not compiled.

* Cleanup the UT handling of the real_<...>.c.  Many of these files were duplicated in UT directory instead of shared source.
  * Delete duplicated files.
  * Cleanup places using absolute path instead of the `SHARED_UTIL_REAL_TEST_FOLDER` macro


danewalton:
17.5 thousand deletions....

img

jspaith:
Note to reviewers - fixing the embarrassing build break at `hsm_client_tpm_decrypt_data_handle_NULL_fail`  (for some reason I left a function that never would compile in an #if 0 block and forgot to pre-build).  I'm in process of reviewing all other spots including re-build/rerun of all UT.

danewalton (approved):
why is `provisioning_client/src/secure_device_tpm.c` being deleted?

Also double checking....are any of these files which are being deleted in any CMakeLists.txt files?

jspaith:
1 -- 
I've gone back and forth with `secure_device_tpm.c`.  

The biggest reason I want to get rid of it is if you look at the [file](https://github.com/Azure/azure-iot-sdk-c/blob/master/provisioning_client/src/secure_device_tpm.c), it leads with an `#if 0` on top but there's not even a matching `#endif` at the bottom.

I also can't find any meaningful references to it across Internet despite multiple search engines to lookup.  It certainly wasn't referenced by us anywhere or it would've broken build.

2 --
"are any of these files which are being deleted in any CMakeLists.txt files" -> some of them are, like the `real_...`.  But here the fix was a straightforward pointing at right file.

iothub_service_client/tests/iothub_msging_ll_ut/iothub_messaging_ll_ut.c
```diff
@@ -1453,76 +1453,6 @@ BEGIN_TEST_SUITE(iothub_messaging_ll_ut)
         IoTHubMessaging_LL_Destroy(iothub_messaging_handle);
     }
 
-#if 0
```
ewertons, line 1456: I'm 100% up for removing everything that has been disabled like that without a justification, and if we verify the tests are really not needed. If they are we maybe should file a task to re-add them appropriately.

ewertons (approved):

iothub_service_client/tests/iothub_msging_ll_ut/iothub_messaging_ll_ut.c
jspaith: 100% agree - if these tests were needed/intereting I would've re-enabled or created an IOU to bring them back.

This case we're OK though.

`IoTHubMessaging_LL_SendModule` is an API we'd considered back in the day but removed prior to GA and it's not something I see coming back for many reasons (both for IoT Hub philosophy & because we don't add new things to service SDK for C).

Thanks for thorough review.

PR #2143, opened 2021-11-03
Cleanup MQTT responseId parsing

jspaith:
Based off of #1436 by @fgiancane8 - once again thanks Francesco for your help finding this & contributing back.

The initial #1436 was about catching some variables that were maybe uninitialized.  From follow-up, this was on a gcc on ESP32 since our gates should normally catch it.  Attempts to repro with various gcc versions and explicit `maybe-uninitialized` along with various optimization flags were unsuccessful on my part.

Refactor code a bit to hopefully make var setting more clear.  

That said, there's another problem here.  The initial line to parse for the requestId part of the token isn't right.
```c
STRING_TOKENIZER_get_next_token(token_handle, token_value, "/?$rid=")
```

The tokenizer is treating the "/?$rid=" as delimiters and not the full string.  As a quick experiment I put "/?$Qid=" as the argument and still got a 0 return value (success) on scanning for the passed string of "rid" and not "qid".  This isn't what we want.

So... as an additional change here, continue to tokenize just on the single "/" and then later check that the "/?$rid=" is received.  I moved the code into the "main" loop (the new `token_count==4` block) to maintain symmetry with existing code now that we're using "/" like rest of it.

This will cause topics that previously "passed" to fail now.  Justification::
* We're on a callback thread so this change won't impact the API caller.  
* Since the requestID would've been busted most of the time the caller would've ended up not matching it to anything in its table anyway.  
* Review of both the [Azure IoT MQTT twin spec](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#retrieving-a-device-twins-properties) and the Azure Embedded SDK where [?rid is explicitly checked](https://github.com/Azure/azure-sdk-for-c/blob/3bff336816e7e03ac5add09c6911683d981ac6ae/sdk/src/azure/iot/az_iot_hub_client_twin.c#L163) make me comfortable making the lack of this value a failure (where it wasn't previously).

Furthermore cleanup UT to actually return the more realistic "/rid=" during test data and some tests to simulate this situation.

### DIFF
--- a/iothub_client/src/iothubtransport_mqtt_common.c
+++ b/iothub_client/src/iothubtransport_mqtt_common.c
@@ -83,7 +83,8 @@ static const char* DEVICE_METHOD_RESPONSE_TOPIC = "$iothub/methods/res/%d/?$rid=
 
 static const char SYS_TOPIC_STRING_FORMAT[] = "%s%%24.%s=%s";
 
-static const char* REQUEST_ID_PROPERTY = "?$rid=";
+static const char REQUEST_ID_PROPERTY[] = "?$rid=";
+static size_t REQUEST_ID_PROPERTY_LEN = sizeof(REQUEST_ID_PROPERTY) - 1;
 
 #define SYS_PROP_MESSAGE_ID "mid"
 #define SYS_PROP_MESSAGE_CREATION_TIME_UTC "ctime"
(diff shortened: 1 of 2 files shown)


danewalton (approved):

iothub_client/src/iothubtransport_mqtt_common.c
```diff
@@ -83,7 +83,8 @@ static const char* DEVICE_METHOD_RESPONSE_TOPIC = "$iothub/methods/res/%d/?$rid=
 
 static const char SYS_TOPIC_STRING_FORMAT[] = "%s%%24.%s=%s";
 
-static const char* REQUEST_ID_PROPERTY = "?$rid=";
+static const char REQUEST_ID_PROPERTY[] = "?$rid=";
+static size_t REQUEST_ID_PROPERTY_LEN = sizeof(REQUEST_ID_PROPERTY) - 1;
```
ewertons, line 87: Not necessary here, but I've been using the following macro:
```c 
#define lengthof(x) (sizeof(x) - 1)
```
Useful when we have several *_LEN defines in a single c module, since you reduce the number of defines to just one.

ewertons (approved):

iothub_client/src/iothubtransport_mqtt_common.c
jspaith: Indeed!  Of all the macro utils we don't have a str_len one like that!  I don't see other places in file using it enough so will keep this as-is for now.

PR #2155, opened 2021-11-16
Address intermittently failing GetFullDesired_e2e tests

momuno:
Intermittent IoTHub_AMQP_WS_Module_GetFullDesired_e2e_x509 test failures appear to be due to the unavoidable race condition mentioned in the test comments.  *It is not clear why the race condition shows up in this specific test more than any others (MQTT/WS/sas).* The twin is never received by the device client after the server client updates the twin on the hub.  As a result, the receive callback loop runs until the timeout, causing the test to fail.

Instead of using `ThreadAPI_Sleep()`, which may (or may not) be long enough to avoid the condition, this change makes the twin callback registration step synchronous with respect to the remaining steps.  This will hopefully help avoid red herring test failures.


iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -651,6 +651,40 @@ void dt_e2e_get_complete_desired_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol,
...
+        (difftime(nowTime, beginOperation) < MAX_CLOUD_TRAVEL_TIME) // time box
+        )
+    {
+        if (Lock(device->lock) != LOCK_OK)
+        {
+            ASSERT_FAIL("Lock failed");
+        }
+        else
+        {
+            if ((device->receivedCallBack) && (device->cb_payload != NULL))
```
jspaith, line 668: I'm not seeing where you reset `receivedCallBack`.  I think way code currently is, when you get to the "check that post-update twin has been received" it'll immediately see `receivedCallBack` as `true` instead of this being set in callback.

jspaith (changes requested):
@momuno - this is a great change.  Thanks for taking on.  I hate any gate instability even in weird cases like this.

I'm marking request changes because I think you really need to reset the received variable (quick look at gates I think is confirming this).

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -651,6 +651,40 @@ void dt_e2e_get_complete_desired_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol,
 
     int initial_twin_version = dt_e2e_gettwin_version(serviceClientDeviceTwinHandle, deviceToUse);
 
+    // Receive first full twin before updating it.
+    time_t beginOperation, nowTime;
+    beginOperation = time(NULL);
+    while (
```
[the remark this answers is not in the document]
momuno: haha yes it is a bit long. How bout I create a separate card to refactor these tests.  I have already refactored part of the code when I wrote the if-not-version and CBOR tests so have some material I can work off of.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
momuno: ohhhh yes I do need to. Thank you!

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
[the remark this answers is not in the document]
jspaith: Yup, if you want to refactor as a separate item that's fine too.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -651,6 +668,48 @@ void dt_e2e_get_complete_desired_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol,
...
+
+    ASSERT_IS_TRUE(difftime(nowTime, beginOperation) < MAX_CLOUD_TRAVEL_TIME, "Timeout waiting for twin message");
+
+    // Reset device for update.
+    device_desired_reset(device);
+
+    // Sleep still needed.  Device callback has been recorded in test app, but Hub is still completing twin PATCH subscription.
+    // This must be in place before the service SDK can update the twin, else the Hub will not send the PATCH
+    // message to the device and the desired twin callback will not be called.
+    ThreadAPI_Sleep(5000); // 5 seconds
```
jspaith, line 711: The `// 5 seconds` is redundant since it's already self-documented in the `5000` itself.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -541,6 +541,23 @@ static void device_desired_deinit(DEVICE_DESIRED_DATA *device)
     }
 }
 
+static void device_desired_reset(DEVICE_DESIRED_DATA *device)
```
jspaith, line 544: Nit - consider this as `device_receivedcallback_reset` or some such as I think that captures more how variable is named & callback is set.  Your call either way as its minor.

jspaith (approved):

momuno:
Yeah there's some dps cert something or other failures going on right now.  I don't see the relation to my change.

PR #2158, opened 2021-11-18
 Refactor of device twin e2e tests

momuno:
This is a general refactor to allow future tests to be more easily added.
All 4 test suites (MQTT/AMQP/module/device) have been run against the changes and pass.


iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,568 +10,728 @@
...
+    JSON_Value* root_value = json_parse_string(twin_payload);
+    ASSERT_IS_NOT_NULL(root_value);
+    JSON_Object* root_object = json_value_get_object(root_value);
+    ASSERT_IS_NOT_NULL(root_object);
+
+    JSON_Array* array = json_object_dotget_array(root_object, full_property_name);
+    ASSERT_IS_NOT_NULL(array, "Array not specified.");
+
+    double double_value = json_array_get_number(array, index);
+    int int_value = (int)(double_value + 0.1); // Account for possible underflow by small increment.
```
jspaith, line 382: I don't think we should cast to `int` - the standard mechanism is do the check as a double and then allow for a tolerance of rounding error.  The test code [here](https://github.com/jspaith/azure-sdk-for-c/blob/84bfe57b50091d88dc993ce80e95ca435aaa0d16/sdk/tests/iot/hub/test_az_iot_hub_client_properties.c#L1237) in embedded SDK demonstrates.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,568 +10,728 @@
...
+        twin_response = IoTHubDeviceTwin_UpdateTwin(serviceclient_devicetwin_handle,
+                                                    device_to_use->deviceId, twin_json);
     }
 
-    ASSERT_ARE_EQUAL(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_OK, result, "SendEventAsync failed");
-}
-
+    ASSERT_IS_NOT_NULL(twin_response, "IoTHubDeviceTwin_Update(Module)Twin failed.");
 
+    LogInfo("_service_client_update_twin(): Twin response from Service SDK after update is <%s>.\n",
```
jspaith, line 419: Please remove the "\n" in LogInfo string - by default it'll concat this for us and most other places don't do this.  I understand some of legacy code in this did the \n too but it should get cleaned up too.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,568 +10,728 @@
...
+}
+
+static void _receive_twin_loop(RECEIVED_TWIN_DATA* received_twin_data, DEVICE_TWIN_UPDATE_STATE expected_update_state)
+{
+    time_t begin_operation = time(NULL);
+    time_t now_time;
+
+    LogInfo("_receive_twin_loop(): Entering loop.\n");
+
+    while (now_time = time(NULL), difftime(now_time, begin_operation) < MAX_CLOUD_TRAVEL_TIME)
```
jspaith, line 598: Nit - could we put a parattheses around the difftime(...) <MAX....  I get this should be OK but whenever I see commas I start thinking about order of operation stuff and the paren would make it clearer.  Similar feedback for other place that does this.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -586,390 +746,238 @@ void client_create_with_properies_and_send_d2c(IOTHUB_PROVISIONED_DEVICE* device
...
-        ASSERT_FAIL("Lock failed");
-    }
-    else
-    {
-        ASSERT_IS_NOT_NULL(root_value, "json_parse_string failed");
-        ASSERT_ARE_EQUAL(char_ptr, expected_desired_string, string_property, "string data retrieved differs from expected");
-        ASSERT_ARE_EQUAL(int, expected_desired_integer, integer_property, "integer data retrieved differs from expected");
-        ASSERT_ARE_EQUAL(char_ptr, expected_desired_string, string_property_from_array, "string data (from array) retrieved differs from expected");
-        ASSERT_ARE_EQUAL(int, expected_desired_integer, integer_property_from_array, "integer data (from array) retrieved differs from expected");
+    ThreadAPI_Sleep(5000);
```
jspaith, line 884: 5000 is magic # in code - could we put it as a `static const unsigned int` at top of file and the places that reference it reference that one instead?

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -586,390 +746,238 @@ void client_create_with_properies_and_send_d2c(IOTHUB_PROVISIONED_DEVICE* device
...
+                                                                       account_auth_method);
+    REPORTED_TWIN_DATA* reported_twin_data = _reported_twin_data_init();
 
-    dt_e2e_create_client_handle(deviceToUse, protocol);
+    _create_client_handle(device_to_use, protocol);
 
-    // subscribe
-    setdevicetwincallback_on_device_or_module(deviceTwinCallback, device);
+    // Connect device client to IoT Hub.
+    // Send reported payload. Register callback. Recieve reported state message.
```
jspaith, line 795: Nit here & other places == "Recieve" =>" Receive"

momuno:
@jspaith thank you so much for doing a review.  There were a lot of changes which can be more work to get through.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,568 +10,728 @@
...
+
+    ASSERT_ARE_EQUAL(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_OK, result,
+                     "IoTHub(Device|Module)Client_SendEventAsync failed.");
 }
 
-void client_create_with_properies_and_send_d2c(IOTHUB_PROVISIONED_DEVICE* deviceToUse, MAP_HANDLE mapHandle)
-{
-    IOTHUB_MESSAGE_HANDLE msgHandle;
 
+void _client_create_with_properties_and_send_d2c(IOTHUB_PROVISIONED_DEVICE* device_to_use,
```
jspaith, line 720: Nit: Since `_client_create_with_properties_and_send_d2c` is only ever used to send the KillTCP message, could we rename thos to like `send_tcp_kill` or something and then move all the code around setting up the MAP_HANDLE in functions below into this one?

I think this would also benefit from some level of comments since this isn't documented as far as I know.  So something like "// IoT Hubs can be configured so that a special device telemetry message will cause them to close the TCP connection.  // This is used to simulate network failures and check the SDK's reconnection logic"

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -586,390 +746,238 @@ void client_create_with_properies_and_send_d2c(IOTHUB_PROVISIONED_DEVICE* device
...
-                root_value = NULL;
-            }
-            Unlock(device->lock);
-        }
-        ThreadAPI_Sleep(1000);
-    }
+    _create_client_handle(device_to_use, protocol);
 
-    ASSERT_IS_TRUE(difftime(nowTime, beginOperation) < MAX_CLOUD_TRAVEL_TIME, "Timeout waiting for twin message");
+    // Twin registrations to the cloud happen asyncronously because we're using the convenience
```
jspaith, line 862: I think this isn't true anymore, because you added the block 
```
    _set_twin_callback(_twin_callback, received_twin_data);
    _receive_twin_loop(received_twin_data, DEVICE_TWIN_UPDATE_COMPLETE);
    _received_twin_data_reset(received_twin_data);
```

Because now we're not doing anything with the service SDK until we've actually setup the SUBSCRIBE and it's happily got the device SDK the twin.  ??  

If this is case we should remove commend and I think (!) we can remove the `ThreadAPI_Sleep(5000);` after this since we've made this deterministic.  Similar feedback in any other functions that Sleep() after block above (though the KillTCP message we absolutely have to Sleep() after sending)

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,568 +10,728 @@
...
-        result = IoTHubDeviceClient_SendReportedState(iothub_deviceclient_handle, (unsigned char *) buffer, bufferLen, reportedStateCallback, device);
+        (void)sprintf(return_value, REPORTED_PAYLOAD_FORMAT, num, string, num, string);
     }
 
-    ASSERT_ARE_EQUAL(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_OK, result, "IoTHub(Device|Module)Client_SendReportedState failed");
+    return return_value;
 }
 
-static void setdevicetwincallback_on_device_or_module(IOTHUB_CLIENT_DEVICE_TWIN_CALLBACK twinCallback, DEVICE_DESIRED_DATA *device)
+static char* _parse_json_twin_char(const char* twin_payload, const char* full_property_name)
```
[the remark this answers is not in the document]
momuno: I recall I split things up because there will need to be differentiation betwen json and cbor eventually.  But I think the same concept could be applied here, having a separate set of functions of similar nature for cbor.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
momuno: lolol

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
momuno: There is still a race condition for subscribe to twin patch topic.  We have to wait for service to finish this otherwise we can end up in an endless loop.  MQTT seems to behave ok, but AMQP takes longer to finish the subscribe.... this is where the amqp test failures were happening.

momuno: i have rewritten the comment to reflect the updated test scneraio

jspaith (approved):
The obligatory "jspaith has to earn his keep nitpicks", but generally looks good.  Thanks for taking this on!

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,1134 +10,955 @@
...
 
-void dt_e2e_deinit(void)
+static unsigned char* calloc_and_copy_unsigned_char(const unsigned char* payload, size_t size)
 {
-    IoTHubAccount_deinit(g_iothubAcctInfo);
-
-    // Need a double deinit
-    IoTHub_Deinit();
-    IoTHub_Deinit();
+    unsigned char* result = (unsigned char*)calloc(1, size + 1);
```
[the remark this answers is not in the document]
momuno: I'm not clear on why we would not need a terminating null.  null is still 0x00 whether its signed or unsigned.

iothub_client/tests/common_dt_e2e/iothubclient_common_dt_e2e.c
```diff
@@ -10,1134 +10,955 @@
...
+    JSON_Object* root_object = json_value_get_object(root_value);
+    ASSERT_IS_NOT_NULL(root_object);
 
-    if (iothub_moduleclient_handle != NULL)
-    {
-        IoTHubModuleClient_Destroy(iothub_moduleclient_handle);
-        iothub_moduleclient_handle = NULL;
-    }
-    LogInfo("Completed destroy of IotHub client handle");
+    const unsigned char* json_str =  (unsigned char*)json_object_dotget_string(root_object,
```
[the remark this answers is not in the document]
momuno: good eye.  all i can think is that got left over from other changes i was making and I didn't revert it?

PR #2166, opened 2021-11-23
update CMake function to be more descriptive

danewalton:
nitpick I've found updating the CMake files. Making it a separate PR to keep the other changes on topic.


configs/azure_iot_sdksFunctions.cmake
```diff
@@ -137,7 +137,7 @@ function(add_longhaul_test_directory test_directory)
 endfunction()
 
 # For targets which set warning switches as project properties (e.g. XCode)
-function(setSdkTargetBuildProperties stbp_target)
+function(applyXcodeBuildFlags stbp_target)
```
jspaith, line 140: Nit - consider `applyXcodeBuildFlagsIfNeeded` as name of this.  Since it's called unconditionally every place else, a very casual observer may or may not assume Xcode somehow tied into MSVC or gcc or even clang somehow.

Not sure if that captures the spirit of CMake or if folks - even casual observers - would just know without `ifNeeded` part.

jspaith (commented):
small nit for you to consider or no; otherwise looks good.

configs/azure_iot_sdksFunctions.cmake
danewalton: ahh good callout yup will do

ericwol-msft (approved):

PR #2168, opened 2021-11-24
Add mbedTLS e2e tests

danewalton:


iothub_service_client/src/iothub_messaging_ll.c
```diff
@@ -1383,6 +1383,7 @@ IOTHUB_MESSAGING_RESULT IoTHubMessaging_LL_Send(IOTHUB_MESSAGING_HANDLE messagin
 
                 if (message_add_body_amqp_data(amqpMessage, binary_data) != 0)
                 {
+                    message_destroy(amqpMessage);
```
jspaith, line 1386: I don't understand this code - namely pre-existing of pattern you're following here.

We unconditionally do a `message_destory` in line 1417 (of code before your change).  It actually looks like this pre-existing code along with this change in new block will cause a double-free.  Am I mis-reading something?

jspaith (changes requested):
The gate changes look great, thank you.

I don't understand the AMQP change though.  I may only need education, but I want to discuss before we merge this in.

iothub_service_client/src/iothub_messaging_ll.c
danewalton: image

Yea I should have added more context here. So the unconditional gets run even in failure cases (lines 1409 and 1417 above triggered it). In the error paths above it, every one destroys the `amqpMessage` and then pops out of the if/else's into the unconditional destroy. So if there is ever an error with any of these, there will be a double free. By removing the unconditional and putting it with the success case too, we do it only once. The remedy here is to either do what I did or remove the destroy from all the errors and destroy unconditionally (after checking it successfully was created).

There was a gate run with valgrind and debug info before which showed line numbers but below is the flow which triggered the issue.

image

jspaith: Sorry - I missed that you had removed the uncoditional free at the bottom.  Code as you have makes sense.

My nit would be that the unconditional free at bottom is safer than having it in each error block, since if we add new elseif(...) {...} blocks that needs to be dup'd.  SDK has that pattern of a per-block free a bunch, which I tend to clean up as needed.

jspaith (approved):

iothub_service_client/src/iothub_messaging_ll.c
danewalton: Yea I can reformat it to follow the pattern which `to_amqp_value` in the same code block has. The unconditional works in that case. I think that's "more correct"

PR #2174, opened 2021-12-03
Fixes for SDLNativeRules issues 

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution


iothub_client/src/iothub_client_core_ll.c
```diff
@@ -833,6 +833,13 @@ static IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* initialize_iothub_client(const IOTHUB_
                     config = client_config;
                 }
             }
+            else if (device_config == NULL)
```
ewertons, line 836: Is this right?? on lines 775+ it looks like device_config == NULL is a valid scenario.

provisioning_client/tools/dice_device_enrollment/dice_device_enrollment.c
```diff
@@ -17,7 +17,14 @@ static const char* const END_CERT_MARKER = "-----END CERTIFICATE-----";
 
 static char* get_user_input(const char* text_value, int max_len)
 {
-    char* result = malloc(max_len + 1);
+    int result_len = max_len + 1;
+    if (result_len <= 1)
+    {
+        (void)printf("invalid max_len size\r\n");
+        return NULL;
```
ewertons, line 24: It breaks our style of single point of return.

provisioning_client/tools/symm_key_provision/symm_key_provision.c
```diff
@@ -20,7 +20,14 @@ typedef struct REGISTRATION_INFO_TAG
 
 static char* get_user_input(const char* text_value, int max_len)
 {
-    char* result = malloc(max_len + 1);
+    int result_len = max_len + 1;
+    if (result_len <= 1)
+    {
+        (void)printf("invalid max_len size\r\n");
+        return NULL;
```
ewertons, line 27: It breaks our style of single point of return.

iothub_client/src/iothub_client_core_ll.c
ericwol-msft: without it, it would fall into the else and at line 847 cause an AV

ericwol-msft: fixed

provisioning_client/tools/dice_device_enrollment/dice_device_enrollment.c
ericwol-msft: line 31 already does an early return

provisioning_client/tools/symm_key_provision/symm_key_provision.c
ericwol-msft: line 34 already does an early return

ewertons (approved):

PR #2181, opened 2021-12-14
Update top level CMake

danewalton:


deps/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+    endif ()
+endif()
+
+########## Parson ##########
+if(NOT ${use_installed_dependencies})
+  add_library(parson
+      ${CMAKE_CURRENT_LIST_DIR}/parson/parson.c
+      ${CMAKE_CURRENT_LIST_DIR}/parson/parson.h
+  )
+  add_library(parson::parson ALIAS parson)
```
momuno, line 42: curious the reason for parson::parson

provisioning_client/CMakeLists.txt
```diff
@@ -405,4 +406,4 @@ else()
         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
         INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
     )
-endif()
\ No newline at end of file
+endif()
```
momuno, line 409: yay!

deps/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+        find_package(azure_macro_utils_c REQUIRED CONFIG)
+    endif ()
+    if (NOT umock_cFOUND)
+        find_package(umock_c REQUIRED CONFIG)
+    endif ()
+endif()
+
+########## Parson ##########
+if(NOT ${use_installed_dependencies})
+  add_library(parson
```
momuno, line 38: Looks like we are not installing parson separately if the app wants to "use_installed_dependencies".  Is this the same behavior as before or different?

deps/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+        add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/umock-c)
+    endif()
+else()
+    if (NOT azure_macro_utils_cFOUND)
+        find_package(azure_macro_utils_c REQUIRED CONFIG)
+    endif ()
+    if (NOT umock_cFOUND)
+        find_package(umock_c REQUIRED CONFIG)
+    endif ()
+endif()
```
momuno, line 34: Why do we not incude the uhttp in this section? 

Side thought: Are you considering moving utpm, umqtt and uamqp from the top level into deps/ ?  That seems more intuitive to me.  But may not be worth it for risk of backwards compatibility.

momuno (approved):
Depending on how much is getting changed with installed dependencies, we may/may not run into an issue with Vcpkg since it uses installed dep.  Just a heads up, I may need to recruit you in January :)

provisioning_client/CMakeLists.txt
momuno: for a new line.  at end. 
that is all.

deps/CMakeLists.txt
danewalton: image

In theory I think I agree with you but at this point a little less of a concern for me. Maybe in the future yes but you're right about the back compat. I'll likely leave for now.

deps/CMakeLists.txt
danewalton: Thanks for the callout it looks like the newer version of parson which we pull already does this. Looking to remove and use parson's CMake via `add_subdirectory()`

deps/CMakeLists.txt
danewalton: this will likely be removed but in general, namespacing has become common practice on CMake targets so that multiple libraries from a provider can be similarly categorized. We might have something like `az::iot_hub_client` and `az::iot_provisioning_client`. Just so happens here that parson only has one thing :)

danewalton:
Back compat is also working so I'm going to go ahead and merge this.

deps/CMakeLists.txt
momuno: oh nice

PR #2186, opened 2021-12-22
Add instruction to sign certs with intermediate CA

nkgami:
Default method "create_device_certificate" will create a device certificate which is chained up to the root certificate directly.
In group enrollment scenario of Device Provisioning Service, users may want the device certs to be chained up to the intermediate certificate. So "create_device_certificate_from_intermediate" should be noted.

 

# Checklist
- [N/A because only .md file will be changed] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

https://docs.microsoft.com/en-us/azure/iot-dps/tutorial-group-enrollments
This tutorial will not work with "create_device_certificate" command, because the device certs need to be signed by the intermediate CA.

# Description of the problem
"create_device_certificate_from_intermediate" should be noted because device certificates need to be signed by the intermediate CA in many scenarios.

# Description of the solution
Add notes about this command to the .md file.

### DIFF
--- a/iothub_client/samples/iothub_ll_client_x509_sample/readme.md
+++ b/iothub_client/samples/iothub_ll_client_x509_sample/readme.md
@@ -2,9 +2,9 @@
 
 ## Build instructions
 
-1. Modify `.\iothub_ll_client_x509_sample.c`: update `connectionString`, `x509certificate` and 
-`x509privatekey`.
-   (Optionally, you can change configuration regarding transport, global endpoint, etc.)
+1. Modify `.\iothub_ll_client_x509_sample.c`: update the `connectionString`, `x509certificate` and 
+`x509privatekey` variables.
+   (Optionally, you can change configuration regarding transport, etc.)
 
 1. If you are using an OpenSSL engine, uncomment the `#define SAMPLE_OPENSSL_ENGINE` and set it to the
    OpenSSL Engine name. Instead of a PEM certificate, `x509privatekey` will contain the private key ID
@@ -32,21 +32,21 @@ cd build
 
 ### TLS Options
 
-Additional TLS stack specific configuration is available through `IoTHubDeviceClient_LL_SetOption`.
+Additional TLS stack specific configuration is available through `IoTHubDeviceClient_LL_SetOption`. Some of the available options can be found [here](https://github.com/Azure/azure-iot-sdk-c/blob/main/doc/Iothub_sdk_options.md#common-transport-options).
 
 ### Long X509 Client Certificate Chains
 
 When using chains with more than one level, the device must send all chain certificates up to but 
-not including the CA configured on the service. The certificates will be sent part of the TLS 
+not including the CA configured on the service. The certificates will be sent as part of the TLS 
 handshake in order for the service to build a valid chain. 
 
 For TLS stacks such as OpenSSL and mbedTLS, this is achieved by creating a PEM certificate store 
 (i.e. concatenating the Intermediates as well as the Device Certificate PEM contents) and passing 
-it to the C-SDK via `x509certificate`.
+it to the C-SDK via the `x509certificate` variable.
 
-In Windows, all intermediates must be present in the Intermediate Certificate Store and 
-`x509certificate` should contain only the device certificate. SChannel will automatically build the 
-chain and send the required certificates to the service.
+In Windows, all intermediates must be present in the Intermediate Certificate Store and the 
+`x509certificate` variable should contain only the device certificate. SChannel will automatically 
+build the chain and send the required certificates to the service.
 
 __Note:__ Using a multi-level chain increases the overhead of each TLS connection. All intermediate 
 certificates unknown to the service must be sent on every new connection. Both the client and 
--- a/iothub_client/samples/iothub_ll_client_x509_sample/readme.md
+++ b/iothub_client/samples/iothub_ll_client_x509_sample/readme.md
@@ -1,9 +1,9 @@
-# Azure IoT Provisioning Client Sample: Using X509 Client Certificates
+# Azure IoT Hub Client Sample: Using X509 Client Certificates
 
 ## Build instructions
 
-1. Modify `.\prov_dev_client_ll_x509_sample.c`: update `id_scope`, `registration_id`, 
-   `x509certificate` and `x509privatekey`.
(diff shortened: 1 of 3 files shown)


nkgami:
Hello Azure IoT SDK C team  @ericwol-msft @danewalton @momuno  , Can you please check and merge this PR?

tools/CACertificates/CACertificateOverview.md
```diff
@@ -109,8 +109,8 @@ This will create files mydevice* that contain the public key, private key, and P
 
 #### IoT Leaf Device
 
-* Run `./certGen.sh create_device_certificate mydevice` to create the new device certificate.  
-  This will create the files ./certs/new-device.* that contain the public key and PFX and ./private/new-device.key.pem that contains the device's private key.  
+* Run `./certGen.sh create_device_certificate mydevice` to create the new device certificate. If the device certificate needs to be signed by intermediate CA, run `./certGen.sh create_device_certificate_from_intermediate mydevice` instead to chain up the device certificate to the intermediate certificate.
```
momuno, line 112:
```suggestion
* Run `./certGen.sh create_device_certificate mydevice` to create the new device certificate, which will be chained to the root CA directly.   If the device certificate needs to be signed instead by intermediate CA, run `./certGen.sh create_device_certificate_from_intermediate mydevice` to chain up the device certificate to the intermediate certificate.  
```

CIPop: @nkgami  Please split this into a different paragraph that explains longer chains. There are 2 important things to add in this new documentation for longer chains:
1. If the service (DPS or Hub) was configured with the (top-level) CA, the device is responsible with sending _all_ certificates up to but not including the CA. This is mandatory and allows the service to create a valid chain. 
    - For most TLS stacks (tested on OpenSSL, mbedTLS) this is achieved by creating a PEM certificate store (i.e. concatenating the Intermediates as well as the Device Certificate PEM contents) and passing that to C-SDK. In Windows, all intermediates must be present in the Intermediate Certificate Store and the chain will be automatically built and sent by SChannel.
2. Using a multi-level chain increases the overheads of each TLS connection because the intermediate certificates must be sent every new connection (e.g. for RSA this can add sending several kB of overhead data). It's common that both the client TLS stack and the service have limitations on the TLS Client Hello packet and so long chains (or chains containing several large certificates) may not work.

CIPop (changes requested):
In general, we don't recommend long chains due to overheads.
Since we're adding a way to generate the chains, we need to document some of the problems these create and C-SDK / TLS stack usage.

CIPop (approved):

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
```diff
@@ -0,0 +1,54 @@
...
+# At the root of the repository
+cd build
+./iothub_client/samples/iothub_ll_client_x509_sample/iothub_ll_client_x509_sample
+```
+
+## Configuring the TLS Stack
+
+### TLS Options
+
+Additional TLS stack specific configuration is available through `IoTHubDeviceClient_LL_SetOption`.
```
danewalton, line 35:
```suggestion
Additional TLS stack specific configuration is available through `IoTHubDeviceClient_LL_SetOption`. [Some of the available options can be found here.](https://github.com/Azure/azure-iot-sdk-c/blob/main/doc/Iothub_sdk_options.md#common-transport-options)
```

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Long X509 Client Certificate Chains
+
+When using chains with more than one level, the device must send all chain certificates up to but 
+not including the CA configured on the service. The certificates will be sent part of the TLS
```
danewalton, line 40:
```suggestion
not including the CA configured on the service. The certificates will be sent as part of the TLS 
```

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Long X509 Client Certificate Chains
+
+When using chains with more than one level, the device must send all chain certificates up to but 
+not including the CA configured on the service. The certificates will be sent part of the TLS 
+handshake in order for the service to build a valid chain. 
+
+For TLS stacks such as OpenSSL and mbedTLS, this is achieved by creating a PEM certificate store 
+(i.e. concatenating the Intermediates as well as the Device Certificate PEM contents) and passing 
+it to the C-SDK via `x509certificate`.
```
danewalton, line 45: is this via the option or via the variable? we should be specific.

provisioning_client/samples/prov_dev_client_ll_x509_sample/readme.md
```diff
@@ -38,3 +40,22 @@ Additional TLS stack specific configuration is available through the following t
...
+
+### Long X509 Client Certificate Chains
+
+When using chains with more than one level, the device must send all chain certificates up to but 
+not including the CA configured on the service. The certificates will be sent part of the TLS 
+handshake in order for the service to build a valid chain. 
+
+For TLS stacks such as OpenSSL and mbedTLS, this is achieved by creating a PEM certificate store 
+(i.e. concatenating the Intermediates as well as the Device Certificate PEM contents) and passing 
+it to the C-SDK via `x509certificate`.
```
danewalton, line 52: same question here

danewalton (approved):

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
CIPop: Good point: this is via the variable. Adding this to the doc here as well as at the top.

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
CIPop: Added in both X509 samples.

iothub_client/samples/iothub_ll_client_x509_sample/readme.md
CIPop: Fixed in both readme files.

PR #2209, opened 2022-01-28
PnP feature gate enablement

jspaith:
`feature/iot_pnp_stage` was having issues building in gate.  Fix (and will require gate runs ongoing to make this not happen again.)

* Check correct value from `strtod`.
* Include parson.h in the build for iothub_client; remove conditional include.
  * Note that we *always* link to parson, so including a header shouldn't impact RAM|ROM impact.  
  * Analysis of gcc and clang ELF binaries created for `pnp_temperature_controller` using `nm` binutil, all parson references are already correctly optimized out.
* Correctly check `snprintf` error cases.  This impacted a fair amount of code but the gist of this is:
  * Use an `int` to store result so that the `< 0` check makes sense.
  * Check for case where sizeof(buffer passed in)== (# of bytes returned).  This implies truncation, which we treat as an error since app won't get properly formatted JSON.
* Don't run the command based x509 tests for `__APPLE__`, to keep inline with existing.

### DIFF
--- a/iothub_client/src/iothub_client_properties.c
+++ b/iothub_client/src/iothub_client_properties.c
@@ -79,7 +79,7 @@ static int PropertySnprintf(char* buffer, size_t count, const char* format, ...)
     va_start(arg_list, format);
 
     int currentOutputBytes = vsnprintf(buffer, count, format, arg_list);
-    if ((currentOutputBytes < 0) || (currentOutputBytes == count))
+    if ((currentOutputBytes < 0) || (currentOutputBytes == (int)count))
     {
         return -1;
     }
--- a/iothub_client/src/iothub_client_properties.c
+++ b/iothub_client/src/iothub_client_properties.c
@@ -56,7 +56,7 @@ static const char* AddCommaIfNeeded(bool lastItem)
 }
 
 // AdvanceCountersAfterWrite updates the traversal pointer and lengths after snprintf'ng JSON
-static void AdvanceCountersAfterWrite(size_t currentOutputBytes, char** currentWrite, size_t* requiredBytes, size_t* remainingBytes)
+static void AdvanceCountersAfterWrite(int currentOutputBytes, char** currentWrite, size_t* requiredBytes, size_t* remainingBytes)
 {
     // The number of bytes to be required is always updated by number the snprintf wrote to.
     *requiredBytes += currentOutputBytes;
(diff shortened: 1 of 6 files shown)


danewalton (approved):

iothub_client/src/iothub_client_properties.c
```diff
@@ -69,137 +71,159 @@ static void AdvanceCountersAfterWrite(size_t currentOutputBytes, char** currentW
...
 
+// To correctly check snprintf for success, we need to check both it's not negative AND 
+// that the (number of bytes written ) != (sizeof buffer).
+static int PropertySnprintf(char* buffer, size_t count, const char* format, ...)
+{
+    va_list arg_list;
+    va_start(arg_list, format);
+
+    int currentOutputBytes = vsnprintf(buffer, count, format, arg_list);
+    if ((currentOutputBytes < 0) || (currentOutputBytes == (int)count))
```
momuno, line 82: Nit:  Would it be more complete to have `currentOutputBytes >= (int)count` ?
Valid range is (0, currentOutputBytes), not (0, currentOutputBytes) U (currentOutputBytes, max positive int]

iothub_client/src/iothub_client_properties.c
```diff
@@ -69,137 +71,159 @@ static void AdvanceCountersAfterWrite(size_t currentOutputBytes, char** currentW
...
-static size_t WriteOpeningBrace(const char* componentName, char* currentWrite, size_t remainingBytes)
+static bool WriteOpeningBrace(const char* componentName, char** currentWrite, size_t* requiredBytes, size_t* remainingBytes)
 {
-    size_t currentOutputBytes;
+    int currentOutputBytes;
 
     if (componentName != NULL)
     {
-        if ((currentOutputBytes = snprintf(currentWrite, remainingBytes, PROPERTY_FORMAT_COMPONENT_START, componentName)) < 0)
+        if ((currentOutputBytes = PropertySnprintf(*currentWrite, *remainingBytes, PROPERTY_FORMAT_COMPONENT_START, componentName)) < 0)
```
momuno, line 98: Nit: Curious why checking for general negative int instead of specific -1 expected return value on failure ?

momuno (approved):

iothub_client/src/iothub_client_properties.c
jspaith: It'd work with -1 as implemented.  But I'm trying to keep this `snprintf` wrapper's caller as close to the original `snprintf` as possible, and on the C API you check negative and not the exact value.

iothub_client/src/iothub_client_properties.c
jspaith: RE differentiating the error - no, I really don't care how I got there just the bytes I wanted to come out didn't come and then to immediatlley cascade error up.

I think code as written is fine, since it captures the two error cases of "-1" and buffer getting terminated we'd actually hit and is a bit closer to other samples I've seen of this being done in past.

Issue #2200, opened 2022-01-24
Crash when using unstable network connection


MartinTJDK:
During shutdown of AMQP - due to errors, logging causes crash due to zero pointer.

**Function:** "static void on_cbs_put_token_complete_callback(void* context, CBS_OPERATION_RESULT operation_result, unsigned int status_code, const char* status_description)"

**Code Line:** "LogError("CBS reported status code %u, error: '%s' for put-token operation for device '%s'", status_code, status_description, instance->device_id);"

**Issue:** The "instance" is nulll

**Stack trace:**
```
==================== CURRENT THREAD STACK =====================
#0  strlen () at /builds/idf/crosstool-NG/.build/HOST-x86_64-w64-mingw32/xtensa-esp32-elf/src/newlib/newlib/libc/machine/xtensa/strlen.S:84
#1  0x4022bdf0 in _vfprintf_r (data=<optimized out>, fp=<optimized out>, fmt0=<optimized out>, ap=...) at /builds/idf/crosstool-NG/.build/HOST-x86_64-w64-mingw32/xtensa-esp32-elf/src/newlib/newlib/libc/stdio/vfprintf.c:1528
#2  0x4022d5da in vprintf (fmt=0x3f441619 \"CBS reported status code %u, error: '%s' for put-token operation for device '%s'\", ap=...) at /builds/idf/crosstool-NG/.build/HOST-x86_64-w64-mingw32/xtensa-esp32-elf/src/newlib/newlib/libc/stdio/vprintf.c:34
#3  0x401697b4 in consolelogger_log (log_category=AZ_LOG_ERROR, file=0x3f44166a \"../esp-azure/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_cbs_auth.c\", func=0x3f441f93 <__func__$5735> \"on_cbs_put_token_complete_callback\", line=182, options=1, format=0x3f441619 \"CBS reported status code %u, error: '%s' for put-token operation for device '%s'\") at ../esp-azure/azure-iot-sdk-c/c-utility/src/consolelogger.c:221
#4  0x4018dffe in on_cbs_put_token_complete_callback (context=0x3fff2c6c, operation_result=CBS_OPERATION_RESULT_CBS_ERROR, status_code=0, status_description=0x0) at ../esp-azure/azure-iot-sdk-c/iothub_client/src/iothubtransport_amqp_cbs_auth.c:182
[...]

PR #2210, opened 2022-01-28
shutdown of AMQP - logging causes crash due to status_description is NULL

ericwol-msft:
#2200

 

# Checklist

# Description of the problem

# Description of the solution


danewalton (approved):

ewertons (approved):
I wonder if anything changed. My understanding was that c-utility logging was capable of handling NULL arguments, but I could be wrong. This fix is good though. Thanks!

ericwol-msft:
> I wonder if anything changed. My understanding was that c-utility logging was capable of handling NULL arguments, but I could be wrong. This fix is good though. Thanks!

I already looked. This has been this way since Dan wrote this (NULL pointer for status description). I suspect his CRT lib is not doing null check on sprintf(), which is not required, but is mostly implemented.

PR #2228, opened 2022-02-12
Update reentrancy notes

jspaith:
* Updated threading notes reference documentation with more details RE which API's shouldn't be called from inside an application callback.
* Fixed spelling around "IoTHub" => "IoT Hub" in reference.
* Updated doxygen documentation for LL layer not to call out don't call DoWork in addition to Destroy.  Also change it to more prescriptive - namely telling people just not to do that instead of saying "undefined".  And making a proper `@warning`.


iothub_client/inc/iothub_device_client.h
```diff
@@ -124,8 +124,8 @@ extern "C"
     * @param    userContextCallback             User specified context that will be provided to the
     *                                           callback. This can be @c NULL.
     *
-    *           @b NOTE: The application behavior is undefined if the user calls
-    *           the IoTHubDeviceClient_Destroy() function from within any callback.
+    * @warning: Do not call IoTHubDeviceClient_Destroy() from inside your application's callback.
```
CIPop, line 127: Should this warning be also added to the `IoTHubDeviceClient_Destroy` API itself?

iothub_client/inc/iothub_device_client_ll.h
```diff
@@ -130,8 +130,8 @@ typedef struct IOTHUB_CLIENT_CORE_LL_HANDLE_DATA_TAG* IOTHUB_DEVICE_CLIENT_LL_HA
     * @param    userContextCallback             User specified context that will be provided to the
     *                                           callback. This can be @c NULL.
     *
-    *           @b NOTE: The application behavior is undefined if the user calls
-    *           the IoTHubDeviceClient_LL_Destroy() function from within any callback.
+    * @warning: Do not call IoTHubDeviceClient_LL_Destroy() or IoTHubDeviceClient_LL_DoWork() from inside your application's callback.
```
CIPop, line 133: Same as above: we should have warnings on all API docs that should not be called within callbacks.

CIPop:
@jspaith should we also warn about how long a callback stack can be held?

For example, I think the following are generally a bad idea to have within any of these callbacks:
* Blocking I/O.
* In general, calling functions that sleep or otherwise suspend the thread.
* While Loops performing client requests (e.g. scheduling hundreds of telemetry messages for example).

Instead of saying what not to do, we could instead say that the correct pattern is to perform fast CPU-bound operations exclusively in order to exit the callback fast. Blocking the callback results in stalling the IoT networking stack which may result in disconnects or running out of memory / message queue space.

iothub_client/inc/iothub_device_client.h
```diff
@@ -280,8 +275,7 @@ extern "C"
     * @param    userContextCallback         User specified context that will be provided to the
```
ericwol-msft, line 275: * @brief This API sends a report of the device's properties and their current values. The device reported properties will be sent on the next call to `DoWork()`.

ericwol-msft (approved):

iothub_client/inc/iothub_device_client.h
jspaith: Yup - adding

iothub_client/inc/iothub_device_client.h
jspaith: If we want this it feels like a bigger change -- since almost every LL API would need something about work being queued.  Which in theory we've already covered in `threading_notes.md`.

I'm going to leave as-is for now and we can revisit if we think we need per-API clarity.

CIPop:
> I started to provide more guidance RE like "Spin a thread or if you don't have threads, then have work in outer while() loop" but this quickly got me into trying to build a solution that I don't know about.

I agree. My recommendation for a design would be guided by principles in this article: https://herbsutter.com/2010/07/12/effective-concurrency-prefer-using-active-objects-instead-of-naked-threads/ which, in short, recommends organizing threads around message pumps. This is in-line with your recommendation of either spinning a parallel thread and implementing a queue or enqueuing work then running it in the outer while(). 

That said, the final decision is the app-developer's and will depend on the hardware (CPU, number of cores, etc), memory (how long should the queues be, how many threads) and the real-time aspects of the system (i.e. task deadlines).

PR #2245, opened 2022-02-24
Test stabilization

CIPop:
- Increasing DirectMethods E2E test timeout from 30s to 120s to decrease test faults caused by bad network.
- Adding RNG seed initialization for DPS E2E random back-off timer.

### DIFF
--- a/iothub_client/tests/common_device_method_e2e/iothubclient_common_device_method_e2e.c
+++ b/iothub_client/tests/common_device_method_e2e/iothubclient_common_device_method_e2e.c
@@ -93,7 +93,7 @@ static const char *METHOD_NAME = "MethodName";
 static const int METHOD_RESPONSE_SUCCESS = 200;
 static const int METHOD_RESPONSE_ERROR = 401;
 static const unsigned int TIMEOUT = 60;
-static const unsigned int IOTHUB_CONNECT_TIMEOUT_SEC = 30;
+static const unsigned int IOTHUB_CONNECT_TIMEOUT_SEC = 120;
 
 static void connection_status_callback(IOTHUB_CLIENT_CONNECTION_STATUS result, IOTHUB_CLIENT_CONNECTION_STATUS_REASON reason, void* userContextCallback)
 {
--- a/provisioning_client/tests/prov_x509_client_e2e/prov_x509_client_e2e.c
+++ b/provisioning_client/tests/prov_x509_client_e2e/prov_x509_client_e2e.c
@@ -66,7 +66,10 @@ BEGIN_TEST_SUITE(prov_x509_client_e2e)
         // DPS fails when having multiple enrollments at the same time. 
         // Since we are running these tests on multiple machines with each test taking about 1 second, 
         // we randomize their start time to avoid collisions.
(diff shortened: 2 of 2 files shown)


ewertons (commented):
If you truly believe we would not miss a bug on the server, I will approve it. For now I still think it's better to verify why the tests are timing out instead. They should not take 30 seconds to connect.

CIPop:
> If you truly believe we would not miss a bug on the server, I will approve it. For now I still think it's better to verify why the tests are timing out instead. They should not take 30 seconds to connect.

I finished my investigation on both client and server side. 

The client side logs are:
```
70: Executing test IotHub_Mqtt_Module_Method_Call_With_Null_sas ...
70: -> 20:20:29 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 192 | USERNAME: iotsdk-c-production.azure-devices.net/csdk_e2eDevice_sas_j_please_delete_5d106ddc-8832-4e59-8c23-3264115c84d4/TestModule/?api-version=2020-09-30&DeviceClientType=iothubclient%2f1.9.0%20(native%3b%20WindowsProduct%3a0x00000008%206.2%3b%20x32%3b%20%7bF627FDC1-B3F2-4D41-B2B9-2551D69FAFBE%7d) | PWD: XXXX | CLEAN: 0
70: Assert failed in line 566 Device Not connected Expected: IOTHUB_CLIENT_CONNECTION_AUTHENTICATED, Actual: IOTHUB_CLIENT_CONNECTION_UNAUTHENTICATED
70: Test IotHub_Mqtt_Module_Method_Call_With_Null_sas result = !!! FAILED !!!
 
From <https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=102308&view=logs&j=ba7a5f1e-fec0-5df0-b362-75c9ac4d787c&t=62e6c06f-c16a-50aa-e966-ce9961055360> **
```
The server shows authentication succeeds on the server side at `20:20:30` (within 1 second of the request).
Server logs show that no traffic is detected on this connection up to `2022-02-24T20:20:58.0335416Z` when the connection is closed with `ClosedByClientReason`.


There are three possible reasons for this behavior:

1. This was a half-opened TCP connection (i.e. something is filtering inbound packets for this connection). The client -> server works as expected but server -> client is somehow throttled or blocked. 
2. The CPU usage is very high and the DoWork thread never gets scheduled. I think this is unlikely for 30 seconds but that would depend on the state of the test machine. I've seen this happen in other instances when the machine starts thrashing (lots of pages swapped to disk).
3. We have a bug in our code that somehow blocked the state machine.

In the first 2 cases, increasing the timeout might allow recovery.
If it's the 3rd case, I doubt that increasing this timeout to 120s would do anything: the code should still fail and we can investigate further why that's happening.

ericwol-msft:
I'm fine with increasing this timeout, as this is more an exception than the norm.

ericwol-msft (approved):

ewertons (approved):

CIPop:
Close and reopen for license/cla bot

PR #2261, opened 2022-03-14
Harden longhaul tests against network failures

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/tests/common_longhaul/iothub_client_statistics.c
+++ b/iothub_client/tests/common_longhaul/iothub_client_statistics.c
@@ -12,7 +12,7 @@
 #include "parson.h"
 
 #define INDEFINITE_TIME ((time_t)-1)
-#define SPAN_3_MINUTES_IN_SECONDS (60 * 3)
+#define SPAN_4_MINUTES_IN_SECONDS (60 * 4)
 
 MU_DEFINE_ENUM_STRINGS_WITHOUT_INVALID(TELEMETRY_EVENT_TYPE, TELEMETRY_EVENT_TYPE_VALUES)
 MU_DEFINE_ENUM_STRINGS_WITHOUT_INVALID(C2D_EVENT_TYPE, C2D_EVENT_TYPE_VALUES)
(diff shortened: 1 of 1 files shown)


iothub_client/tests/common_longhaul/iothub_client_statistics.c
```diff
@@ -12,7 +12,7 @@
 #include "parson.h"
 
 #define INDEFINITE_TIME ((time_t)-1)
-#define SPAN_3_MINUTES_IN_SECONDS (60 * 3)
+#define SPAN_4_MINUTES_IN_SECONDS (60 * 4)
```
CIPop, line 15: Nit - CONNECTION_TIMEOUT_SECONDS ?

iothub_client/tests/common_longhaul/iothub_client_statistics.c
```diff
@@ -1460,6 +1460,11 @@ int iothub_client_statistics_get_device_twin_reported_summary(IOTHUB_CLIENT_STAT
 
                 summary->updates_received = summary->updates_received + 1;
             }
+            else if (device_twin_info->send_status_code == 408)  // http status 408 = Request Timeout
+            {
+                summary->updates_sent--;
+                LogInfo("Removing twin reported update id (%d) because of [(408) Request Timeout] network error", (int)device_twin_info->update_id);
```
CIPop, line 1466:
```suggestion
                LogInfo("Removing twin reported update id (%d) because of [(408) Request Timeout] internal server error", (int)device_twin_info->update_id);
```

CIPop: Since the device wasn't disconnected (that's the case handled below) and we got a server response error code, I assume the time-out is the server, not the device->FE connection timeout.

Can we also print other error details such as trackingID (if not already logged) in case the service team wants to investigate?

CIPop (approved):

iothub_client/tests/common_longhaul/iothub_client_statistics.c
ericwol-msft: No, this is coming from the http REST service call. When this happen, I see also network disconnect on the device logs around the same time, so most likely an Azure network infra issue. The service calls does not have a persistent connection, so no reasonable way to track that.

PR #2294, opened 2022-05-10
suppress dump of connection_status_callback

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/tests/global_valgrind_suppression.supp
+++ b/iothub_client/tests/global_valgrind_suppression.supp
@@ -367,3 +367,16 @@
    fun:tickcounter_create
    ...
 }
+{
+   connection_status_callback
+   Helgrind:Race
+   fun:strnlen
+   fun:__vfprintf_internal
+   fun:__vsprintf_internal
+   fun:__sprintf_chk
+   fun:EventData_Create
+   fun:client_create_and_send_d2c
+   fun:e2e_d2c_with_svc_fault_ctrl_with_transport_status
+   fun:RunTests
+   fun:main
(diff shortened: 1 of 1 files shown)


iothub_client/tests/global_valgrind_suppression.supp
```diff
@@ -367,3 +367,16 @@
    fun:tickcounter_create
    ...
 }
+{
+   connection_status_callback
+   Helgrind:Race
+   fun:strnlen
+   fun:__vfprintf_internal
+   fun:__vsprintf_internal
+   fun:__sprintf_chk
```
jspaith, line 376: I'm assuming this is part of the debug check enablements, since we already have a rule like this for the retail version?  If yes, makes sense.

I'd consider though simplifying this rule so it's just EventData_Create/__sprintf_chk but then "..." after.  This way if the underlying CRT changes so that the callstack changes, we'll continue to be protected.  Also if there's different callers on our side we're already protected from the false positives instead of patching up as we go.

Existing place in file I'm thinking of as template.
```
{
    Race conflicting store by thread in EventData_Create/sprintf    
    Helgrind:Race 
    ...
    fun:sprintf
    fun:EventData_Create
}
```

jspaith (approved):

iothub_client/tests/global_valgrind_suppression.supp
ericwol-msft: ugh, found the issue with the other one, there was an extra space before the ...

PR #2310, opened 2022-05-27
remove redundant security file

danewalton:


danewalton:
> 

For all curious, I am deleting the previous security file we had which had the `md` extension capitalized instead of the newer one with lowercase.

ericwol-msft (approved):

PR #2324, opened 2022-06-20
longhaul/E2E framework update

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution


ericwol-msft:
Added UUIDs to Message ID and the CorrelationId
Added additional trace logic
Added conductivity check against hub endpoint
Print AMQP message receiver state

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -287,12 +291,33 @@ static IOTHUBMESSAGE_DISPOSITION_RESULT ReceiveMessageCallback(IOTHUB_MESSAGE_HA
...
 
+static void GenerateUniqueId(const char* prefix, size_t prefix_len, char* buffer, size_t buffer_len)
+{
+    if (prefix_len <= buffer_len)
+    {
+        memcpy(buffer, prefix, prefix_len);
+        buffer_len -= prefix_len;
+        if ((MSG_UNIQUE_ID_STAMP_LEN + 1) <= buffer_len)
+        {
+            buffer[prefix_len - 1] = '.';
```
danewalton, line 302: ?
```suggestion
            buffer[prefix_len] = '.';
```

ericwol-msft: prefix_len is the length of a string literal aka sizeof(""), so the -1 points to the null char and overrides it to a period

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -52,10 +53,11 @@ static TEST_PROTOCOL_TYPE test_protocol_type;
 const char* TEST_EVENT_DATA_FMT = "{\"data\":\"%.24s\",\"id\":\"%d\"}";
 const char* TEST_EVENT_DATA_FMT_SPECIAL_CHAR = "{\"#data\":\"$%.24s\",\";id\":\"*%d\"}";
 
-const char* MSG_ID = "MessageIdForE2E";
-const char* MSG_ID_SPECIAL = "MessageIdForE2E*";
-const char* MSG_CORRELATION_ID = "MessageCorrelationIdForE2E";
-const char* MSG_CORRELATION_ID_SPECIAL = "MessageCorrelationIdFor#E2E";
+#define MSG_UNIQUE_ID_STAMP_LEN 5
+#define MSG_ID "MessageIdForE2E"
+#define MSG_ID_SPECIAL "MessageIdForE2E*"
```
jspaith, line 58: Nit - could we keep the `const char` stuff so we're less pre-processor heavy?

You can below to make the `sizeof` work.
```c
const char* MSG_ID[] = "..."
```

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -126,15 +130,15 @@ static void openCompleteCallback(void* context)
...
+    (void) context;
     if (messagingResult == IOTHUB_MESSAGING_OK)
     {
-        LogInfo("Message has been sent successfully");
+        LogInfo("C2D Message has been sent successfully");
     }
     else
     {
-        LogError("Send failed");
+        LogError("C2D send failed!");
```
jspaith, line 141: Nit - recommend logging out `messagingResult` .

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -639,44 +664,48 @@ D2C_MESSAGE_HANDLE client_create_and_send_d2c(TEST_MESSAGE_CREATION_MECHANISM te
...
         }
         else
         {
-            if (Map_AddOrUpdate(mapHandle, MSG_PROP_KEYS[i], MSG_PROP_VALS[i]) != MAP_OK)
+            msgHandle = NULL;
+            ASSERT_FAIL("Unknown test message creation mechanism specified");
+        }
+        ASSERT_IS_NOT_NULL(msgHandle, "Could not create the D2C message to be sent");
+
+        MAP_HANDLE mapHandle = IoTHubMessage_Properties(msgHandle);
```
jspaith, line 684: Nit - there's API's that make this code a bit easier, specifically `IoTHubMessage_SetProperty` which saves direct manipulation of the list.  A lot of older test code uses this `IoTHubMessage_Properties` semantic but it's officially deprecated.

jspaith: Also in the `IoTHubMessage_SetProperty` below for like "AzIoTHub_FaultOperationType", these end up ASSERT on the whole test.  Is there a reeason we only LogError but don't similarly ASSERT out of test in this path?

jspaith (approved):

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
danewalton: nit would be to have a comment stating that is what's happening here since char* are usually given with length not including the `NULL`.

PR #2359, opened 2022-08-26
Add E2E test to verify AMQP telemetry batching against IoT Hub GWv2

ewertons:
# Checklist

# Description of the problem
The azure-iot-sdk-c e2e tests do not cover AMQP telemetry batching.

# Description of the solution
Add e2e for AMQP telemetry batching.
The test is aimed at the IoT Hub GWv2 only, as there is a bug in GWv1 that prevents proper functionality of batching. 
Note: this PR builds the E2E tests using iothub client LL (as suggested by John).

### DIFF
--- a/iothub_client/tests/common_e2e/iothubclient_common_e2e.c
+++ b/iothub_client/tests/common_e2e/iothubclient_common_e2e.c
@@ -71,7 +71,9 @@ const char* MSG_PROP_VALS[MSG_PROP_COUNT] = { "Val1", "Val2", "Val3" };
 const char* MSG_PROP_KEYS_SPECIAL[MSG_PROP_COUNT] = { "&ey1", "K/y2", "Ke?3"};
 const char* MSG_PROP_VALS_SPECIAL[MSG_PROP_COUNT] = { "=al1", "V@l2", "Va%3" };
 
-static size_t g_iotHubTestId = 0;
+// Due to the way some test messages are composed, please do not set the initial
+// value of g_iotHubTestId to zero (due to potential issues with strlen).
+static size_t g_iotHubTestId = 1;
 IOTHUB_ACCOUNT_INFO_HANDLE g_iothubAcctInfo = NULL;
 E2E_TEST_OPTIONS g_e2e_test_options;
 
@@ -97,6 +99,7 @@ static IOTHUB_MODULE_CLIENT_LL_HANDLE iothub_moduleclient_ll_handle = NULL;
 // `test_message_creation` for creating the telemetry message.
 #define DEFAULT_MESSAGE_SIZE    0
 #define LARGE_MESSAGE_SIZE      64536
+#define LARGE_MESSAGE_COUNT     5
 
 TEST_DEFINE_ENUM_TYPE(IOTHUB_TEST_CLIENT_RESULT, IOTHUB_TEST_CLIENT_RESULT_VALUES);
 TEST_DEFINE_ENUM_TYPE(IOTHUB_CLIENT_RESULT, IOTHUB_CLIENT_RESULT_VALUES);
@@ -114,7 +117,7 @@ typedef struct EXPECTED_SEND_DATA_TAG
     IOTHUB_MESSAGE_HANDLE msgHandle;
 } EXPECTED_SEND_DATA;
 
-#define MAX_SEND_DATA_BUNDLE_ITEM_COUNT 10
+#define MAX_SEND_DATA_BUNDLE_ITEM_COUNT 5
 
 typedef struct SEND_DATA_BUNDLE_TAG
 {
@@ -438,30 +441,19 @@ static EXPECTED_SEND_DATA* EventData_Create(void)
 static EXPECTED_SEND_DATA* EventData_Create_With_Custom_Size(size_t messageSize)
 {
     EXPECTED_SEND_DATA* result;
-    char* message = malloc(sizeof(char) * messageSize);
+    char* message = calloc(messageSize, sizeof(char));
 
-    if (message == NULL)
-    {
-        LogError("Failed to allocate EventData message");
-        result = NULL;
-    }
-    else
-    {
-        char idString[10];
-        size_t idStringLength = sprintf(idString, "%d", (int)g_iotHubTestId++);
+    ASSERT_IS_NOT_NULL(message, "Failed to allocate EventData message");
 
-        (void)memset(message, (int)'a', messageSize - 1);
-        (void)memcpy(message + 1, idString, idStringLength);
-        message[messageSize - 1] = '\0';
+    char idString[10];
+    size_t idStringLength = sprintf(idString, "%d", (int)g_iotHubTestId++);
 
-        result = EventData_Create_With_String(message);
+    (void)memcpy(message, idString, idStringLength);
+    (void)memset(message + idStringLength, 'a', messageSize - idStringLength - 1);
 
-        if (result == NULL)
-        {
-            LogError("Failed allocating EventData");
-            free(message);
-        }
-    }
+    result = EventData_Create_With_String(message);
+
+    ASSERT_IS_NOT_NULL(result, "Failed allocating EventData");
 
     return result;
 }
@@ -880,7 +872,7 @@ D2C_MESSAGE_HANDLE client_create_and_send_d2c(TEST_MESSAGE_CREATION_MECHANISM te
 bool client_wait_for_d2c_confirmations(SEND_DATA_BUNDLE* d2cMessages, IOTHUB_CLIENT_CONFIRMATION_RESULT expectedClientResult)
 {
     time_t beginOperation, nowTime;
-    bool allMessagesSent = false;
+    bool allMessagesDelivered = false;
 
     for (int i = 0; i < d2cMessages->count; i++)
     {
@@ -890,16 +882,16 @@ bool client_wait_for_d2c_confirmations(SEND_DATA_BUNDLE* d2cMessages, IOTHUB_CLI
     beginOperation = time(NULL);
     while (
         (nowTime = time(NULL)),
-        !allMessagesSent && (difftime(nowTime, beginOperation) < MAX_CLOUD_TRAVEL_TIME) // time box
+        !allMessagesDelivered && (difftime(nowTime, beginOperation) < MAX_CLOUD_TRAVEL_TIME) // time box
         )
     {
-        allMessagesSent = true;
+        allMessagesDelivered = true;
 
         for (int i = 0; i < d2cMessages->count; i++)
         {
             if (!client_received_confirmation(d2cMessages->items[i], expectedClientResult))
             {
-                allMessagesSent = false;
+                allMessagesDelivered = false;
                 break;
             }
         }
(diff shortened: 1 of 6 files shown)


iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -949,14 +1214,51 @@ static void send_event_test(IOTHUB_PROVISIONED_DEVICE* deviceToUse, IOTHUB_CLIEN
...
+        // arrange
+        SEND_DATA_BUNDLE d2cMessages = { 0 };
+
+        clear_connection_status_info_flags();
+
+        // Create the IoT Hub Data
+        client_ll_connect_to_hub(deviceToUse, protocol);
+
+        // Send the Events from the client
+        client_create_and_send_d2c_messages(test_message_creation[i], 5, LARGE_MESSAGE_SIZE, &d2cMessages);
```
jspaith, line 1189: `5` is hard-coded magic constant should be declared at top of file.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -342,50 +368,101 @@ static void ReceiveUserContext_Destroy(EXPECTED_RECEIVE_DATA* data)
...
+}
+
+static EXPECTED_SEND_DATA* EventData_Create_With_Custom_Size(size_t messageSize)
+{
+    EXPECTED_SEND_DATA* result;
+    char* message = malloc(sizeof(char) * messageSize);
+
+    if (message == NULL)
+    {
+        LogError("Failed to allocate EventData message");
```
jspaith, line 445: Couldn't this just be an `ASSERT_NOT_NULL(message, "Failed to allocate EventData message")

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -342,50 +368,101 @@ static void ReceiveUserContext_Destroy(EXPECTED_RECEIVE_DATA* data)
...
+
+        (void)memset(message, (int)'a', messageSize - 1);
+        (void)memcpy(message + 1, idString, idStringLength);
+        message[messageSize - 1] = '\0';
+
+        result = EventData_Create_With_String(message);
+
+        if (result == NULL)
+        {
+            LogError("Failed allocating EventData");
```
jspaith, line 461: Similar to above I think this should just be `ASSERT_NOT_NULL(result)`;;  

There's some other places that have been modified I think can follow the ASSERT pattern which I'm not going to call out individually.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -342,50 +368,101 @@ static void ReceiveUserContext_Destroy(EXPECTED_RECEIVE_DATA* data)
...
+    {
+        LogError("Failed to allocate EventData message");
+        result = NULL;
+    }
+    else
+    {
+        char idString[10];
+        size_t idStringLength = sprintf(idString, "%d", (int)g_iotHubTestId++);
+
+        (void)memset(message, (int)'a', messageSize - 1);
```
jspaith, line 453: So if I'm reading this right, it looks like the message will be something like "a<whatever g_iotHubTestId set>aaaaaa... (up to like 64KB, depending on caller)".  Correct?

Having the memcpy of the `idstring` on the line immediately after the `memset` of all the a's while OK since it's overwriting memory we own gave me pause for a bit.  Does the 1st byte *have* to be an a?

So (I have NOT vetted this for all the offby one) IMO below is a bit easier to understand since we're not overwriting ourselves.

```c
memcpy(message, idString, idStrengLength);
memset(message+idStringLength, 'a', messageSize -idStringLength - 1);
```

This is more a nit/consider, just in general and maybe i'm missing something.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -104,6 +114,14 @@ typedef struct EXPECTED_SEND_DATA_TAG
     IOTHUB_MESSAGE_HANDLE msgHandle;
 } EXPECTED_SEND_DATA;
 
+#define MAX_SEND_DATA_BUNDLE_ITEM_COUNT 10
```
jspaith, line 117: Not sure how `MAX_SEND_DATA_BUNDLE_ITEM_COUNT` is tied to other places in program other than defining memory size of `items` right below.  Is this what was supposed to be 5, since it represest max count outstanding?

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -681,56 +843,86 @@ D2C_MESSAGE_HANDLE client_create_and_send_d2c(TEST_MESSAGE_CREATION_MECHANISM te
...
+    client_create_and_send_d2c_messages(test_message_creation, 1, DEFAULT_MESSAGE_SIZE, &d2cMessages);
+    return d2cMessages.items[0];
 }
 
-bool client_wait_for_d2c_confirmation(D2C_MESSAGE_HANDLE d2cMessage, IOTHUB_CLIENT_CONFIRMATION_RESULT expectedClientResult)
+bool client_wait_for_d2c_confirmations(SEND_DATA_BUNDLE* d2cMessages, IOTHUB_CLIENT_CONFIRMATION_RESULT expectedClientResult)
 {
     time_t beginOperation, nowTime;
-    bool ret;
+    bool allMessagesSent = false;
```
jspaith, line 883: Nit - I think this is more like `allMessagesAcked` or something?  Since we could've sent the message technically but the underlying of this (on MQTT specifically) is a PUBACK.  

Though I don't know if that terminology makes sense in AMQP land?  Maybe allMessagesConfirmed to keep parallel with the function name here?

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -144,29 +162,37 @@ static void sendCompleteCallback(void* context, IOTHUB_MESSAGING_RESULT messagin
...
+                ASSERT_FAIL("unable to lock");
+            }
+            else
+            {
+                if (
+                    (strlen(expectedData->expectedString) == size) &&
+                    (memcmp(expectedData->expectedString, data, size) == 0)
+                    )
+                {
+                    expectedData->wasFound = true;
```
jspaith, line 184: A log message(s) don't match could help debugging.

jspaith: Also the size of the message (given its 64KB don't think you should print whole thing out, but the expected and actual sizes can help.)

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -144,29 +162,37 @@ static void sendCompleteCallback(void* context, IOTHUB_MESSAGING_RESULT messagin
 
 static int IoTHubCallback(void* context, const char* data, size_t size)
```
jspaith, line 163: Is this always going to work the way we want?  It seems it depends on all messages being delivered on the same callback?  Since we're intentionally going too large for the batch size - never mind distributed system won't promise the service side gets in same chunk size as was sent - I'm not sure we can make this assumption?

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -681,56 +843,86 @@ D2C_MESSAGE_HANDLE client_create_and_send_d2c(TEST_MESSAGE_CREATION_MECHANISM te
         }
         ASSERT_IS_NOT_NULL(msgHandle, "Could not create the D2C message to be sent");
 
-        for (size_t i = 0; i < MSG_PROP_COUNT; i++)
+        for (size_t j = 0; j < MSG_PROP_COUNT; j++)
         {
             if (g_e2e_test_options.use_special_chars)
             {
-                if (IoTHubMessage_SetProperty(msgHandle, MSG_PROP_KEYS_SPECIAL[i], MSG_PROP_VALS_SPECIAL[i]) != IOTHUB_MESSAGE_OK)
+                if (IoTHubMessage_SetProperty(msgHandle, MSG_PROP_KEYS_SPECIAL[j], MSG_PROP_VALS_SPECIAL[j]) != IOTHUB_MESSAGE_OK)
```
jspaith, line 850: I get this is pre-existing condition, but... shouldn't a `IoTHubMessage_SetProperty` cause an ASSERT_FAIL.  It suggets something is really wrong if something this basic fails and also later in test we'll see if these are set and fail there if not.  I'd rather stop test at root problem.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -949,14 +1168,51 @@ static void send_event_test(IOTHUB_PROVISIONED_DEVICE* deviceToUse, IOTHUB_CLIEN
         // cleanup
         destroy_d2c_message_handle(d2cMessage);
     }
+}
+
+static void send_batch_event_test(IOTHUB_PROVISIONED_DEVICE* deviceToUse, IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol)
```
jspaith, line 1173: This is different enough from rest of tests that I think some basic comments here are worthwhile.  Like we need to use LL to force the batching to occur since we lose control on convenience layer as at least one hint to future maintainers.

danewalton (approved):

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Defined LARGE_MESSAGE_COUNT

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Done.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Done.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: I will focus on the goal of the changes proposed for AMQP max message size.
Here is a github issue filed to track this request: https://github.com/Azure/azure-iot-sdk-c/issues/2360

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: The way this implementation works is that this `IoTHubCallback` will be called once for each message received by the **test service client**.
So I am not sure what assumption you are bringing up. I'll keep this open until we clarify the ask here.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: That's a good suggestion. Done.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Makes sense. Done.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
jspaith: Apologies - I completely misread how this code and the callback generally works.  This is indeed fine.  And it's a reading error on my part - given all the moving pieces I think this is the best (or even only) way to solve this as you've implemented.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
```diff
@@ -438,30 +441,19 @@ static EXPECTED_SEND_DATA* EventData_Create(void)
...
-    else
-    {
-        char idString[10];
-        size_t idStringLength = sprintf(idString, "%d", (int)g_iotHubTestId++);
+    ASSERT_IS_NOT_NULL(message, "Failed to allocate EventData message");
 
-        (void)memset(message, (int)'a', messageSize - 1);
-        (void)memcpy(message + 1, idString, idStringLength);
-        message[messageSize - 1] = '\0';
+    char idString[10];
```
jspaith, line 448: Nit - couldn't you `sprintf` into the `message` buffer directly instead of needing a separate `idString`?

jspaith (approved):

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Thanks :)
I'm glad we double-checked it.

iothub_client/tests/common_e2e/iothubclient_common_e2e.c
ewertons: Taken.

PR #2393, opened 2022-10-21
Allow empty file blob uploads

ericwol-msft:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/iothub_client/src/blob.c
+++ b/iothub_client/src/blob.c
@@ -13,6 +13,9 @@
 #include "azure_c_shared_utility/azure_base64.h"
 #include "azure_c_shared_utility/shared_util_options.h"
 
+#define HTTP_STATUS_CODE_OK                 200
+#define IS_HTTP_STATUS_CODE_SUCCESS(x)      (x < 300)
+
 static const char blockListXmlBegin[]  = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\r\n<BlockList>";
 static const char blockListXmlEnd[] = "</BlockList>";
 static const char blockListUriMarker[] = "&comp=blocklist";
@@ -105,7 +108,7 @@ BLOB_RESULT Blob_UploadBlock(
                                 LogError("unable to HTTPAPIEX_ExecuteRequest");
                                 result = BLOB_HTTP_ERROR;
                             }
-                            else if (*httpStatus >= 300)
+                            else if (!IS_HTTP_STATUS_CODE_SUCCESS(*httpStatus))
                             {
                                 LogError("HTTP status from storage does not indicate success (%d)", (int)*httpStatus);
                                 result = BLOB_OK;
@@ -152,6 +155,7 @@ static BLOB_RESULT InvokeUserCallbackAndSendBlobs(HTTPAPIEX_HANDLE httpApiExHand
         {
             uploadOneMoreBlock = 0;
             result = BLOB_OK;
+            *httpStatus = HTTP_STATUS_CODE_OK;
         }
         else
         {
@@ -195,7 +199,7 @@ static BLOB_RESULT InvokeUserCallbackAndSendBlobs(HTTPAPIEX_HANDLE httpApiExHand
                     LogError("unable to Blob_UploadBlock. Returned value=%d", result);
                     isError = 1;
                 }
-                else if (*httpStatus >= 300)
+                else if (!IS_HTTP_STATUS_CODE_SUCCESS(*httpStatus))
                 {
                     LogError("unable to Blob_UploadBlock. Returned httpStatus=%u", (unsigned int)*httpStatus);
                     isError = 1;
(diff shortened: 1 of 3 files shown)


iothub_client/src/blob.c
```diff
@@ -13,6 +13,9 @@
 #include "azure_c_shared_utility/azure_base64.h"
 #include "azure_c_shared_utility/shared_util_options.h"
 
+#define HTTP_STATUS_CODE_OK                 200
+#define IS_HTTP_STATUS_CODE_SUCCESS(x)      (x < 300)
```
jspaith, line 17: Nit should be `((x)) < 300` to avoid weird preprocessor expansion rules.

iothub_client/tests/blob_ut/blob_ut.c
```diff
@@ -1252,4 +1254,24 @@ TEST_FUNCTION(Blob_UploadMultipleBlocksFromSasUri_returns_BLOB_ABORTED_when_call
...
+    fakeContext.blockSize = 1;
+    fakeContext.blocksCount = 0;
+    fakeContext.fakeData = NULL;
+    fakeContext.abortOnBlockNumber = 5;
+
+    ///act
+    BLOB_RESULT result = Blob_UploadMultipleBlocksFromSasUri("https://h.h/something?a=b", FileUpload_GetFakeData_Callback, &fakeContext, &httpResponse, testValidBufferHandle, NULL, NULL, NULL);
+
+    ///assert
+    ASSERT_ARE_EQUAL(BLOB_RESULT, BLOB_OK, result);
```
jspaith, line 1271: I get this is a pre-existing condition in the other tests, but... couldn't we 

```c
ASSERT_ARE_EQUAL(int, 200, httpResponse);
```

Also I'd never noticed before but `httpResponse` being a global seems weird should be just on stack per each fcn?  Maybe more concerning to keep it non-global if(?) you start ASSERT on its value in test.

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
```diff
@@ -353,32 +355,34 @@ static void uploadToBlobGetData(IOTHUB_CLIENT_FILE_UPLOAD_RESULT result, unsigne
...
     ASSERT_IS_NOT_NULL(deviceToUse);
-
-    IOTHUB_CLIENT_HANDLE iotHubClientHandle = IoTHubClient_CreateFromConnectionString(deviceToUse->connectionString, protocol);
-    ASSERT_IS_NOT_NULL(iotHubClientHandle, "Could not invoke IoTHubClient_CreateFromConnectionString");
-
-    uploadToBlobCauseAbort = causeAbort;
-    g_uploadToBlobStatus = UPLOADTOBLOB_CALLBACK_PENDING;
-    uploadBlobNumber = 0;
-    if (useExMethod)
+    if (deviceToUse != NULL)
```
jspaith, line 363: Do we need the `if (deviceToUse)` check?  Right above there's an `ASSERT_IS_NOT_NULL(deviceToUse)` so if this is NULL the test should immediately exit.

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
```diff
@@ -251,32 +251,34 @@ static void e2e_uploadtoblob_test(IOTHUB_CLIENT_TRANSPORT_PROVIDER protocol, IOT
         deviceToUse = IoTHubAccount_GetSASDevice(g_iothubAcctInfo);
     }
     ASSERT_IS_NOT_NULL(deviceToUse);
-
-    IOTHUB_CLIENT_HANDLE iotHubClientHandle = IoTHubClient_CreateFromConnectionString(deviceToUse->connectionString, protocol);
-    ASSERT_IS_NOT_NULL(iotHubClientHandle, "Could not invoke IoTHubClient_CreateFromConnectionString");
-
-    if (accountAuthMethod == IOTHUB_ACCOUNT_AUTH_X509)
+    if (deviceToUse != NULL)
```
jspaith, line 254: Do we need the if (deviceToUse) check? Right above there's an ASSERT_IS_NOT_NULL(deviceToUse) so if this is NULL the test should immediately exit.

jspaith (approved):

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
ericwol-msft: Yes, this is flagged by SDL check. I fix them as I see them.

iothub_client/tests/iothubclient_uploadtoblob_e2e/iothubclient_uploadtoblob_e2e.c
ericwol-msft: Yes, this is flagged by SDL. It helps keep the noise down.

iothub_client/src/blob.c
```diff
@@ -13,6 +13,9 @@
 #include "azure_c_shared_utility/azure_base64.h"
 #include "azure_c_shared_utility/shared_util_options.h"
 
+#define HTTP_STATUS_CODE_OK                 200
+#define IS_HTTP_STATUS_CODE_SUCCESS(x)      ((x) < 300)
```
CIPop, line 17: Nit: should also be >100

CIPop (approved):

PR #2394, opened 2022-10-31
Add an option to define a custom target network interface for e2e tests

ewertons:
# Checklist

# Description of the problem
Add an option to define a custom target network interface for e2e tests that explicitly choose the network interface to be used (not all e2e tests do so).
The tests are hardcoded to use the network interface "eth0", and it works ok on older distros and docker, but not on more recent linux systems.

# Description of the solution
Add an option to set the following environment variable to define which network interface the e2e tests should use:  IOTHUB_E2E_TARGET_NETWORK_INTERFACE

Example:
```
export IOTHUB_E2E_TARGET_NETWORK_INTERFACE=wlp4s0
```

### DIFF
--- a/testtools/iothub_test/src/iothubtest.c
+++ b/testtools/iothub_test/src/iothubtest.c
@@ -165,6 +165,8 @@ const char* AMQP_ADDRESS_PATH_FMT = "/devices/%s/messages/deviceBound";
 const char* AMQP_SEND_TARGET_ADDRESS_FMT = "amqps://%s/messages/deviceBound";
 const char* AMQP_SEND_AUTHCID_FMT = "iothubowner@sas.root.%s";
 
+const char* DEFAULT_TARGET_NETWORK_INTERFACE = "eth0";
+
 #define THREAD_CONTINUE             0
 #define THREAD_END                  1
 #define MAX_DRAIN_TIME              1000.0
@@ -2004,6 +2006,34 @@ IOTHUB_TEST_CLIENT_RESULT IoTHubTest_SendMessage(IOTHUB_TEST_HANDLE devhubHandle
             xio_destroy(tls_io);
             saslmechanism_destroy(sasl_mechanism_handle);
         }
+    }
+
+    return result;
+}
+
+const char* IoTHubTest_GetTargetNetworkInterface(IOTHUB_TEST_HANDLE devhubHandle)
+{
+    const char* result = NULL;
+    static char networkInterface[32];
+
+    if (devhubHandle == NULL)
+    {
+        LogError("Invalid parameter: devhubHandle is NULL", devhubHandle);
+    }
+    else
+    {
+        char* envVarValue = getenv("IOTHUB_TEST_TARGET_NETWORK_INTERFACE");
+
+        if (envVarValue != NULL)
+        {
+            (void)strcpy(networkInterface, envVarValue);
+        }
+        else
+        {
+            (void)strcpy(networkInterface, DEFAULT_TARGET_NETWORK_INTERFACE);
+        }
+
+        result = networkInterface;
     }
 
     return result;
--- a/testtools/iothub_test/src/iothubtest.c
+++ b/testtools/iothub_test/src/iothubtest.c
@@ -2018,7 +2018,7 @@ const char* IoTHubTest_GetTargetNetworkInterface(IOTHUB_TEST_HANDLE devhubHandle
 
     if (devhubHandle == NULL)
     {
-        LogError("Invalid parameter: devhubHandle is NULL", devhubHandle);
+        LogError("Invalid parameter: devhubHandle is NULL");
     }
     else
     {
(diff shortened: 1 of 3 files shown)


testtools/iothub_test/src/iothubtest.c
```diff
@@ -2008,3 +2010,23 @@ IOTHUB_TEST_CLIENT_RESULT IoTHubTest_SendMessage(IOTHUB_TEST_HANDLE devhubHandle
 
     return result;
 }
+
+const char* IoTHubTest_GetTargetNetworkInterface()
+{
+    const char* result = NULL;
+    static char networkInterface[32];
+    char* envVarValue = getenv("IOTHUB_TEST_TARGET_NETWORK_INTERFACE");
```
danewalton, line 2018: nit: make the string a macro like other values in this file?

danewalton (approved):

testtools/iothub_test/src/iothubtest.c
ewertons: Sounds like a good idea :)

testtools/iothub_test/src/iothubtest.c
```diff
@@ -2008,3 +2010,23 @@ IOTHUB_TEST_CLIENT_RESULT IoTHubTest_SendMessage(IOTHUB_TEST_HANDLE devhubHandle
 
     return result;
 }
+
+const char* IoTHubTest_GetTargetNetworkInterface()
+{
+    const char* result = NULL;
+    static char networkInterface[32];
```
jspaith, line 2017: Nit - 32 is magic #.

jspaith, line 2017: Actually why does this need to be a static char in 1st place?  Can't we just return value from getenv() or the DEFAULT_TARGET_NETWORK_INTERFACE to the caller without strcpy().

Or if we need this, I'd prefer caller to have to pass the buffer and length in.  I get the odds of this on being used on multiple threads seem unlikely but if that ever happens (since the function otherwise looks thread safe, but it's not with a static) it could cause issues.

jspaith (changes requested):

testtools/iothub_test/src/iothubtest.c
ewertons: The reason I didn't do that was to keep the same code design used for the other functions that ready env vars. 
I wanted to keep this simple, but if I rather break with the design I'll just return the value read from getenv

jspaith: Interesting!  I hadn't looked at rest of file and just looking now.

So `IoTHubAccount_GetEventhubListenName` looks like we did this since there's some "post-processing" (which is fancier than saying sprint:)) and relying on nature of tests being single threaded.

But `IoTHubAccount_GetEventhubConsumerGroup` which the new function is closest to, also could just return the getenv result directly.

I'd personally vote that we return the getenv result (or the constant when env not set) directly from this function (and potentially change `IoTHubAccount_GetEventhubConsumerGroup` but I feel less strongly as that's legacy test code). 
 And otherwise leave other stuff alone.

testtools/iothub_test/src/iothubtest.c
ericwol-msft, line 2018:
```suggestion
    char* envNetworkInterfaceVal = getenv("IOTHUB_TEST_TARGET_NETWORK_INTERFACE");
```

testtools/iothub_test/src/iothubtest.c
ericwol-msft: +1 for John

ericwol-msft (approved):

testtools/iothub_test/src/iothubtest.c
CIPop: > Can't we just return value from getenv() or the DEFAULT_TARGET_NETWORK_INTERFACE to the caller without strcpy().

According to the manual anything returned from `getenv()` must be copied or a future call to `getenv()` might change the previous return value.

> The implementation of getenv() is not required to be reentrant. The string pointed to by the return value of getenv() may be statically allocated, and ~~can be modified by a subsequent call to getenv()__, [putenv]

CIPop (approved):
Thanks @ewertons ! 

This is required ever since the [predictable interfaces names](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-understanding_the_predictable_network_interface_device_names#:~:text=Understanding%20the%20Predictable%20Network%20Interface%20Device%20Names%201,3%20ww%20for%20wireless%20wide%20area%20network%20%28WWAN%29.) were added to Linux.

Usually a problem running on a real Linux box - from what I see, VMs usually do have `eth0` which I see remained the default.

jspaith (approved):
Per offline discussion RE `getenv` behavior potentially needing a strcpy on some distros which is more a concern than my hypothetical multi-thread (which is very unlikely given we control both this & caller and this is test util code).

testtools/iothub_test/src/iothubtest.c
ewertons: Done.

testtools/iothub_test/src/iothubtest.c
ewertons: Is it really? :)

ewertons: Ok. Changed to a define.

testtools/iothub_test/src/iothubtest.c
ewertons: Ok. Done.

Issue #2399, opened 2022-11-11
Incorrect handle passed to Unsubscribe_InputQueue


Cavaler:
In iothub_client/src/iothub_client_core_ll.c, in function remove_event_unsubscribe_if_needed(), incorrect handle passed in call to IoTHubTransport_Unsubscribe_InputQueue(), resulting in SIGSEGV.

It should be transport handle, not client handle:
`handleData->IoTHubTransport_Unsubscribe_InputQueue(handleData->transportHandle);`

Cavaler:
Patch for release 1.9.1:
```
--- iothub_client/src/iothub_client_core_ll.c.orig	2022-11-11 20:50:54.994743800 +0200
+++ iothub_client/src/iothub_client_core_ll.c	2022-11-11 21:02:53.297248700 +0200
@@ -2919,7 +2919,7 @@
             {
                 if (singlylinkedlist_get_head_item(handleData->event_callbacks) == NULL)
                 {
-                    handleData->IoTHubTransport_Unsubscribe_InputQueue(handleData);
+                    handleData->IoTHubTransport_Unsubscribe_InputQueue(handleData->transportHandle);
                 }
                 result = IOTHUB_CLIENT_OK;
             }

```

PR #2403, opened 2022-11-14
update to correct handle

danewalton:
Related to #2399

image

The function signature for this function is an `IOTHUB_DEVICE_HANDLE` not an `IOTHUB_CLIENT_CORE_LL_HANDLE_DATA`


ewertons (approved):
Yes, correct. That shows that most likely this code path is not seeing the daylight.

As shown below, this is the signature of the function:
```c
   typedef void(*pfIoTHubTransport_Unsubscribe_InputQueue)(IOTHUB_DEVICE_HANDLE handle);
```
https://github.com/Azure/azure-iot-sdk-c/blob/main/iothub_client/inc/internal/iothub_transport_ll_private.h#L86

So the correct would be passing the deviceHandle, not handleData.

PR #2439, opened 2023-02-01
Propagate file upload timeout

vaavva:
# Checklist

# Description of the problem

Setting the option 'OPTION_BLOB_UPLOAD_TIMEOUT_SECS' only persists for step 1 and 3 of the file upload (telling IoTHub there's an upload and telling IoTHub the upload finished)

# Description of the solution
 
This PR makes sure the timeout gets set in step 2 as well (the upload of the file)


iothub_client/inc/internal/blob.h
```diff
@@ -57,10 +57,11 @@ MU_DEFINE_ENUM_WITHOUT_INVALID(BLOB_RESULT, BLOB_RESULT_VALUES)
 * @param  certificates      A null terminated string containing CA certificates to be used
 * @param    proxyOptions    A structure that contains optional web proxy information
 * @param  networkInterface    An optional null terminated string containing the network interface
+* @param  timeout           An optional timeout value for sending and receiving messages
 *
 * @return    A @c BLOB_RESULT. BLOB_OK means the blob has been uploaded successfully. Any other value indicates an error
 */
-MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadMultipleBlocksFromSasUri, const char*, SASURI, IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_CALLBACK_EX, getDataCallbackEx, void*, context, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates, HTTP_PROXY_OPTIONS*, proxyOptions, const char*, networkInterface)
+MOCKABLE_FUNCTION(, BLOB_RESULT, Blob_UploadMultipleBlocksFromSasUri, const char*, SASURI, IOTHUB_CLIENT_FILE_UPLOAD_GET_DATA_CALLBACK_EX, getDataCallbackEx, void*, context, unsigned int*, httpStatus, BUFFER_HANDLE, httpResponse, const char*, certificates, HTTP_PROXY_OPTIONS*, proxyOptions, const char*, networkInterface, const long*, timeout)
```
ewertons, line 64: Suggestion: rename `timeout` to `timeoutÌnMilliseconds`

iothub_client/src/iothub_client_ll_uploadtoblob.c
```diff
@@ -745,7 +745,8 @@ IOTHUB_CLIENT_RESULT IoTHubClient_LL_UploadMultipleBlocksToBlob_Impl(IOTHUB_CLIE
                                         }
                                         else
                                         {
-                                            BLOB_RESULT uploadMultipleBlocksResult = Blob_UploadMultipleBlocksFromSasUri(STRING_c_str(sasUri), getDataCallbackEx, context, &httpResponse, responseToIoTHub, upload_data->certificates, &(upload_data->http_proxy_options), upload_data->networkInterface);
+                                            long timeout = ((long)upload_data->blob_upload_timeout_secs * 1000);
```
ewertons, line 748: Suggestion: rename `timeout` to `timeoutÌnMilliseconds`

iothub_client/inc/internal/blob.h
ewertons: Why not pass const long instead of the pointer, and check for 0 within Blob_UploadMultipleBlocksFromSasUri ?

ewertons (approved):

iothub_client/inc/internal/blob.h
vaavva: I was thinking the parameter had to be nullable because of the other parameters (and since it's optional), but it doesn't look like it does since this is an internal function - I'll make this change

vaavva: I couldn't pass it as a const or the unit tests got mad (`Could not copy type, type const size_t not registered.`) - if there's a way around this, I'm happy to change it back to a const though!

danewalton: I think you might be able to get around it with a registered alias type:

https://github.com/Azure/azure-c-shared-utility/blob/718ec00a1df4917f1ee8aa528bdfa37e56bd8e92/tests/tlsio_options_ut/tlsio_options_ut.c#L95

iothub_client/src/iothub_client_ll_uploadtoblob.c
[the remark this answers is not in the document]
vaavva: the 1000 is the conversion from seconds to milliseconds! I can pull it into a variable instead of a magic number though

iothub_client/inc/internal/blob.h
vaavva: That worked, thanks!

iothub_client/src/iothub_client_ll_uploadtoblob.c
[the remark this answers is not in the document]
vaavva: It looks like we usually use 1000 as a magic number for ms conversions, so I'm going to leave this as is
https://github.com/Azure/azure-iot-sdk-c/search?q=1000

iothub_client/src/blob.c
```diff
@@ -328,6 +328,11 @@ BLOB_RESULT Blob_UploadMultipleBlocksFromSasUri(const char* SASURI, IOTHUB_CLIEN
                     LogError("unable to create a HTTPAPIEX_HANDLE");
                     result = BLOB_ERROR;
                 }
+                else if ((timeoutInMilliseconds != 0) && (HTTPAPIEX_SetOption(httpApiExHandle, "timeout", &timeoutInMilliseconds) == HTTPAPIEX_ERROR))
```
jspaith, line 331: This should be `OPTION_HTTP_TIMEOUT` from `shared_util_options.h`.  The `TrustedCerts" pre-existing online 336 ideally could get piggy back fixed here to use `OPTION_TRUSTED_CERT`.

iothub_client/tests/blob_ut/blob_ut.c
```diff
@@ -343,7 +343,7 @@ TEST_FUNCTION(Blob_UploadMultipleBlocksFromSasUri_with_NULL_SasUri_fails)
     unsigned int httpResponse = HTTP_OK;
 
     ///act
-    BLOB_RESULT result = Blob_UploadMultipleBlocksFromSasUri(NULL, FileUpload_GetData_Callback, &context, &httpResponse, testValidBufferHandle, NULL, NULL, NULL);
+    BLOB_RESULT result = Blob_UploadMultipleBlocksFromSasUri(NULL, FileUpload_GetData_Callback, &context, &httpResponse, testValidBufferHandle, NULL, NULL, NULL, 0);
```
jspaith, line 346: Please add a test where we're passing a non-zero timeout here and make sure that the HTTPAPI_SetOption() is being executed.  See the blocks in `Blob_UploadMultipleBlocksFromSasUri_various_sizes_happy_path_Impl` like say below to give a flavor of how this works with optional certs.

```c
        if (certificate != NULL)
        {
            STRICT_EXPECTED_CALL(HTTPAPIEX_SetOption(IGNORED_PTR_ARG, "TrustedCerts", IGNORED_PTR_ARG));
        }
```

jspaith (approved):

iothub_client/src/blob.c
vaavva: Fixed!

iothub_client/tests/blob_ut/blob_ut.c
vaavva: Added!

CIPop:
This is failing because of a CI bug introduced by a new policy.

PR #2444, opened 2023-02-07
Remove public-preview.

JoseSanchez7:
# Description of the problem
Public-preview was removed from vcpkg and the documentation did not reflect the changes.

# Description of the solution
Updated the documentation to reflect the changes that were made.

### DIFF
--- a/doc/setting_up_vcpkg.md
+++ b/doc/setting_up_vcpkg.md
@@ -21,17 +21,6 @@ The `use-prov-client` feature uses the syntax: `vcpkg install azure-iot-sdk-c[us
     -Duse_prov_client=ON
 ```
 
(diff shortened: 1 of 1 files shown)


doc/setting_up_vcpkg.md
```diff
@@ -21,17 +21,6 @@ The `use-prov-client` feature uses the syntax: `vcpkg install azure-iot-sdk-c[us
     -Duse_prov_client=ON
```
vaavva, line 21: should `use-prov-client` be removed from the docs as well?

JoseSanchez7: Yes, good catch, thank you!

vaavva (approved):

ewertons (approved):
I checked and there are no other files in the repo that would also need to be cleaned.

PR #2507, opened 2023-07-24
Install iothub_client internal headers ALSO under azureiot/internal

ewertons:
# Checklist

# Description of the problem

ARM and MIPS32 crosscompilation builds are failing due to changes in upload-to-blob header includes throughout azure-iot-sdk-c.

# Description of the solution
 
- Install iothub_client internal headers ALSO under "azureiot/internal", since many #include use the "internal" relative path.
- Add "include/azureiot" to include_directories of ARM and MIPS32 toolchains, since that's where the headers is installed by default.

Note: These tasks (cross-comp of ARM and MIPS32) apparently ran when the PR for upload to blob changes was validated for merging, but they test against main in https://github.com/azure/azure-iot-sdk-c, not against the PR code itself.

### DIFF
--- a/iothub_client/CMakeLists.txt
+++ b/iothub_client/CMakeLists.txt
@@ -27,6 +27,14 @@ set(iothub_client_c_files
     ${CMAKE_CURRENT_LIST_DIR}/src/version.c
 )
 
+set(iothub_client_internal_h_files
+    ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothub_client_diagnostic.h
+    ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothub_internal_consts.h
+    ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothub_client_private.h
+    ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothub_message_private.h
+    ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothubtransport.h
+)
+
 set(iothub_client_h_files
     ${CMAKE_CURRENT_LIST_DIR}/inc/iothub.h
     ${CMAKE_CURRENT_LIST_DIR}/inc/iothub_client_core.h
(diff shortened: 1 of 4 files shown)


iothub_client/CMakeLists.txt
```diff
@@ -76,23 +80,28 @@ if (use_edge_modules)
...
+    set(iothub_client_internal_h_files
+        ${iothub_client_internal_h_files}
         ${CMAKE_CURRENT_LIST_DIR}/inc/internal/iothub_client_hsm_ll.h
     )
 endif()
 #end back compat only
 
 set(iothub_client_h_install_files
     ${iothub_client_h_files}
+    ${iothub_client_internal_h_files}
```
vaavva, line 100: Isn't `iothub_client_internal_h_files` included in `iothub_client_h_files`?

vaavva (approved):

samples/dockerbuilds/ARM/Dockerfile
```diff
@@ -91,6 +94,7 @@ RUN echo "SET(CURL_LIBRARIES ${TOOLCHAIN_PREFIX}/lib/libcurl.so)" >> toolchain.c
 RUN echo "SET(ENV{LDFLAGS} -L${TOOLCHAIN_PREFIX}/lib)" >> toolchain.cmake
 RUN echo "SET(set_trusted_cert_in_samples true CACHE BOOL \"Force use of TrustedCerts option\" FORCE)" >> toolchain.cmake
 RUN echo "include_directories(${TOOLCHAIN_PREFIX}/include)" >> toolchain.cmake
+RUN echo "include_directories(${TOOLCHAIN_PREFIX}/include/azureiot)" >> toolchain.cmake
```
CIPop, line 97: I don't understand: how was this working in the past? If we change the install directory, wouldn't all samples require something like `#include <azureiot/...>` before and now without azureiot?

CIPop (commented):
If the files were installing in `include/` before and will now install in `azureiot/include`, that is a breaking API change as existing applications will either need to change their INCLUDE definitions (as this PR does), or their sources, to prepend `azureiot/` to each `#include ` statement.

iothub_client/CMakeLists.txt
ewertons: yes, not to break the existing customers.

samples/dockerbuilds/ARM/Dockerfile
ewertons: That's something I was not able to pinpoint.

ewertons:
This PR will be reworked.

PR #2509, opened 2023-07-25
Fix cross-comp CI pipeline (ARM & MIPS32) to build against target branch instead of main

ewertons:
# Checklist

# Description of the problem

The ARM and the MIPS32 cross-compilation jobs in .vsts-ci.yml will always target the main branch, ignoring the branch a given PR is based on and letting potential issues sip through. 

# Description of the solution
 
This change fixes the ARM and MIPS32 jobs so they run against whatever branch the pipeline is running against.

### DIFF
--- a/samples/dockerbuilds/ARM/Dockerfile
+++ b/samples/dockerbuilds/ARM/Dockerfile
@@ -1,5 +1,6 @@
 FROM ubuntu:latest
 ARG DEBIAN_FRONTEND=noninteractive
+ARG TARGET_BRANCH=main
 
 #########################################
 # start from home directory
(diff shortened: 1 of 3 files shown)


samples/dockerbuilds/ARM/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM ubuntu:latest
 ARG DEBIAN_FRONTEND=noninteractive
+ARG TARGET_BRANCH=main
```
vaavva, line 3: Should this be the source branch?

vaavva (approved):

samples/dockerbuilds/ARM/Dockerfile
CIPop: Looks like --build-arg will override: https://docs.docker.com/build/guide/build-args/

CIPop (approved):
LGTM
Because of a known breaking issue the build is _expected_ to fail with an include location failure. 

The build should still show the docker images being built correctly, out of the PR branch (instead of `main`). If this is true, @ewertons and I agreed to force-push the CI change.

samples/dockerbuilds/ARM/Dockerfile
ewertons: Yes. This is the default value, "main". Its value get set by the pipeline job in .vsts-ci.yml

PR #2512, opened 2023-07-27
Add LTS Table entry for LTS_07_2023_Ref01

RLeclair:


readme.md
```diff
@@ -276,10 +276,10 @@ Below is a table showing the mapping of the LTS branches to the packages release
 
   | Package | GitHub Branch | LTS Tag | LTS Start Date | Maintenance End Date |
   | :-----: | :-----------: | :-----: | :------------: | :------------------: |
+  | vcpkg: 2023-07-27 | lts_07_2023 | LTS_07_2023_Ref01 | 2023-07-27 | 2024-07-27 |
   | vcpkg: 2023-04-20 | lts_01_2023 | LTS_01_2023_Ref02 | 2023-04-20 | 2024-04-20 |
   | vcpkg: 2023-01-13 | lts_01_2023 | LTS_01_2023_Ref01 | 2023-01-13 | 2024-01-13 |
   | vcpkg: 2022-09-15 | lts_07_2022 | LTS_07_2022_Ref02 | 2022-09-15 | 2023-09-15 |
-  | vcpkg: 2022-07-18 | lts_07_2022 | LTS_07_2022_Ref01 | 2022-07-18 | 2023-07-18 |
```
CIPop, line 282: I think it is useful to have a history of all LTSs and when the support ended.

readme.md
```diff
@@ -276,10 +276,10 @@ Below is a table showing the mapping of the LTS branches to the packages release
 
   | Package | GitHub Branch | LTS Tag | LTS Start Date | Maintenance End Date |
   | :-----: | :-----------: | :-----: | :------------: | :------------------: |
+  | vcpkg: 2023-07-27 | lts_07_2023 | LTS_07_2023_Ref01 | 2023-07-27 | 2024-07-27 |
```
CIPop, line 279: I think the date should be changed last: it's premature to state the release will happen today. Instead, I would first create the release, then check-in the change.
Also, changes should always be made in main then backported. The LTS branch should not contain additional check-ins to those in main unless there is a revision.

CIPop (changes requested):
There should be no additional commits to LTS vs main. LTS should be _identical_ to main to a particular point in time.

Further additions to LTS (revisions) will make the branch diverge at later points, only if necessary.

ewertons (approved):

readme.md
vaavva: Our regular procedure is to remove any entries in this table once they pass their end date.

CIPop (commented):
Removing my request for changes - looks like this is how LTSs were done for a while.

CIPop (approved):

PR #2513, opened 2023-07-28
Bump version from 1.11.0 to 1.12.0. Add LTS Table entry for LTS_07_20…

RLeclair:


ewertons (approved):

iothub_client/inc/iothub_client_version.h
```diff
@@ -9,7 +9,7 @@
 #define IOTHUB_CLIENT_VERSION_H
 
 /** @brief Macro representing current IoT Hub Client SDK version. */
-#define IOTHUB_SDK_VERSION "1.11.0"
+#define IOTHUB_SDK_VERSION "1.12.0"
```
CIPop, line 12: We use semver: I assume we increased to 1.12 because of the new public APIs in blob upload? /cc @ewertons

provisioning_client/inc/azure_prov_client/prov_client_const.h
```diff
@@ -4,7 +4,7 @@
 #ifndef PROV_CLIENT_CONST_H
 #define PROV_CLIENT_CONST_H
 
-#define PROV_DEVICE_CLIENT_VERSION "1.11.0"
+#define PROV_DEVICE_CLIENT_VERSION "1.12.0"
```
CIPop, line 7: Are there public APIs modified in provisioning? Should this be 1.11.1 instead?

CIPop (commented):
@RLeclair it would be useful to also list the reason of the minor version increase.
AFAIK, we're using https://semver.org/.

Issue #2514, opened 2023-07-31
Duplicate messages seen in Azure IoT Hub


acode-x:
Hi,

I am using `IoTHubDeviceClient_SendEventAsync` to push messages to Azure IoT Hub every 1min.
```C
IoTHubMessage_CreateFromString(content.c_str());
(void)IoTHubMessage_SetContentTypeSystemProperty(static_cast<MY_HANDLE>(msg), "application%2Fjson");
(void)IoTHubMessage_SetContentEncodingSystemProperty(static_cast<MY_HANDLE>(msg), "utf-8");
```

We notice that some messages are sent multiple times, and there are duplicate entries created in Azure Hub for the same message.
Also, it reconnects quite often (not sure if bandwidth is a problem here as CONNACK and SUBACK is quickly received).
This scenario occurs multiple times a day for quite some time.
Is seeing the duplicate messages in Azure Hub expected behaviour when we use QoS 1? If so, could you guide me how to use QoS 2?

Attaching logs for reference.
```bash
-> 01:30:07 CONNECT | VER: 4 | KEEPALIVE: 240 | FLAGS: 128 | USERNAME: aaa/temperature/?api-version=2020-09-30&DeviceClientType=iothubclient%2f1.10.0%20(native%3b%20Linux%3b%20x86_64) | CLEAN: 0
<- 01:30:07 CONNACK | SESSION_PRESENT: true | RETURN_CODE: 0x0
-> 01:30:07 SUBSCRIBE | PACKET_ID: 22534 | TOPIC_NAME: $iothub/twin/res/# | QOS: 0 | TOPIC_NAME: $iothub/twin/PATCH/properties/desired/# | QOS: 0
-> 01:30:07 PUBLISH | IS_DUP: false | RETAIN: 0 | QOS: DELIVER_AT_LEAST_ONCE | TOPIC_NAME: devices/temperature/messages/events/%24.ct=application%2Fjson&%24.ce=utf-8 | PACKET_ID: 21852 | PAYLOAD_LEN: 688479
-> 01:30:07 PUBLISH | IS_DUP: false | RETAIN: 0 | QOS: DELIVER_AT_LEAST_ONCE | TOPIC_NAME: devices/temperature/messages/events/%24.ct=application%2Fjson&%24.ce=utf-8 | PACKET_ID: 21853 | PAYLOAD_LEN: 188830
[...]

PR #2519, opened 2023-08-15
Fix DUP flag in re-sent telemetry over MQTT protocol

ewertons:
# Checklist

https://github.com/azure/azure-iot-sdk-c/issues/2514

# Description of the problem

Unfortunately telemetry messages that are resent by the MQTT transport are not marked with the DUP flag as TRUE (as should by spec MQTT-3.3.1-2).

# Description of the solution
 
Set the DUP flag for telemetry messages resent by the MQTT transport.

Note that telemetry messages are always sent by the azure-iot-sdk-c MQTT transport with QOS1, so no control is necessary for the DUP flag related to QOS0.

azure-iot-sdk-c's MQTT transport does not attempt ot resend TWIN messages or direct methods responses, so this fix does not apply to these messaging features.


CIPop (approved):
Thanks @ewertons !

vaavva (approved):

ericwol-msft (approved):

PR #2564, opened 2024-01-22
Switching usage of srand / rand to srandom / random

RLeclair:
# Checklist
- [ x] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/main/.github/CONTRIBUTING.md).
  - [ x] This pull-request is submitted against the `main` branch. 
  - [ x] I have merged the latest `main` branch prior to submission and re-merged as needed after I took any feedback.
  - [ x] I have squashed my changes into one with a clear description of the change.

# Description of the problem

srand and rand can be less random in some systems. 

# Description of the solution
 
Switching to srandom and random.

### DIFF
--- a/iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
+++ b/iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
@@ -41,7 +41,7 @@ int main(void)
     IOTHUB_MODULE_CLIENT_LL_HANDLE iotHubModuleClientHandle;
     EVENT_INSTANCE messages[MESSAGE_COUNT];
 
-    srand((unsigned int)time(NULL));
+    srandom((unsigned int)time(NULL));
     double avgWindSpeed = 10.0;
     double minTemperature = 20.0;
     double minHumidity = 60.0;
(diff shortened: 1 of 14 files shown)


iothub_client/samples/iothub_client_sample_module_sender/iothub_client_sample_module_sender.c
```diff
@@ -41,7 +41,7 @@ int main(void)
     IOTHUB_MODULE_CLIENT_LL_HANDLE iotHubModuleClientHandle;
     EVENT_INSTANCE messages[MESSAGE_COUNT];
 
-    srand((unsigned int)time(NULL));
+    srandom((unsigned int)time(NULL));
```
ewertons, line 44: Here and everywhere:
srand is a C99-spec function. It is supported by every platform that supports C99, as azure-iot-sdk-c is expected to support.

srandom is a POSIX-1 function.
https://linux.die.net/man/3/srandom

This means that with this change, not all platforms will be supported by the azure-iot-sdk-c. 
We might disrupt customers with this change.

I'm inclined to say that we will need an adapter in c-utility for this (like it has been done for other related problems). If the platform is Linux, then use srandom. Otherwise, use srand.

ewertons (changes requested):

ericwol-msft (changes requested):
blocking the task was to understand the rand usage and not make a fix.

PR #2671, opened 2025-03-18
Fix upload to blob when using DPS with x509 certificate

ewertons:
# Checklist

# Description of the problem

The credentials were not being propagated to the upload to blob layer when
using device provisioning with x509 certificate.

# Description of the solution
 
Implement passing the x509 information from the hsm framework to the upload to blob layer.

### DIFF
--- a/iothub_client/src/iothub_client_core_ll.c
+++ b/iothub_client/src/iothub_client_core_ll.c
@@ -398,9 +398,10 @@ static int create_blob_upload_module(IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* handle_d
     }
     else
     {
+        IOTHUB_CREDENTIAL_TYPE credential_type = IoTHubClient_Auth_Get_Credential_Type(handle_data->authorization_module);
+
         if (use_dev_auth &&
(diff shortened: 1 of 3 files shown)


iothub_client/src/iothub_client_core_ll.c
```diff
@@ -399,7 +398,44 @@ static int create_blob_upload_module(IOTHUB_CLIENT_CORE_LL_HANDLE_DATA* handle_d
     }
     else
     {
-        result = 0;
+        IOTHUB_CREDENTIAL_TYPE credential_type = IoTHubClient_Auth_Get_Credential_Type(handle_data->authorization_module);
+
+        if (use_dev_auth &&
```
RLeclair, line 403: Are there other types of credentials we'd want to propagate as well?

RLeclair (approved):
LGTM!

vaavva (approved):

iothub_client/src/iothub_client_core_ll.c
ewertons: Yes, there is DEVICE KEYs and SAS, but these are discouraged methods when using DPS. Customers should use x509 with DPS. Thus we are not prioritizing fixing the path for other auth methods (with DPS! When using IoT Hub directly users can still use any auth mode they desire to use upload to blob).

PR #2697, opened 2026-04-06
Add CSR E2E pipeline for Azure IoT C SDK

maximsemenov80:
This pull request introduces end-to-end (E2E) test coverage for X.509 certificate signing request (CSR) scenarios in the Azure IoT C SDK provisioning client. It adds new E2E tests, corresponding build and pipeline configurations, and makes several supporting code improvements to better handle CSR flows and test infrastructure. The most important changes are grouped below:

**New E2E Test Infrastructure for X.509 CSR:**

* Added a new E2E test suite for X.509 CSR (`prov_x509_csr_client_e2e`) with corresponding CMake build configuration, test runner (`main.c`), and per-function test registration for integration with CTest/Azure DevOps. [[1]](diffhunk://#diff-17981923f36287381a0f39a9cc4d892d2b63553646c8392830b1268aa59ea2dbR1-R106) [[2]](diffhunk://#diff-95d43d0b7578808f6e833943e23e021a36a51b9b21c32be8be8476f35bc893c2R1-R27)
* Enabled the new test suite in the main provisioning client test CMake configuration, allowing it to run as part of the standard E2E test matrix.
* Introduced a dedicated Azure DevOps pipeline (`vsts/dps-e2e-cert-mgmt.yaml`) for automated provisioning, building, testing, and cleanup of CSR E2E scenarios, including support for resource reuse and artifact management.

**Enhancements to CSR and Security Handling:**

* Improved the handling of intermediate HTTP 202 responses in the CSR completion logic to ensure requests are not prematurely removed from the queue and that the correct callback is invoked for the convenience layer.
* Updated the fallback and error handling for unexpected status codes in CSR completion to clarify behavior and ensure requests expire naturally when needed.

**Security Factory and HSM Interface Improvements:**

* Reset security type and device HSM type to `UNKNOWN` during deinitialization to ensure a clean state for subsequent tests or SDK usage. [[1]](diffhunk://#diff-8214c69aa5bcd4dc39e63aada1d1a531223b8e88ce8c55f91069023b9e0f8595R104) [[2]](diffhunk://#diff-2574fd851181560bdd9100dea7a520ade4adf7c0bec6a44e2750da0e463a9227R106)
* Added a new API (`symm_key_info_set`) to set symmetric key and registration name in the test HSM implementation, and exposed this via the key interface for improved testability. [[1]](diffhunk://#diff-0fd41f0175f50af2f92e99f2678c8abbdab1679326c93c9750b534fa2efe9ffdR79-R108) [[2]](diffhunk://#diff-8656c71fae5a4773ef308adf4bb1d712d22fe163bfaa4a21e58dba5ece31aafdR22) [[3]](diffhunk://#diff-083004c7b63098a0435406e682fc62e9d8bf1e95ae8ed84ae7dcac45643078f2R445-R467)

These changes collectively enable robust automated validation of X.509 CSR provisioning scenarios and improve the maintainability and correctness of the provisioning client and test codebase.


provisioning_client/tests/prov_x509_csr_client_e2e/main.c
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
```
ewertons, line 1: e2e test with iot hub + dps can live unde the dps e2e tests. We never had an e2e test with both clients together, but that looks good.

provisioning_client/tests/common_prov_e2e/prov_hsm/symm_key.c
```diff
@@ -76,3 +76,33 @@ const char* symm_key_info_get_reg_id(SYMM_KEY_INFO_HANDLE handle)
     }
     return result;
 }
+
+int symm_key_info_set(SYMM_KEY_INFO_HANDLE handle, const char* reg_name, const char* symm_key)
```
ewertons, line 87: hsm should not allow setting the key. That's against the HSM spec

ewertons: If you need to use another key, create a new instance of the hsm module with the new key

provisioning_client/tests/prov_x509_csr_client_e2e/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
...
+if (${use_http})
+    add_definitions(-DUSE_HTTP)
+    set(prov_transport ${prov_transport} prov_http_transport)
+endif()
+if (${use_mqtt})
+    add_definitions(-DUSE_MQTT)
+    set(prov_transport ${prov_transport} prov_mqtt_transport prov_mqtt_ws_transport)
+endif()
+if (${use_amqp})
+    add_definitions(-DUSE_AMQP)
```
ewertons: CSR is only over MQTT (that's my last info from them), so make sure these amqp defines do not run the test unnecessarily

provisioning_client/tests/prov_x509_csr_client_e2e/prov_x509_csr_client_e2e.c
```diff
@@ -0,0 +1,876 @@
...
+// Tests CSR scenarios using the convenience layer API:
+//   - X509 individual enrollment: full 4-phase lifecycle
+//   - Symmetric key group enrollment: full 4-phase lifecycle
+
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <time.h>
+
+#include <openssl/evp.h>
```
ewertons: Our tests are expected to run on Linux and Windows, with their native SSL APis. 
openssl should not be called here directly, it needs to be abstrated.
Either move these to a common e2e module that selects the appropriate SSL api based on OS, or move the functions to the adapters in c-utility.
Or a combination of both.

ewertons (changes requested):

provisioning_client/tests/common_prov_e2e/prov_hsm/symm_key.c
maximsemenov80: **What I changed.** In symm_key.c I made symm_key_info_set set-once: if the instance is already populated it logs "Refusing to re-key an already-initialized HSM instance; destroy and recreate instead" and returns non-zero. 

**Why it is only called once**. HSM_CLIENT_CREATE takes no arguments, so the vtable uses a two-step construction handshake: hsm_client_key_create() allocates an empty instance, and then hsm_client_set_symm_key_info(handle, reg, key) injects the credentials that prov_dev_set_symmetric_key_info(...) stashed in globals. This set call happens exactly once per HSM instance, immediately after create, in both auth-client paths — there is no code path anywhere in the SDK that re-keys an existing instance.

**Why "create with the key" can't land in this PR.** Removing set_symm_key_info and moving the credentials into construction requires changing the HSM_CLIENT_CREATE typedef (or adding a new one) and updating every HSM_CLIENT_KEY_INTERFACE implementation: the reference hsm_client_key.c adapter, the custom_hsm_example sample, edgelet HSM consumers, plus the UTs that mock them (hsm_client_key_ut, prov_auth_client_ut, iothub_auth_client_ut). That is an SDK-wide vtable break and well out of scope for a CSR E2E PR.

PR #2717, opened 2026-05-10
build: use pipeline variable fallback for AZURE_LOCATION

ewertons:
## Summary

Refresh `AZURE_LOCATION` handling in `build/.vsts-ci.yml` to use a pipeline variable with YAML fallback, instead of a runtime parameter.

## Design

- Define a YAML default:
  - `DEFAULT_AZURE_LOCATION: centraluseuap`
- Compute effective region from pipeline variable first, then fallback:
  - `AZURE_LOCATION_EFFECTIVE: $[ coalesce(variables['AZURE_LOCATION'], variables['DEFAULT_AZURE_LOCATION']) ]`
- Use `AZURE_LOCATION_EFFECTIVE` where test resources are created.

## Runtime behavior

1. If pipeline variable `AZURE_LOCATION` is defined in pipeline settings, that value is used.
2. User can edit `AZURE_LOCATION` at run time only if the pipeline variable is marked overridable.
3. If `AZURE_LOCATION` is not defined, YAML defaults to `centraluseuap`.

## Why this model

- Keeps default behavior in code and reviewable.
- Uses Azure DevOps variable UI exactly as intended (visible/editable in the Variables section when enabled).
- Removes the extra runtime-parameter input surface from Run Pipeline.

## Scope

- This change replaces the previous runtime-parameter approach in this PR.


RLeclair (approved):

ewertons:
@copilot resolve the merge conflicts in this pull request

RLeclair (approved):

PR #2724, opened 2026-05-14
Consume macro-utils-c PR #337 (migrated macro-utils)

ewertons:
## Summary

Migrate azure-iot-sdk-c to consume the new **macro-utils-c** (PR [Azure/macro-utils-c#337](https://github.com/Azure/macro-utils-c/pull/337)) via a top-level submodule, replacing the legacy `azure-macro-utils-c`.

This also bumps **c-utility**, **uamqp**, **umqtt**, **uhttp**, **umock-c**, **azure-ctest**, and **c-logging** to their upstream masters which already consume the new macro-utils-c.

### Dependency wiring
- Add `deps/azure-macro-utils-c` as a top-level submodule pointing to the new `macro-utils-c` repo
- CMake compatibility shims so legacy `find_package(azure_macro_utils_c)` and `MACRO_UTILS_INC_FOLDER` continue to work
- Install rules for **c-logging-v2**, **umock_c** library + headers (including `umock_c_prod_compat.h`), and azure_macro_utils forwarding headers

### Build fixes (all platforms)
- Fix unused-variable warnings across ~15 unit test files (artifact of new macro expansion patterns)
- Fix `WINDOWS_EXPORT_ALL_SYMBOLS` crash for DLL targets, .def symbol names, `prov_device_client_dll` LIBRARY name
- Fix `hsm_client_tpm_ut` mock function signature mismatch
- Fix `CMP0079` policy reset in provisioning_client scope

### Warning suppressions

Added in top-level `CMakeLists.txt`, scoped per compiler:

- **Clang C**: `-Wno-typedef-redefinition` (umock_c duplicate typedefs in C99 mode), `-Wno-unused-value`, `-Wno-unused-variable`, `-Wno-unused-but-set-variable`, `-Wno-unused-const-variable`, `-Wno-unknown-pragmas`
- **Clang C++**: `-Wno-unused-variable`, `-Wno-unused-but-set-variable`, `-Wno-unused-const-variable`, `-Wno-unknown-pragmas`, `-Wno-overloaded-virtual`
- **GCC**: `-Wno-strict-aliasing`, `-Wno-array-bounds`, `-Wno-unused-value`, `-Wno-unused-variable`, `-Wno-unused-but-set-variable`, `-Wno-unused-const-variable`, `-Wno-maybe-uninitialized`, `-Wno-unknown-pragmas`, `-Wno-array-parameter`
- **MSVC**: `/wd4702` (unreachable code in c-utility/utpm under /O2)

**What triggers each warning:**

[...]

PR #2731, opened 2026-06-17
Fix E2E wrapper build: emit azure_macro_utils forwarding header into deps inc folder

maximsemenov80:
## Problem

The Horton C E2E image build (`build docker image c`) currently fails on `main` for **every** PR while compiling the `edge_e2e_rest_server` test wrapper:

```
/sdk/iothub_client/inc/internal/iothub_client_edge.h:6:10: fatal error:
    azure_macro_utils/macro_utils.h: No such file or directory
```

## Root cause

The macro-utils-c migration (#2724) moved the submodule to `Azure/macro-utils-c`, which ships headers at `macro_utils/macro_utils.h`. The SDK still includes the legacy `azure_macro_utils/macro_utils.h` in 200+ files and keeps them working via a forwarding header generated into `${CMAKE_BINARY_DIR}` and added with `include_directories(${CMAKE_BINARY_DIR})`.

That shim is only visible to targets in the **SDK's own directory scope**. The E2E test wrappers build their own targets in the **top-level** CMake scope (the SDK is pulled in via `add_subdirectory`) and put `deps/azure-macro-utils-c/inc` directly on their include path — they never see `${CMAKE_BINARY_DIR}`, so the legacy include fails.

Bumping the submodule does not help: `Azure/macro-utils-c` has no `azure_macro_utils/` compatibility header even at latest `master`.

## Fix

Also emit the `azure_macro_utils/` forwarding headers **inside the macro-utils-c include folder** (`deps/azure-macro-utils-c/inc`) at configure time, so external/source-tree consumers that add that folder to their include path keep resolving the legacy `azure_macro_utils/...` includes used throughout the SDK's public headers.

Single-file change in `CMakeLists.txt` (15 lines), mirroring the existing build-dir forwarding logic. This unblocks the Horton C E2E image build for all PRs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,6 +417,21 @@ else ()
     file(WRITE ${CMAKE_BINARY_DIR}/azure_macro_utils/macro_utils_generated.h
         "#include \"macro_utils/macro_utils_generated.h\"\n")
 
+    # The forwarding headers above are only visible to targets in the SDK's
+    # own directory scope (via include_directories(${CMAKE_BINARY_DIR})).
+    # External consumers that add the deps directly to their include path -
+    # e.g. the IoT E2E test wrappers, which put deps/azure-macro-utils-c/inc
+    # on the compiler include path and build their own targets in the
+    # top-level scope - never see ${CMAKE_BINARY_DIR}. Emit the same
+    # forwarding headers inside the macro-utils-c include folder so the
+    # legacy "azure_macro_utils/..." includes used throughout the SDK's
+    # public headers keep resolving for those consumers too.
+    file(MAKE_DIRECTORY ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils)
+    file(WRITE ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils/macro_utils.h
+        "#include \"macro_utils/macro_utils.h\"\n")
+    file(WRITE ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils/macro_utils_generated.h
+        "#include \"macro_utils/macro_utils_generated.h\"\n")
+
     # Install the forwarding headers so consumers of the installed SDK can
     # still #include "azure_macro_utils/macro_utils.h".
     install(FILES


CMakeLists.txt
```diff
@@ -417,6 +417,21 @@ else ()
...
+    # top-level scope - never see ${CMAKE_BINARY_DIR}. Emit the same
+    # forwarding headers inside the macro-utils-c include folder so the
+    # legacy "azure_macro_utils/..." includes used throughout the SDK's
+    # public headers keep resolving for those consumers too.
+    file(MAKE_DIRECTORY ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils)
+    file(WRITE ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils/macro_utils.h
+        "#include \"macro_utils/macro_utils.h\"\n")
+    file(WRITE ${MACRO_UTILS_INC_FOLDER}/azure_macro_utils/macro_utils_generated.h
+        "#include \"macro_utils/macro_utils_generated.h\"\n")
+
```
ewertons, line 434: # Issue

PR #2731's approach writes forwarding headers **into the `deps/azure-macro-utils-c/inc/azure_macro_utils/` folder** — i.e., the SDK generates files into *another library's* (submodule's) source tree at configure time. That's the wrong pattern: it mutates a vendored dependency's checkout, can dirty the submodule working tree, and bakes a workaround for an external consumer into the SDK itself.

**Root cause:** The macro-utils-c migration (#2724) switched to `Azure/macro-utils-c`, which ships headers under `macro_utils/`. The SDK still `#include "azure_macro_utils/macro_utils.h"` in 200+ public headers and bridges that with a forwarding header generated into `${CMAKE_BINARY_DIR}/azure_macro_utils/`, added to *the SDK's own* targets via `include_directories(${CMAKE_BINARY_DIR})`.

The Horton E2E wrapper (`edge_e2e_rest_server` in iot-sdks-e2e-fx) builds its **own** target in the **top-level CMake scope** and includes SDK public headers like iothub_client_edge.h. Since `include_directories(${CMAKE_BINARY_DIR})` is directory-scoped to the SDK subtree, the wrapper's target never sees the forwarding header — hence the failure.

## The correct fix

The fix belongs in the **consumer** (iot-sdks-e2e-fx), not the SDK. The SDK already does the right thing by generating the shim into its own build tree. Since `add_subdirectory` shares one build tree, `${CMAKE_BINARY_DIR}` is the same top-level dir for both — the wrapper just needs to add it to its include path. 


I'll make those changes.

ewertons (commented):
I'll close this PR once I fix the build issue in horton, unblocking https://github.com/Azure/azure-iot-sdk-c/pull/2730

Issue #485, opened 2018-05-10
iothub_client_http_transport should not be required when running a MQTT sample code


formulahendry:
**SDK version used:** Apt-get Xenial: 0.2.0.0-4xenial

**Repro steps:** 
Build the c sample code using https://github.com/VSChina/azure-iot-c-docker/blob/bc83f7ca17e3e255ca6a22525534d249d6b585fd/iothub_client_sample_mqtt/CMakeLists.txt ,

**Actual:**
will meet with below error:
```
Scanning dependencies of target iothub_client_sample_mqtt
[ 50%] Building C object CMakeFiles/iothub_client_sample_mqtt.dir/iothub_client_sample_mqtt.c.o
[100%] Linking C executable iothub_client_sample_mqtt
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `IoTHubClient_LL_Destroy':
(.text+0x9ba): undefined reference to `IoTHubClient_Auth_Destroy'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `initialize_iothub_client':
(.text+0x1b65): undefined reference to `IoTHubClient_Auth_Create'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `initialize_iothub_client':
(.text+0x1dbd): undefined reference to `IoTHubClient_Auth_CreateFromDeviceAuth'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `initialize_iothub_client':
(.text+0x20c7): undefined reference to `IoTHubClient_Auth_Destroy'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `initialize_iothub_client':
(.text+0x2187): undefined reference to `IoTHubClient_Auth_Destroy'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libiothub_client.a(iothub_client_ll.c.o): In function `initialize_iothub_client':
(.text+0x2219): undefined reference to `IoTHubClient_Auth_Destroy'
[...]

Issue #484, opened 2018-05-09
Cross Compilation of Azure SDK with MIPS32 compiler is not working!!


Pavan2358:
# Console log of the issue:
In file included from /home/test/azure_3/azure-iot-sdk-c/c-utility/src/x509_openssl.c:4:0:
/home/test/azure_3/azure-iot-sdk-c/c-utility/inc/azure_c_shared_utility/x509_openssl.h:7:25: fatal error: openssl/ssl.h: No such file or directory
compilation terminated.
make[2]: *** [c-utility/CMakeFiles/aziotsharedutil.dir/build.make:661: c-utility/CMakeFiles/aziotsharedutil.dir/src/x509_openssl.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....

I have attached the toolchain that I am using to cross compile
Note: build.sh works fine with normal gcc compiler and creakes make file for sample code

I just want to compile the code using mips compiler and run sample in my Dlink device... 
please help!!:(
[toolchain-mips.zip](https://github.com/Azure/azure-iot-sdk-c/files/1986577/toolchain-mips.zip)

lakshmisivareddy:
Hi @Pavan2358 
try setting your environment path 
SET(CMAKE_FIND_ROOT_PATH $ENV{SYS_ROOT})
your environment path should contain all your library paths 
and remove all comment ssl,curl,crypto library path

Pavan2358:
Hi @lakshmisivareddy 

I commented as suggested by you....I am stll facing the below mentioned issue

-- Looking for include file stdint.h
-- Looking for include file stdint.h - found
-- Looking for include file stdbool.h
-- Looking for include file stdbool.h - found
-- target architecture: GENERIC
-- Performing Test CXX_FLAG_CXX11
-- Performing Test CXX_FLAG_CXX11 - Failed
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  **Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY) (found
  version "1.1.0h")**
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
[...]

PR #2750, opened 2026-09-11
Bump c-utility and dependents to pick up the Schannel RSA x509 CNG key import fix

ewertons:
Final step of the c-utility submodule dance for [azure-c-shared-utility#699](https://github.com/Azure/azure-c-shared-utility/pull/699).

Bumps `c-utility` directly, and bumps `umqtt`, `uamqp`, `deps/uhttp` and `provisioning_client/deps/utpm` to their master tips so that every copy of the shared utility in the tree carries the fix. The dependent repos were updated first:

- Azure/azure-umqtt-c#177
- Azure/azure-uamqp-c#486
- Azure/azure-uhttp-c#143
- Azure/azure-utpm-c#138

## Why

`x509_schannel.c` published the RSA private key to Schannel as a legacy CryptoAPI provider handle, which restricts Schannel to `rsa_pkcs1_sha1` when signing the TLS 1.2 CertificateVerify message. Against a peer that no longer offers SHA-1 signature algorithms -- anything fronted by OpenSSL 3.x, including an IoT Edge transparent gateway -- there is no mutually supported signature algorithm, and Schannel aborts with `SEC_E_ALGORITHM_MISMATCH` (0x80090331) without sending a certificate and without sending a TLS alert. Azure IoT Hub and DPS still advertise `rsa_pkcs1_sha1`, which is why direct connections kept working and only the gateway path failed.

Submodule bumps only, no source changes in this repo.

Note: the `uamqp` bump also picks up Azure/azure-uamqp-c#483, #484 and #485, which were already merged there.


maximsemenov80 (approved):

ewertons:
The 11 red Linux legs are a real, deterministic break, not e2e flake. Fixed in 63c05d9a.

## What was happening

Every failure was `Run Tests` exiting 8, and in each one only e2e tests failed while all unit tests passed. The failing e2e set varied per leg, which initially looked like flake, but the underlying error is identical everywhere:

```
socketio_berkeley.c:557 set_target_network_interface  Did not find a network interface matching MAC ADDRESS
socketio_berkeley.c:652 initiate_socket_connection    Failure: failed selecting target network interface (MACADDR=00:00:0A:01:00:B4)
socketio_berkeley.c:985 socketio_open                 lookup_address_and_connect_socket failed
```

Only the e2e tests that set `OPTION_NET_INT_MAC_ADDRESS` are affected, which is why the rest of the suite passes and why the failing set differs between legs. The client then retried the connection every 30s until ctest timed the test out.

## Root cause

`get_target_mac_address` in `iothub_client/tests/common_e2e/iothubclient_common_e2e.c` reads the hardware address and the IP address into the **same** `struct ifreq`:

```c
ioctl(s, SIOCGIFHWADDR, &ifr)   // writes ifr.ifr_hwaddr
ioctl(s, SIOCGIFADDR,   &ifr)   // ifr_addr is in the same union - clobbers the MAC
unsigned char* mac = (unsigned char*)ifr.ifr_hwaddr.sa_data;   // formats IP bytes as the MAC
```

That is why the requested MAC is `00:00:0A:01:00:B4` -- two zero bytes followed by `10.1.0.180`, the interface's IPv4 address, not a hardware address at all.

`socketio_berkeley.c` had the exact same bug, so both sides computed the same wrong value and the lookup matched by accident. Azure/azure-c-shared-utility#684 fixed the adapter, and this PR is what brings that fix in: the adapter now reports the real hardware address while the test helper still asked it to bind to an IP-derived one, so the match legitimately fails.

## Fix

Query the hardware address into its own `struct ifreq`, mirroring the adapter fix. The `SIOCGIFADDR` result was never read by this helper; I kept the call so the set of interfaces considered is unchanged.

This was the only occurrence of the pattern in the SDK's own sources.

avishekpant (approved):

PR #2753, opened 2026-09-21
CI: create test resources in eastus2euap

ewertons:
Every run currently fails in resource provisioning, before anything is built:

```
ERROR: (400059) Requested Location centraluseuap is not available.
Subscription is pinned to ihsu-eastus2euap-3
ERROR: "Create Azure IoT Hub" failed (exit code 1)
```

Affects scheduled `main` builds (163947, 163924, 163912) and every PR. Last green run was 2026-09-18.

- `DEFAULT_AZURE_LOCATION`: `centraluseuap` -> `eastus2euap`.
- `.horton-e2e.yml` passes `azure_location: eastus2euap` to the create-resources template, whose default is also `centraluseuap`.

Both are canary regions, so canary coverage is unchanged. The `AZURE_LOCATION` override still takes precedence.

Not verified by running the provisioning script; confirmation comes from this PR's own CI run.


ewertons:
Partial result from this PR's run: `horton-c-gate (setup create_azure_resources)` now passes.

`C-canary (Setup create_azure_resources)` still fails with the same IH400059 on `centraluseuap`. Cause: the C-canary pipeline definition sets `AZURE_LOCATION=centraluseuap`, and `AZURE_LOCATION_EFFECTIVE` coalesces `AZURE_LOCATION` ahead of `DEFAULT_AZURE_LOCATION`, so the YAML default here cannot take effect.

The YAML change is still correct and needed, but C-canary additionally requires that pipeline variable to be set to `eastus2euap`. That change is outside the repo and has been requested separately.

PR #2754, opened 2026-09-21
[lts_03_2025] CI: create test resources in eastus2euap

ewertons:
Backport of #2753 to `lts_03_2025`.

Every run on this branch fails in resource provisioning, before anything is built:

```
ERROR: (400059) Requested Location centraluseuap is not available.
Subscription is pinned to ihsu-eastus2euap-3
ERROR: "Create Azure IoT Hub" failed (exit code 1)
```

`main` fails the same way, so this is not branch-specific.

- `DEFAULT_AZURE_LOCATION`: `centraluseuap` -> `eastus2euap`.
- `.horton-e2e.yml` passes `azure_location: eastus2euap` to the create-resources template, whose default is also `centraluseuap`.

Both are canary regions, so canary coverage is unchanged. The `AZURE_LOCATION` override still takes precedence.

Unblocks CI on #2752.

Not verified by running the provisioning script; confirmation comes from this PR's own CI run.


ewertons:
Partial result from this PR's run: `horton-c-gate (setup create_azure_resources)` now passes.

`C-canary (Setup create_azure_resources)` still fails with the same IH400059 on `centraluseuap`. Cause: the C-canary pipeline definition sets `AZURE_LOCATION=centraluseuap`, and `AZURE_LOCATION_EFFECTIVE` coalesces `AZURE_LOCATION` ahead of `DEFAULT_AZURE_LOCATION`, so the YAML default here cannot take effect.

The YAML change is still correct and needed, but C-canary additionally requires that pipeline variable to be set to `eastus2euap`. That change is outside the repo and has been requested separately.

ewertons:
C-canary re-run on this branch with the region change completed: **all 35 jobs succeeded**, including `create_azure_resources`, which previously failed with IH400059 on `centraluseuap`.

The red `C-canary` mark above is the earlier run, queued before the pipeline variable was updated, and is stale.

